PR #2, opened 2018-05-11
Add assessment section

chris-wood:


draft-rescorla-tls-esni.md
```diff
@@ -365,6 +369,66 @@ has cleaned out most such proxies.
...
+a key. Thus, when one ESNI key is provided, sharing is optimally bound by the number 
+of hosts that share an IP address. Server operators may further limit sharing
+by including multiple keys, with distinct labels, in an ESNIKeys structure.
+
+3. Prevent SNI-based DoS attacks
+
+This design requires servers to decrypt ClientHello messages with EncryptedSNI
+extensions carrying valid labels. Thus, it is possible for an attacker to force
+decryption operations on the server. Servers SHOULD NOT be configured to handle
+multiple keys, as this then require servers to perform trial decryption before
```
ekr: Why is this a problem? You have a label?

draft-rescorla-tls-esni.md
```diff
@@ -365,6 +369,66 @@ has cleaned out most such proxies.
...
+learning of the attack. This attack is bound by the number of valid TCP connections
+an attacker can open. 
+
+4. Do not stick out
+
+By sending SNI and ESNI values (with illegitimate labels), or by sending 
+legitimate ESNI values for and "fake" SNI values, clients do not display
+clear signals of ESNI intent to passive eavesdroppers. As more clients 
+enable ESNI support with keys supplied by shared hosting providers, the 
+presence of ESNI extensions becomes less suspicious.
```
ekr: I think you should sharpen this point a bit. The argument I would make is that this is a property of the browser and the server, so it's not suspicious at all. Now, it's true that you need more clients to avoid *blocking* but if you're using a generic browser, there's nothing suspicious about running Safari, Firefox, etc.

draft-rescorla-tls-esni.md
```diff
@@ -365,6 +369,66 @@ has cleaned out most such proxies.
...
+
+5. Forward secrecy
+
+This design achieves forward secrecy by mixing the client's ephemeral key share
+into the ESNI key derivation.
```
ekr: I don't think this is true, because the server's share is static.

draft-rescorla-tls-esni.md
```diff
@@ -365,6 +369,66 @@ has cleaned out most such proxies.
...
+6. Proper security context
+
+This design permits servers operating in Fronting Mode to forward connections 
+directly to hidden origin servers, thereby avoiding unnecessary MiTM attacks.
+
+7. Fronting server spoofing
+
+Assuming ESNIKeys retrieved from DNS are validated, e.g., via DNSSEC or fetched
+from a trusted Recursive Resolver, spoofing a server operating in Fronting Mode
+is not possible.
```
ekr: This is true, but I think it's worth pointing to the cleartext DNS for why that's not even needed.

draft-rescorla-tls-esni.md
chris-wood: Oh, yes, oops. I was forgot we said above just to fail open if there's a label mismatch.

draft-rescorla-tls-esni.md
chris-wood: I'm assuming a short TTL for these keys --

draft-rescorla-tls-esni.md
chris-wood: Done.

PR #4, opened 2018-05-12
Small rewrite

ekr:


martinthomson:
maybe also point out that once you control DNS, getting a certificate is usually fairly easy

draft-rescorla-tls-esni.md
```diff
@@ -379,13 +379,19 @@ has cleaned out most such proxies.
...
-connect without ESNI in the usual way. Thus, doing so specifically for ESNI yields 
-no new attack vectors.
+information. This means that any attacker which can inject DNS responses or poison
+DNS caches can supply clients with fake ESNIKeys or strip the ESNIKeys from
+the response. However, in the face
+of an attacker that controls DNS, no SNI encryption scheme can work because
+the attacker can replace the IP address, thus blocking client connections, or
+substituting a unique IP address which is 1:1 with the DNS name that was
+looked up (module DNS wildcards). Thus, allowing the ESNIKeys in the clear
+does not make the situation significantly worse.
```
chris-wood: I'd add that the attacker can also inject its own keys, which causes the client to encrypt the SNI to them.

draft-rescorla-tls-esni.md
```diff
@@ -379,13 +379,19 @@ has cleaned out most such proxies.
...
+
+Clearly, DNSSEC (if the client validates and hard fails) is a defense against
+this form of attack, but DoH/DPRIVE are also defenses against DNS attacks
+by attackers on the local network, which is a common case where SNI
+encryption is useful.
```
chris-wood: Maybe add: In fact, SNI encryption is less useful without encryption of DNS queries in transit via DoH or DPRIVE mechanisms.

PR #9, opened 2018-05-12
Address MT's issues.

chris-wood:
Instead of a Fronting SNI field, we should just re-use the Alt-Svc record, which already carries an SNI value to be used.

### DIFF
--- a/draft-rescorla-tls-esni.md
+++ b/draft-rescorla-tls-esni.md
@@ -198,13 +198,10 @@ of "_esni" and the query domain name. That is, if a client queries
 example.com, the ESNI TXT name is _esni.example.com.
 Servers SHOULD configure DNS such that, upon querying a domain name
 with ESNI support, at most one A, AAAA, TXT ESNI, and ALTSVC {{?I-D.schwartz-httpbis-dns-alt-svc}} 
-Resource Record is returned. Providing more address options and ESNI keys could lead
-a client to choose an incorrect key for an address. Alt-Svc records 
-may be used to inform the client of the plaintext (fronting) SNI 
-
-Also, servers operating
-in Fronting Mode SHOULD have DNS configured to return the same A (or AAAA) 
-record for all hidden servers they service.
+Resource Record is returned. Alt-Svc records 
+may be used to inform the client of the plaintext (fronting) SNI. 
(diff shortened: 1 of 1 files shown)


draft-rescorla-tls-esni.md
```diff
@@ -193,10 +193,23 @@ The cipher suite list is orthogonal to the
...
-encoded above. The Resource Record TTL determines the lifetime of
-the published ESNI keys. Clients MUST NOT use ESNI keys beyond
-their recommended lifetime. Note that the length of this structure
-MUST NOT exceed 2^16 - 1, as the RDLENGTH is only 16 bits {{RFC1035}}.
+encoded above. The name of each TXT record MUST match the name composed
+of "_esni" and the query domain name. That is, if a client queries 
+example.com, the ESNI TXT name is _esni.example.com.
+Servers SHOULD configure DNS such that, upon querying a domain name
+with ESNI support, at most one A, AAAA, TXT ESNI, and ALTSVC {{?I-D.schwartz-httpbis-dns-alt-svc}} 
+Resource Record is returned. Providing more address options and ESNI keys could lead
```
ekr: I'm having trouble following this. Is this each record?

draft-rescorla-tls-esni.md
```diff
@@ -193,10 +193,23 @@ The cipher suite list is orthogonal to the
...
+a client to choose an incorrect key for an address. Alt-Svc records 
+may be used to inform the client of the plaintext (fronting) SNI
```
ekr: This sentence seems ot just end.

draft-rescorla-tls-esni.md
chris-wood: Missed period.

draft-rescorla-tls-esni.md
chris-wood: Missed period -- fixed now.

draft-rescorla-tls-esni.md
chris-wood: Indeed, the sentence is confusing. I just removed it. The point I was trying to make is that if there are multiple IP addresses to choose from with multiple keys, and the server does not configure each of those addresses with the set of keys, the client might pick the wrong address/key combo.

PR #11, opened 2018-05-12
Simplify key schedule. Remove Nonce. Fixes #6

### LINKED ISSUE #6: Do you need the nonce?
You are mixing ClientHello.random in.  That would seem to be sufficient.  From that you can generate the AEAD nonce and key as though this were record protection.

```
esni_secret = HKDF-Extract(0, Z)
k = HKDF-Expand-Label(esni_secret, "esni key", ClientHello.random, L)
iv = HKDF-Expand-Label(esni_secret, "esni key", ClientHello.random, L)
```

We don't need another reason for people to complain about exposing the random number generator state to the network.

ekr:


draft-rescorla-tls-esni.md
```diff
@@ -247,22 +243,28 @@ The SNI encryption key is computed from the DH shared secret Z as
...
+
+Note: future extensions may end up reusing the server's ESNIKeyShare
+for other purposes within the same message (e.g., encrypting other
+values). Those usages MUST have their own HKDF labels to avoid
+reuse.
+
+[[OPEN ISSUE: If in future you were to reuse these keys for
+0-RTT priming, then you would have to worry about potentially
+expanding twice of Z_extracted. We should think about how
+to harmonize these.]]
```
chris-wood: This is an implementation detail, I think. If a client chooses to expand twice when it's not needed then that's their fault. We could simply state (in the other draft, or here) that "if clients have computed Z for other purposes, an additional extraction is not needed."

ekr: Sorry, I'm not worried about the implementation, but about whether we're guaranteeing key separation

PR #18, opened 2018-05-12
Add clarity to the DNS encoding section, require base64 encoding, and…

chris-wood:
… deal with TXT string limits.

### DIFF
--- a/draft-rescorla-tls-esni.md
+++ b/draft-rescorla-tls-esni.md
@@ -223,7 +223,8 @@ returned. Alt-Svc records may be used to inform the client of the
 plaintext (fronting) SNI. If present, clients SHOULD use its value
 in the SNI extension of the subsequent ClientHello.
 
-Servers operating in Fronting Mode SHOULD have DNS configured to return 
+Clients obtain these records by querying DNS for hidden server domains.
+Thus, servers operating in Fronting Mode SHOULD have DNS configured to return 
 the same A (or AAAA) record for all hidden servers they service. This yields
 an anonymity set of cardinality equal to the number of hidden server domains
 supported by a given fronting server. Thus, even with SNI encryption,
--- a/draft-rescorla-tls-esni.md
+++ b/draft-rescorla-tls-esni.md
@@ -203,16 +203,26 @@ be used to encrypt the SNI for the associated domain name.
 The cipher suite list is orthogonal to the
 list of keys, so each key may be used with any cipher suite.
 
-This structure is placed in the RRData section of a TXT record as
-encoded above. The name of each TXT record MUST match the name composed
-of "_esni" and the query domain name. That is, if a client queries
-example.com, the ESNI TXT name is _esni.example.com.
+This structure is placed in the RRData section of a TXT record 
+as a base64-encoded string. If this encoding exceeds the 255 octet 
+limit of TXT strings, it must be split across multiple concatenated
+strings as per Section 3.1.3. of {{RFC4408}}.
+
+The name of each TXT record MUST match the name composed
+of \_esni and the query domain name. That is, if a client queries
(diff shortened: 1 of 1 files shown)


chris-wood:
@huitema @ghedo please take a look.

draft-rescorla-tls-esni.md
```diff
@@ -203,16 +203,26 @@ be used to encrypt the SNI for the associated domain name.
...
 
-This structure is placed in the RRData section of a TXT record as
-encoded above. The name of each TXT record MUST match the name composed
-of "_esni" and the query domain name. That is, if a client queries
-example.com, the ESNI TXT name is _esni.example.com.
+This structure is placed in the RRData section of a TXT record 
+as a base64-encoded string. If this encoding exceeds the 255 octet 
+limit of TXT strings, it must be split across multiple concatenated
+strings as per Section 3.1.3. of {{RFC4408}}.
+
```
huitema: I read the text in RFC 4408 and I am a bit concerned that this is one of those things that does not work in theory but does work in practice. In theory, when there are multiple records of the same type for the same name, DNS servers or recursive resolvers can return the records in whichever order they see fit. If they actually do that, the concatenation becomes unpredictable.

draft-rescorla-tls-esni.md
```diff
@@ -203,16 +203,26 @@ be used to encrypt the SNI for the associated domain name.
...
 Servers SHOULD configure DNS such that, upon querying a domain name
-with ESNI support, at most one each of A, AAAA, TXT ESNI, and ALTSVC {{?I-D.schwartz-httpbis-dns-alt-svc}}
-Resource Record is returned. Alt-Svc records
-may be used to inform the client of the plaintext (fronting) SNI.
-Also, servers operating in Fronting Mode SHOULD have DNS configured to
-return the same A (or AAAA) record for all hidden servers they service.
+with ESNI support, at most one each of A, AAAA, TXT ESNI, and 
+ALTSVC {{?I-D.schwartz-httpbis-dns-alt-svc}} Resource Record is 
+returned. Alt-Svc records may be used to inform the client of the 
+plaintext (fronting) SNI. Also, servers operating in Fronting Mode SHOULD
```
huitema: At most one [each] ... TXT ESNI... Is that compatible with the idea that the string may be split over multiple concatenated records?

draft-rescorla-tls-esni.md
```diff
@@ -203,16 +203,26 @@ be used to encrypt the SNI for the associated domain name.
...
 list of keys, so each key may be used with any cipher suite.
 
-This structure is placed in the RRData section of a TXT record as
-encoded above. The name of each TXT record MUST match the name composed
-of "_esni" and the query domain name. That is, if a client queries
-example.com, the ESNI TXT name is _esni.example.com.
+This structure is placed in the RRData section of a TXT record 
+as a base64-encoded string. If this encoding exceeds the 255 octet 
+limit of TXT strings, it must be split across multiple concatenated
+strings as per Section 3.1.3. of {{RFC4408}}.
```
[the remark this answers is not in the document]
chris-wood: Fixed.

draft-rescorla-tls-esni.md
chris-wood: Hmm, interesting. I didn't see records being concatenated. Do we think servers will offer up key sets that need to split up across records?

draft-rescorla-tls-esni.md
chris-wood: The text here suggests that we'd split up text within a single record, not across multiple records. Indeed, if the latter were the case, I don't think this would work.

chris-wood:
@martinthomson @huitema updated based on your feedback -- please take a look!

PR #44, opened 2018-05-17
Remove reference to ALTSVC

### LINKED ISSUE #34: altsvc
I suggest dropping the references to Ben's individual alt-svc in DNS draft.. it hasn't been adopted and its future is murky.

actual alt-svc (the non dns kind) actually requires the use of the plaintext SNI. As those are both cachable and connections to new IP addresses (less linkable) you could consider updating it with mike bishop's https://tools.ietf.org/html/draft-bishop-httpbis-sni-altsvc-01 (also an individual draft atm)

ghedo:
Fixes #34

Also see https://github.com/ekr/draft-rescorla-tls-esni/pull/31#discussion_r188117285

### DIFF
--- a/draft-rescorla-tls-esni.md
+++ b/draft-rescorla-tls-esni.md
@@ -244,10 +244,9 @@ example.com, the ESNI TXT Resource Record might be:
 _esni.example.com. 60S IN TXT "..." "..."
 ~~~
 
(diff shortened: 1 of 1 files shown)


chris-wood:
LGTM -- thanks!

PR #55, opened 2018-07-02
Make server send empty encrypted_server_name in EE

ghedo:
Since encrypted_server_name mirrors server_name, I think it makes sense
for the server to send this back.

---

I don't quite understand the objection regarding "stock servers", @ekr ?

### DIFF
--- a/draft-rescorla-tls-esni.md
+++ b/draft-rescorla-tls-esni.md
@@ -408,6 +408,10 @@ serves the connection directly (if in Shared Mode), in which case
 it executes the steps in the following section, or forwards
 the TLS connection to the hidden server (if in Split Mode).
 
+The server MUST include an extension of type "encrypted_server_name" in the
+EncryptedExtensions message. The "extension_data" field of this extension MUST
+be empty.
+
 ## Shared Mode Hidden Server Behavior
 
 A server operating in Shared Mode uses PaddedServerNameList.sni as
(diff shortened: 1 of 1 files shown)


ekr:
This will make it impossible to run in split mode without modifying your server, which I think is bad.

ekr:
Not putting this in -00

ghedo:
Oh, I see what you mean now. FWIW, we could just make the Shared Mode server echo the extension. RFC6066 says:
```
   A server that receives a client hello containing the "server_name"
   extension MAY use the information contained in the extension to guide
   its selection of an appropriate certificate to return to the client,
   and/or other aspects of security policy.  In this event, the server
   SHALL include an extension of type "server_name" in the (extended)
   server hello.
```

So if we adopt a similar text, the split mode hidden server technically doesn't use ESNI to select the certificate, so it wouldn't need to echo the extension.

ghedo:
Though now I'm questioning how useful this is at all... even with plain "server_name", do clients care if they get back the empty extension?

martinthomson:
Clients just ignore SNI from a server: there is no point in a server ever sending it.  It's just another bit of cruftiness in RFC 6066.  (Oh, how I wish that I could enumerate the ways in which that RFC is bad...)

ghedo:
Ok, so, I guess we don't really want this at all then. I'll make a PR to remove the OPEN ISSUE.

ekr:
Concur.

PR #72, opened 2018-07-11
Support both specific RR type and TXT

### LINKED ISSUE #35: define rules for cname (aka multi-cdn or multi-cloud)

ghedo:
The idea here is to have a new RR type which makes DNS people happier, but also be able to fallback to the plain TXT lookup to address @kazuho's concerns.

This should also fix #35 since the ESNI lookup will follow the CNAME, I _think_.

The ESNI RRData format is kind of dumb, so we probably need to make it better.


ghedo:
There is general opposition to doing both ESNI and TXT records, so I made https://github.com/ekr/draft-rescorla-tls-esni/pull/73 to just add a new ESNI RR type.

Issue #81, opened 2018-07-29
Mix client KeyShare into AEAD to prevent malleability


chris-wood:
See 
https://mailarchive.ietf.org/arch/msg/tls/rc_tgkgVIeB4yeNCgPcyPzRnrBI

ekr:
Done.

PR #85, opened 2018-08-05
Mix the server record (digest), client key share, and client random into the ESNI key derivation.

chris-wood:
Addresses #81.


ekr:
This is an improvement, but I thought we had agreed to move towards the separate keys/AAD approach?

chris-wood:
We did, but that’s slightly orthogonal. I’ll do that in a separate PR.

PR #86, opened 2018-08-06
Add advice about query strategy.

chris-wood:


chris-wood:
@mcmanus please have a look!

PR #87, opened 2018-08-07
Separate the key shares used by the TLS handshake and ESNI to avoid downgrade

chris-wood:


draft-rescorla-tls-esni.md
```diff
@@ -288,20 +288,25 @@ extension, which contains an EncryptedSNI structure:
 ~~~~
    struct {
        CipherSuite suite;
+       KeyShareEntry entry;
```
ekr: I would call this |key_share|

draft-rescorla-tls-esni.md
```diff
@@ -312,25 +317,27 @@ encrypted_sni
...
 ~~~~
    Zx = HKDF-Extract(0, Z)
-   key = HKDF-Expand-Label(Zx, "esni key", Hash(ClientHello.Random), key_length)
-   iv = HKDF-Expand-Label(Zx, "esni iv", Hash(ClientHello.Random), iv_length)
+   key = HKDF-Expand-Label(Zx, "esni key", ESNIContents, key_length)
+   iv = HKDF-Expand-Label(Zx, "esni iv", ESNIContents, iv_length)
 ~~~~
 
+where ESNIContents = Hash(record_digest || ClientHello.KeyShareClientHello || ClientHello.Random),
+and Hash is the hash function associated with the HKDF instantiation.
```
ekr: I think this is not quite what we want.

I think rather the hash should be:

```Hash(record_digest || EncryptedSNI.entry || ClientHello.Random)```

And then you want to have the AAD include ClientHello.key_shares.

draft-rescorla-tls-esni.md
chris-wood: That works for me.

draft-rescorla-tls-esni.md
chris-wood: EncryptedSNI.entry is already mixed in by virtue of being used to derive the KDF input secret. Putting ClientHello.KeyShareClientHello here or in the AAD has the same affect on malleability.

PR #90, opened 2018-09-15
Version number

ekr:


draft-ietf-tls-esni.md
```diff
@@ -236,7 +243,11 @@ list of keys, so each key may be used with any cipher suite.
 This structure is placed in the RRData section of a TXT record
 as a base64-encoded string. If this encoding exceeds the 255 octet
 limit of TXT strings, it must be split across multiple concatenated
-strings as per Section 3.1.3 of {{!RFC4408}}.
+strings as per Section 3.1.3 of {{!RFC4408}}. Servers MAY supply
+multiple TXT records, either of the same or of different versions.
```
chris-wood: I’d replace “multiple TXT records” with “multiple ESNIKeys structures.”

draft-ietf-tls-esni.md
```diff
@@ -236,7 +243,11 @@ list of keys, so each key may be used with any cipher suite.
...
+This allows a server to support multiple versions at once.
+[[OPEN ISSUE: What do you do with multiple concurrent ESNIs of the
```
chris-wood: This should be fine, as clients tell servers which one via the digest. If a client-facing server can’t handle any ESNI for a version it has published then that seems to be a configuration error, no?

PR #94, opened 2018-09-18
Calculate checksum over the whole structure

ghedo:


chris-wood:
@ekr @kazuho what do you think?

PR #98, opened 2018-09-24
Clarify what the server must do with server_name

ghedo:


draft-ietf-tls-esni.md
```diff
@@ -476,7 +476,8 @@ any value other than 0, then the server MUST abort the
 connection with an illegal_parameter alert. Otherwise, the
 server uses the PaddedServerNameList.sni value as if it were
 the "server_name" extension. Any actual "server_name" extension is
-ignored.
+ignored, which also means the server MUST NOT echo back an empty
+"server_name" to the client.
```
ekr: Maybe instead MUST NOT send the "server_name" extension to the client.

chris-wood: I'm not sure this is consistent with the text above, which reads: "The client MAY either omit the "server_name" extension or provide  an innocuous dummy one (this is required for technical conformance with {{!RFC7540}}; Section 9.2.)." It seems we can only prohibit servers from echoing server_name iff the client is also prohibited from sending it in conjunction with ESNI.

ghedo: The server doesn't have to send server_name even if the client does. That is, irrespective of ESNI, the server can decide to ignore the server_name from the client and not send server_name at all. So, the client can send a dummy server_name, but the server is not going to use it (if nothing else, because it's "dummy"), so it shouldn't send server_name.

ghedo: Ok, fixed.

chris-wood:
Discussed offline.

PR #104, opened 2018-10-24
Restrict keys to address records in the same canonical name.

### LINKED ISSUE #35: define rules for cname (aka multi-cdn or multi-cloud)

mcmanus:
fixes #35 

this would be a bit cleaner if we got rid of the prefix.. which I think we ought to do as well but is separable.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -570,7 +570,7 @@ SHOULD be used but is likely to lead to connection failure.
  AAAA ? www.example.com.provider-A.com
      -> AAAA 2001:DB8::AAAA
  TXT ? _esni.www.example.com.provider-B.com
-     -> ".. KEY FOR B.."
+     -> "... KEY FOR B ..."
 ~~~~
 
 The client determines the canonical names for the www.example.com TXT
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -579,7 +579,7 @@ TXT record that matches the canonical address record it already has.
 
 ~~~~
     TXT ? _esni.www.example.com.provider-A.com
-        -> ".. KEY FOR A.."
+        -> "... KEY FOR A ..."
 ~~~~
 
 ## Misconfiguration
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -550,7 +550,7 @@ different providers which in turn will cause connection failures.
 
 In order to prevent this failure the client confirms that the two
 records being used share the same canonical provenance. Specifically,
-the client MUST confirm that the unprefixed (i.e. lacking the
+the client MUST confirm that the unprefixed (i.e., lacking the
 _esni. prefix) canonical name of the TXT record matches the canonical
 name of the address record.
 
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -533,7 +533,7 @@ apply, as detailed below.
 ## Load Balancing Multiple Independent Providers {#multi-cdn}
 
 The DNS is used to independently transfer two linked pieces of
-informaton to implement ESNI. The server is identified through DNS
+information to implement ESNI. The server is identified through DNS
 address records and the server's ESNI configuration is specified in a TXT
 record that is transported separately. {{publishing-key}} requires
 that these records are coordinated.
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -528,6 +530,58 @@ servers which have opted in, thus minimizing compatibility
 issues. However, there are two scenarios where that does not
 apply, as detailed below.
 
+## Load Balancing Multiple Independent Providers {#multi-cdn}
+
+The DNS is used to independently transfer two linked pieces of
+informaton to implement ESNI. The server is identified through DNS
```
```suggestion
information to implement ESNI. The server is identified through DNS
```

draft-ietf-tls-esni.md
```diff
@@ -528,6 +530,58 @@ servers which have opted in, thus minimizing compatibility
...
+address records and the server's ESNI configuration is specified in a TXT
```
martinthomson: does "address records" mean A or AAAA?

draft-ietf-tls-esni.md
```diff
@@ -528,6 +530,58 @@ servers which have opted in, thus minimizing compatibility
...
+     -> CNAME _esni.www.example.com.provider-B.com
+ AAAA ? www.example.com.provider-A.com
+     -> AAAA 2001:DB8::AAAA
+ TXT ? _esni.www.example.com.provider-B.com
+     -> ".. KEY FOR B.."
+~~~~
+
+The client determines the canonical names for the www.example.com TXT
+and AAAA records do not match. It proceeds by obtaining the canonical
+TXT record that matches the canonical address record it already has.
```
martinthomson: How does it know that the TXT record is the one that needs updating?  It is presented with a mismatch, with TTLs.  Should we instead recommend requesting both, or does that invite problems?

Does it start over from _esni.example.com, or does this allow for an end-run on the CNAME resolution process from the prefixed domain.  That is, if www.example.com is CNAMEd through various intermediaries to providerA.example, does the client just construct _esni.providerA.example, or does it have to go back to _esni.example.com?  I get the problem - if this is load balanced, how do we ensure that this ends up routing the same way?  So why would you not just start with _esni.providerA.example then?

draft-ietf-tls-esni.md
```diff
@@ -528,6 +530,58 @@ servers which have opted in, thus minimizing compatibility
...
+implemented as a DNS CNAME record. This provides a level of indirection
+between the domain owner and the hosting provider. When a domain is
+load balanced, the canonical name of the aliased domain potentially
+changes between different providers on a per-request basis. This can
+lead to the use of uncoordinated addressing and key information from
+different providers which in turn will cause connection failures.
+
+In order to prevent this failure the client confirms that the two
+records being used share the same canonical provenance. Specifically,
+the client MUST confirm that the unprefixed (i.e. lacking the
```
```suggestion
the client MUST confirm that the unprefixed (i.e., lacking the
```

draft-ietf-tls-esni.md
```diff
@@ -528,6 +530,58 @@ servers which have opted in, thus minimizing compatibility
...
+
+~~~~
+ AAAA ? www.example.com
+     -> CNAME www.example.com.provider-A.com
+ TXT ? _esni.www.example.com
+     -> CNAME _esni.www.example.com.provider-B.com
+ AAAA ? www.example.com.provider-A.com
+     -> AAAA 2001:DB8::AAAA
+ TXT ? _esni.www.example.com.provider-B.com
+     -> ".. KEY FOR B.."
```
```suggestion
     -> "... KEY FOR B ..."
```

draft-ietf-tls-esni.md
```diff
@@ -528,6 +530,58 @@ servers which have opted in, thus minimizing compatibility
...
+~~~~
+
+The client determines the canonical names for the www.example.com TXT
+and AAAA records do not match. It proceeds by obtaining the canonical
+TXT record that matches the canonical address record it already has.
+
+~~~~
+    TXT ? _esni.www.example.com.provider-A.com
+        -> ".. KEY FOR A.."
```
```suggestion
        -> "... KEY FOR A ..."
```

draft-ietf-tls-esni.md
mcmanus: yes.. I don't mind saying a/aaaa I was trying to leave room for other ways of specifying the address. srv? new A* type?

draft-ietf-tls-esni.md
mcmanus: My thought here is that the TXT record is the new kid on the block, so deference ought to be given to the A for backwards compatibility.

if you're trying to loadbalance, you may well be doing so to add/delete traffic from certain regions (and people have been known to choose providers due to their strengths in particular regions) so capturing the traffic in service of using a valid key seems sub-optimal.. 

I wouldn't be surprised to see keys with bigger TTLs than the A's - creating an easy scenario for the mismatch to occur.. and we wouldn't want the routing of traffic impacted by that imo.

mcmanus:
mostly obe #136 and #137

PR #111, opened 2018-10-26
break RRData into ESNIKeys and ESNIInclude

### LINKED ISSUE #110: ESNIInclude
issues #105 and #79 advocate for the removal of the _esni prefix.

A significant side effect of that is that it becomes impossible to use a CNAME at the apex of a zone in order to refer to ESNIKeys of a server operated elsewhere. Inlining keys, as is often done with A/AAAA at the apex, is not plausible due to key rotation.

The issue can largely be mitigated with the introduction of an inclusion mechanism directly in the TXT/ESNI record.

mcmanus:
fixes #110


chris-wood:
Closing as this is OBE.

Issue #109, opened 2018-10-26
Use ESNI RRType instead of TXT


mcmanus:
PR #73 from @ghedo implements the ESNI RRType, but there doesn't seem to be another issue currently open to track discussion of it.

* a new RRType allows the removal of the _esni prefix (#105 and #79)
* reusing TXT has the potential for conflicts with other uses even with the prefix. (imagine the definition of a TXT record that simply applied to all DNS entries)
* while both TXT and ESNI are 8 bit clean, some people are surprised to learn that about TXT so using a new type may be more robust. Some generic tools processing TXT records would benefit from this.
* TXT is a set of  <= 255 byte strings, defining a new type removes that complexity and reduces overhead
* There was initially concern about the viability of new RRTypes, but I believe people are more comfortable with that now than when the text was drafted.

chris-wood:
Fixed in #144.

Issue #79, opened 2018-07-19
_esni prefix is challenging for DNS wildcard syntax with TXT type


mcmanus:
if you dynamically syntehsize a whole bunch of names in example.com you would probably use a *.example.com a/aaaa configuration to do so and give them all the same addresses.

but you can't do the same things for TXT records with the _esni prefix..  basically _esni.*.example.com doesn't do the trick (rfc 4592).

initially I viewed this as a configuration and implementation issue because queries and responses do not carry the * at all - the records are synthesized on demand from this config.

however, the wildcard is part of AXFR which forms an open interface between resolvers so its syntax and meaning is quite reasonably standardized and should be considered.

kazuho:
> however, the wildcard is part of AXFR which forms an open interface between resolvers

Minor point but I am not sure if AXFR is something that is exposed to the resolvers. My understanding is standardized interface for synchronizing authoritative servers. So this will be an issue only for websites that use a DNS that either depends on AXFR (or in other ways depend on the vocabulary of the AXFR).

I agree that it would be preferable to address the issue.

mcmanus:
my thought here is that axfr represents an interface between, potentially,
different implementations so we should try and respect that to whatever
extent is possible.

kazuho:
My view of the issue is that we can support all the models that we need to support if we adopt #73 so that ESNIKeys will be distributed using a dedicated type with a prefix.

The patterns we need to consider are the combination of three types of DNS records (i.e. single-host, APEX, wildcard) and a boolean indicating if DNS is maintained by someone other than the CDN.

The table below illustrates how you could configure the DNS on each case. There are certain cases in which you could use CNAME instead of A / ESNI.

|**DNS operated by**|**single-host**|**APEX**|**wildcard**|
|:---|:---|:---|:---|
|**CDN**|host IN A addr<br>_esni.host IN ESNI ...|@ IN A addr<br>_esni IN ESNI ...|* IN A addr<br>* IN ESNI ...<sup>1</sup>|
|**not CDN**|host IN CNAME addr<br>_esni.host CNAME _esni.cdn|@ IN A cdn-ipaddr<br>_esni IN CNAME _esni.cdn|* IN CNAME _all.cdn<sup>1<sup>|

1: You will be required to create _esni records for all hosts that you specify explicitly, if it is necessary to disable ESNI for those hosts (or if those hosts use different ESNI keys).

chris-wood:
Fixed in #144.

PR #112, opened 2018-10-27
Convenient integration branch for DNS issues

mcmanus:
includes multicdn #35 
esni rrtype #109 #74
remove _esni prefix #105 #79 
esniinclude #110

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -240,27 +240,24 @@ be used to encrypt the SNI for the associated domain name.
 The cipher suite list is orthogonal to the
 list of keys, so each key may be used with any cipher suite.
 
-This structure is placed in the RRData section of a TXT record
-as a base64-encoded string. If this encoding exceeds the 255 octet
-limit of TXT strings, it must be split across multiple concatenated
-strings as per Section 3.1.3 of {{!RFC4408}}. Servers MAY supply
-multiple ESNIKeys values, either of the same or of different versions.
-This allows a server to support multiple versions at once.
-If the server does not supply any ESNIKeys values with a version
-known to the client, then the client MUST behave as if no
-ESNIKeys were found.
+This structure is placed in the RRData section of an ESNI record
+as-is.  Servers MAY supply multiple ESNIKeys values, either of the
+same or of different versions.  This allows a server to support
+multiple versions at once.  If the server does not supply any ESNIKeys
+values with a version known to the client, then the client MUST behave
+as if no ESNIKeys were found.
 
-The name of each TXT record MUST match the name composed
+The name of each ESNI record MUST match the name composed
 of \_esni and the query domain name. That is, if a client queries
-example.com, the ESNI TXT Resource Record might be:
+example.com, the ESNI Resource Record might be:
 
 ~~~
-_esni.example.com. 60S IN TXT "..." "..."
+_esni.example.com. 60S IN ESNI "..."
 ~~~
 
 Servers MUST ensure that if multiple A or AAAA records are returned for a
 domain with ESNI support, all the servers pointed to by those records are
-able to handle the keys returned as part of a ESNI TXT record for that domain.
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -299,8 +295,42 @@ Client MAY cache the ESNIKeys for a particular domain based on the TTL of the
...
+
+The DNS {{RFC1035}} allows most names to act as an alias via the CNAME
+record type. However, the apex of a zone cannot be an alias. Domain
+owners who cooperate with other organizations for server operation
+sometimes directly publish the A or AAAA address information for the
+servers at the apex of their zone, while relying on the more flexible
+CNAME for other hosted names. Unfortunately, a similar approach does
+not work well for ESNIKeys because they are expected to be rotated
+frequently and thus become out of date if published directly in
+another zone.
```
ekr: This is kind of unclear, but I think the point is that the server operator provides the domain operator with the A record which the domain operator then installs in the DNS, but this works badly for ESNI Keys because then the server operator cannot frequently rotate the ENSI Keys.

draft-ietf-tls-esni.md
```diff
@@ -528,6 +558,57 @@ servers which have opted in, thus minimizing compatibility
 issues. However, there are two scenarios where that does not
 apply, as detailed below.
 
+## Load Balancing Multiple Independent Providers {#multi-cdn}
+
+The DNS is used to independently transfer two linked pieces of
+information to implement ESNI. The server is identified through DNS
+address records and the server's ESNI configuration is specified in a
```
ekr: I would say A/AAAA.

draft-ietf-tls-esni.md
```diff
@@ -528,6 +558,57 @@ servers which have opted in, thus minimizing compatibility
...
+
+~~~~
+ AAAA ? www.example.com
+     -> CNAME www.example.com.provider-A.com
+ ESNI ? www.example.com
+     -> CNAME www.example.com.provider-B.com
+ AAAA ? www.example.com.provider-A.com
+     -> AAAA 2001:DB8::AAAA
+ ESNI ? www.example.com.provider-B.com
+     -> "... KEY FOR B ..."
```
ekr: You should indicate that this is a failure example.

draft-ietf-tls-esni.md
```diff
@@ -528,6 +558,57 @@ servers which have opted in, thus minimizing compatibility
...
+In order to prevent this failure the client confirms that the two
+records being used share the same canonical provenance. Specifically,
+the client MUST confirm that the canonical name of the ESNI record
+matches the canonical name of the address record.
+
+Clients that hold mismatched records MUST resolve the conflict by
+obtaining a new ESNI record (or determining its absence) for the
+canonical name of the address record. If the necessary canonical ESNI
+record itself turns out to be an alias this information SHOULD be used
+but is likely to lead to connection failure.
```
ekr: This text is unclear at least to me.

draft-ietf-tls-esni.md
```diff
@@ -528,6 +558,57 @@ servers which have opted in, thus minimizing compatibility
...
+~~~~
+
+The client determines the canonical names for the www.example.com ESNI
+and AAAA records do not match. It proceeds by obtaining the canonical
+ESNI record that matches the canonical address record it already has.
+
+~~~~
+    ESNI ? www.example.com.provider-A.com
+        -> "... KEY FOR A ..."
+~~~~
```
ekr: This algorithm seems over-complicated. I.e., there's no reason to query for the ESNI at a different CNAME than A/AAAA. Also, it seems like there is an open case here.

````
Query for ESNI and A/AAAA records.
If A/AAAA is CNAME=X {
   Query for ESNI at X
} else {
   if ESNI is not CNAME {
     success
} else {
     ???
}
```

mcmanus:
obe. but #110 should live on.

PR #124, opened 2018-12-17
Improve ESNI robustness

### LINKED ISSUE #82: Consider new extension that permits server to supply new ESNI key upon mismatch
Currently, the connect fails when a mismatch occurs. An alternative strategy might be to return a new ESNI key that is then used for a subsequent connection.

davidben:
(WG email coming shortly.)

Closes #82.

ESNI has different compatibility characteristics than most TLS
extensions. It involves client state, and encrypting the SNI makes the
ClientHello unusable for servers that do not have the key or do not
implement ESNI. This breaks common rollout and rollback strategies and
places strong requirements on DNS/server consistency. This makes
deploying ESNI risky for servers.

This change aims to reduce these deployment risks. ESNI is much more
useful if widely deployed. Moreover, if ESNI is unreliable due to
complex requirements on the server, clients will be disincentivized to
implement it. The DNS/server consistency requirements also
disincentivize rotating keys (section 7.2.5).

This change introduces the notion of a "public name" which allows the
client to receive authenticated signals from the server that the
ESNIKeys are out of sync, so clients can safely repair key mismatches.
It also aligns with existing server behavior before ESNI, so partial
rollouts and rollbacks can also be handled.

The fix for the latter also handles MITM scenarios, replacing the
existing client recommendation:

>   A Web client client can securely detect case (2) because it will
>   result in a connection which has an invalid identity (most likely)
>   but which is signed by a certificate which does not chain to a
>   publicly known trust anchor.  The client can detect this case and
>   disable ESNI while in that network configuration.
>
>   https://tools.ietf.org/html/draft-ietf-tls-esni-02#section-6.2

The public name notion fixes a bug in that text: an attacker could
disable ESNI on all servers, given the existence of *any* such server.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -705,21 +705,20 @@ They MAY attempt to use another server from the DNS results, if one is provided.
 
 A more serious problem is MITM proxies which do not support this
 extension. {{RFC8446}}; Section 9.3 requires that
-such proxies remove any extensions they do not understand.
+such proxies remove any extensions they do not understand. The handshake will
+then present a certificate based on the public name, without echoing the
+"encrypted_server_name" extension to the client.
 
-A non-conformant MITM proxy which forwards the ESNI extension,
+Depending on whether the client is configured to accept the proxy's certificate
+as authoritative for the public name, this may trigger the retry logic described
+in {{handle-server-response}} or result in a connection failure. A proxy which
+is not authoritative for the public name cannot forge a signal to disable ESNI.
+
+A non-conformant MITM proxy which instead forwards the ESNI extension,
 substituting its own KeyShare value, will result in
 the client-facing server recognizing the key, but failing to decrypt
-the SNI. This causes a hard failure.
-
-The public name, however, makes this protocol compatible with deployments that
-use correctly-implemented MITM proxies. If the client has cached an ESNIKey for
-the origin server, the MITM proxy will process the cleartext SNI field and
-terminate a connection to the public name instead. If the client is configured
-to trust the proxy's certificate, it will accept the connection as valid for the
-public name and retry with ESNI disabled. If the client is not configured to
-trust the proxy, it will reject this and not retry, meeting the security
-requirements of this protocol and TLS.
+the SNI. This causes a hard failure. Clients SHOULD NOT attempt to repair the
+connection in this case.
 
 # Security Considerations
 
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -515,17 +515,22 @@ then processes the extension's "response_type" field:
 
   Otherwise, when the handshake completes successfully with the public name
   verified, the client MUST abort the connection with an "esni_required" alert.
-  The client can then regard the ESNI keys as securely replaced by the server.
-  It processes the "retry_keys" field from the server's "encrypted_server_name"
-  extension. If one of the values used a version known to the client, the client
-  SHOULD retry the handshake with a new transport connection, using that value
-  to encrypt the SNI. If no value is applicable, the client SHOULD retry with
-  ESNI disabled.
+  It then processes the "retry_keys" field from the server's
+  "encrypted_server_name" extension.
 
-  These retry keys may only be applied to the retry connection. The client MUST
-  continue to use the previously-advertised keys for subsequent connections. This
-  avoids introducing pinning concerns or a tracking vector, should a malicious
-  server present client-specific retry keys to identify clients.
+  If one of the values contains a version supported by the client, it can regard
+  the ESNI keys as securely replaced by the server. It SHOULD retry the
+  handshake with a new transport connection, using that value to encrypt the
+  SNI. The value may only be applied to the retry connection. The client
+  MUST continue to use the previously-advertised keys for subsequent
+  connections. This avoids introducing pinning concerns or a tracking vector,
+  should a malicious server present client-specific retry keys to identify
+  clients.
+
+  If none of the values provided in "retry_keys" contains a supported version,
+  the client can regard ESNI as securely disabled by the server. As below, it
+  SHOULD then retry the handshake with a new transport connection and ESNI
+  disabled.
 
 - If the field contains any other value, the client MUST abort the connection
   with an "illegal_parameter" alert.
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -158,7 +158,6 @@ server.
 
 ## SNI Encryption
 
-
 First, the provider publishes a public key and metadata which is used for SNI
 encryption for all the domains for which it serves directly or indirectly (via
 Split Mode). This document defines a publication mechanism using DNS, but other
@@ -387,11 +386,12 @@ accepted by the server.
    } AlertDescription;
 ~~~~
 
-Finally, implementations MUST track whether each PSK was established by a
-connection which negotiated this extension with the "esni_accept" response type.
-If so, this is referred to as an "ESNI PSK". Otherwise, it is a "non-ESNI PSK".
-This may be implemented by adding a new field to client and server session
-states.
+Finally, requirements in {{client-behavior}} and {{server-behavior}} require
+implementations to track, alongside each PSK established by a previous
+connection, whether the connection which this extension with the "esni_accept"
+response type. If so, this is referred to as an "ESNI PSK".  Otherwise, it is a
+"non-ESNI PSK". This may be implemented by adding a new field to client and
+server session states.
 
 
 ## Client Behavior {#client-behavior}
@@ -505,21 +505,23 @@ then processes the extension's "response_type" field:
 
 - If the value is "esni_retry_request", the client proceeds with the handshake,
   verifying the certificate against ESNIKeys.public_name as described in
-  {{verify-public-name}}. If the handshake completes successfully, the client
-  MUST abort the connection with an "esni_required" alert.
+  {{verify-public-name}}. If verification or the handshake fails, the client
+  MUST return a failure to calling application. It MUST NOT use the retry keys
+  as described below.
 
-  The client then processes the "retry_keys" field from the server's
-  "encrypted_server_name" extension. If one of the values used a version known
-  to the client, the client SHOULD retry the handshake with a new transport
-  connection, using that value to encrypt the SNI. If no value is applicable,
-  the client SHOULD retry with ESNI disabled.
+  Otherwise, when the handshake completes successfully with the public name
+  verified, the client MUST abort the connection with an "esni_required" alert.
+  The client can then regard the ESNI keys as securely replaced by the server.
+  It processes the "retry_keys" field from the server's "encrypted_server_name"
+  extension. If one of the values used a version known to the client, the client
+  SHOULD retry the handshake with a new transport connection, using that value
+  to encrypt the SNI. If no value is applicable, the client SHOULD retry with
+  ESNI disabled.
 
-  The client MUST NOT use the server-provided retry keys until the handshake
-  completes successfully. On success, it MUST NOT overwrite the DNS-provided
-  keys with the retry keys. It MUST use the retry keys at most once and
-  continue offering DNS-provided keys for subsequent connections. This avoids
-  introducing a tracking vector, should a malicious server present
-  client-specific retry keys.
+  These retry keys may only be applied to the retry connection. The client MUST
+  continue to use the previously cached keys for subsequent connections. This
+  avoids introducing pinning concerns or a tracking vector, should a malicious
+  server present client-specific retry keys to identify clients.
 
 - If the field contains any other value, the client MUST abort the connection
   with an "illegal_parameter" alert.
@@ -528,17 +530,19 @@ If the server negotiates an earlier version of TLS, or if it does not
 provide an "encrypted_server_name" extension in EncryptedExtensions, the
 client proceeds with the handshake, verifying the certificate against
 ESNIKeys.public_name as described in {{verify-public-name}}. The client MUST
-NOT enable the False Start optimization {{RFC7918}}. If the handshake completes
-successfully, the client MUST abort the connection with an "esni_required" alert.
-It then SHOULD retry the handshake with a new transport connection and ESNI
-disabled.
+NOT enable the False Start optimization {{RFC7918}} for this handshake. If
+verification or the handshake fails, the client MUST return a failure to the
+calling application. It MUST NOT treat this as a signal to disable ESNI as
+described below.
+
+Otherwise, when the handshake completes successfully with the public name
+verified, the client MUST abort the connection with an "esni_required" alert.
+The client can then regard ESNI as securely disabled by the server. It SHOULD
+retry the handshake with a new transport connection and ESNI disabled.
 
 [[TODO: Key replacement is significantly less scary than saying that ESNI-naive
   servers bounce ESNI off. Is it worth defining a strict mode toggle in the ESNI
   keys, for a deployment to indicate it is ready for that? ]]
-
-The client MUST NOT treat this as a signal to disable encrypted SNI until the
-handshake completes successfully.
 
 Clients SHOULD implement a limit on retries caused by "esni_retry_request" or
 servers which do not acknowledge the "encrypted_server_name" extension. If the
@@ -552,8 +556,8 @@ extension, it continues with the handshake using the cleartext "server_name"
 extension instead (see {{server-behavior}}). Clients that offer ESNI then
 verify the certificate with the public name, as follows:
 
-- If the server resumed a session or negotiated parameters not based on
-  certificates, the client MUST abort the connection with an illegal_parameter
+- If the server resumed a session or did not negotiate certificate-based
+  authentication, the client MUST abort the connection with an illegal_parameter
   alert. This case is invalid because {{client-behavior}} requires the client
   to only offer ESNI-established sessions, and {{server-behavior}} requires
   the server to decline ESNI-established sessions if it did not accept ESNI.
@@ -565,7 +569,7 @@ verify the certificate with the public name, as follows:
   empty Certificate message, denoting no client certificate.
 
 Note that verifying a connection for the public name does not verify it for the
-origin. The TLS implementation MUST NOT surface such connections as successful to
+origin. The TLS implementation MUST NOT report such connections as successful to
 the application. It additionally MUST ignore all session tickets and session IDs
 presented by the server. These connections are only used to trigger retries, as
 described in {{client-behavior}}. This may be implemented, for instance, by
@@ -665,23 +669,25 @@ specification.
 Unlike most TLS extensions, placing the SNI value in an ESNI extension
 is not interoperable with existing servers, which expect the value in
 the existing cleartext extension. Thus server operators SHOULD ensure
-servers understand a given set of ESNI keys before advertising them in
-DNS. Additionally, servers SHOULD retain support for any
+servers understand a given set of ESNI keys before advertising them.
+Additionally, servers SHOULD retain support for any
 previously-advertised keys for the duration of their validity.
 
 However, in more complex deployment scenarios, this may be difficult
 to fully guarantee. Thus this protocol was designed to be robust in case
-of inconsistencies between DNS and servers, at the cost of extra
-round-trips due to a retry. Two specific scenarios are detailed below.
+of inconsistencies between systems that advertise ESNI keys and servers, at the
+cost of extra round-trips due to a retry. Two specific scenarios are detailed
+below.
 
 ## Misconfiguration and Deployment Concerns {#misconfiguration}
 
-It is possible for DNS and servers to become inconsistent. This may occur, for instance,
-from DNS misconfiguration, caching issues, or an incomplete rollout in a multi-server
-deployment. This may also occur if a server loses its ESNI keys, or if a deployment
-of ESNI must be rolled back on the server.
+It is possible for ESNI advertisements and servers to become inconsistent. This
+may occur, for instance, from DNS misconfiguration, caching issues, or an
+incomplete rollout in a multi-server deployment. This may also occur if a server
+loses its ESNI keys, or if a deployment of ESNI must be rolled back on the
+server.
 
-The retry mechanism repairs most such inconsistencies. If server and DNS keys mismatch,
+The retry mechanism repairs most such inconsistencies. If server and advertised keys mismatch,
 the server will respond with esni_retry_requested. If the server does not understand the
 "encrypted_server_name" extension at all, it will ignore it as required by {{RFC8446}};
 Section 4.1.2. Provided the server can present a certificate valid for the public name,
@@ -700,8 +706,7 @@ such proxies remove any extensions they do not understand.
 A non-conformant MITM proxy which forwards the ESNI extension,
 substituting its own KeyShare value, will result in
 the client-facing server recognizing the key, but failing to decrypt
-the SNI. This causes a hard failure. Hopefully, the TLS 1.3 deployment
-experience has cleaned out most such proxies.
+the SNI. This causes a hard failure.
 
 The public name, however, makes this protocol compatible with deployments that
 use correctly-implemented MITM proxies. If the client has cached an ESNIKey for
@@ -709,8 +714,8 @@ the origin server, the MITM proxy will process the cleartext SNI field and
 terminate a connection to the public name instead. If the client is configured
 to trust the proxy's certificate, it will accept the connection as valid for the
 public name and retry with ESNI disabled. If the client is not configured to
-trust the proxy, it will reject this and not retry, meeting this protocol's and
-TLS's security requirements.
+trust the proxy, it will reject this and not retry, meeting the security
+requirements of this protocol and TLS.
 
 # Security Considerations
 
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -564,11 +564,12 @@ extension, it continues with the handshake using the cleartext "server_name"
 extension instead (see {{server-behavior}}). Clients that offer ESNI then
 verify the certificate with the public name, as follows:
 
-- If the server resumed a session or did not negotiate certificate-based
-  authentication, the client MUST abort the connection with an illegal_parameter
-  alert. This case is invalid because {{send-esni}} requires the client
-  to only offer ESNI-established sessions, and {{server-behavior}} requires
-  the server to decline ESNI-established sessions if it did not accept ESNI.
+- If the server resumed a session or negotiated a session that did not use a
+  certificate for authentication, the client MUST abort the connection with an
+  "illegal_parameter" alert. This case is invalid because {{send-esni}} requires
+  the client to only offer ESNI-established sessions, and {{server-behavior}}
+  requires the server to decline ESNI-established sessions if it did not accept
+  ESNI.
 
 - The client MUST verify that the certificate is valid for ESNIKeys.public_name.
   If invalid, it MUST abort the connection with the appropriate alert.
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -367,7 +367,7 @@ structure:
 
 response_type
 : Indicates whether the server processed the client ESNI extension. (See
-{{client-behavior}} and {{server-behavior}}.}
+{{handle-server-response}} and {{server-behavior}}.}
 
 nonce
 : The contents of ClientESNIInner.nonce. (See {{client-behavior}}.)
@@ -395,6 +395,8 @@ server session states.
 
 
 ## Client Behavior {#client-behavior}
+
+### Sending an encrypted SNI {#send-esni}
 
 In order to send an encrypted SNI, the client MUST first select one of
 the server ESNIKeyShareEntry values and generate an (EC)DHE share in the
@@ -493,6 +495,8 @@ The client MUST place the value of ESNIKeys.public_name in the "server_name"
 extension. (This is required for technical conformance with {{!RFC7540}};
 Section 9.2.)
 
+### Handling the server response {#handle-server-response}
+
 If the server negotiates TLS 1.3 or above and provides an
 "encrypted_server_name" extension in EncryptedExtensions, the client
 then processes the extension's "response_type" field:
@@ -558,7 +562,7 @@ verify the certificate with the public name, as follows:
 
 - If the server resumed a session or did not negotiate certificate-based
   authentication, the client MUST abort the connection with an illegal_parameter
-  alert. This case is invalid because {{client-behavior}} requires the client
+  alert. This case is invalid because {{send-esni}} requires the client
   to only offer ESNI-established sessions, and {{server-behavior}} requires
   the server to decline ESNI-established sessions if it did not accept ESNI.
 
@@ -572,7 +576,7 @@ Note that verifying a connection for the public name does not verify it for the
 origin. The TLS implementation MUST NOT report such connections as successful to
 the application. It additionally MUST ignore all session tickets and session IDs
 presented by the server. These connections are only used to trigger retries, as
-described in {{client-behavior}}. This may be implemented, for instance, by
+described in {{handle-server-response}}. This may be implemented, for instance, by
 reporting a failed connection with a dedicated error code.
 
 ## Client-Facing Server Behavior {#server-behavior}
@@ -691,7 +695,7 @@ The retry mechanism repairs most such inconsistencies. If server and advertised 
 the server will respond with esni_retry_requested. If the server does not understand the
 "encrypted_server_name" extension at all, it will ignore it as required by {{RFC8446}};
 Section 4.1.2. Provided the server can present a certificate valid for the public name,
-the client can safely retry with updated settings, as described in {{client-behavior}}.
+the client can safely retry with updated settings, as described in {{handle-server-response}}.
 
 If the public name does not verify or the retry fails, the client SHOULD NOT
 fall back to cleartext SNI, as this allows a network attacker to disclose the SNI.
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -446,33 +482,112 @@ to harmonize these to make sure that we maintain key separation.]]
...
+  "esni_retry_requested" and the "retry_keys" field set to one or more
+  ESNIKeys structures with up-to-date keys. Servers MAY supply multiple
+  ESNIKeys values of different versions. This allows a server to support
+  multiple versions at once.
 
-- If it is unable to negotiate TLS 1.3 or greater, it MUST
-  abort the connection with a "handshake_failure" alert.
+- The server MUST NOT resume any sessions offered by the client that
+  were established with ESNI. Such sessions are associated with the ESNI name,
+  which the server was unable to decrypt.
```
martinthomson: I had real trouble with this text because it doesn't say which direction the resumption applies to.  In general, when talking about resumption, you need to be very clear about whether you are talking about accepting a PSK for *this* session, or whether you are talking about maybe later establishing a resumed connection.  In this case, it's possible that both apply, so I had trouble with this.

draft-ietf-tls-esni.md
```diff
@@ -157,18 +158,17 @@ server.
...
-This document defines a publication mechanism using DNS, but other mechanisms
-are also possible. In particular, if some of the clients of
-a private server are applications rather than Web browsers, those
-applications might have the public key preconfigured.
+First, the provider publishes a public key and metadata which is used for SNI
+encryption for all the domains for which it serves directly or indirectly (via
+Split Mode). This document defines a publication mechanism using DNS, but other
+mechanisms are also possible. In particular, if some of the clients of a private
+server are applications rather than Web browsers, those applications might have
+the public key preconfigured.
```
chris-wood: s/public key/public key and metadata?

draft-ietf-tls-esni.md
```diff
@@ -446,33 +482,112 @@ to harmonize these to make sure that we maintain key separation.]]
...
+- If the field contains any other value, the client MUST abort the connection
+  with an "illegal_parameter" alert.
+
+If the server negotiates an earlier version of TLS, or if it does not
+provide an "encrypted_server_name" extension in EncryptedExtensions, the
+client proceeds with the handshake, verifying the certificate against
+ESNIKeys.public_name as described in {{verify-public-name}}. The client MUST
+NOT enable the False Start optimization {{RFC7918}}. If the handshake completes
+successfully, the client MUST abort the connection with an "esni_required" alert.
+It then SHOULD retry the handshake with a new transport connection and encrypted
```
chris-wood: s/encrypted SNI/ENSI

draft-ietf-tls-esni.md
chris-wood: My take is that the text refers to the former, not the latter, though it is a little unclear.

draft-ietf-tls-esni.md
```diff
@@ -509,8 +624,15 @@ if it were the "server_name" extension to finish the handshake. It
...
 such that its length equals the size of the largest possible Certificate
 (message) covered by the same ESNI key. Moreover, the server MUST
-include the "encrypted_server_name" extension in EncryptedExtensions,
-and the value of this extension MUST match PaddedServerNameList.nonce.
+include the "encrypted_server_name" extension in EncryptedExtensions
+with the "response_type" field set to "esni_accept" and the "nonce"
+field set to PaddedServerNameList.nonce.
+
+If the server sends a NewSessionTicket message, the ticket MUST be
+ignored by servers not negotiating ESNI, including servers which do
```
chris-wood: s/by servers/by (other) servers?

draft-ietf-tls-esni.md
davidben: It's just the former, yeah.

The intent of the text is to prevent resumption/SNI mismatches. With ESNI and, in particular, this ESNI to SNI fallback, a client offering {(non-GREASE) ESNI, SNI, PSK} now needs to consider all four possible responses:

1. Server ACK'd ESNI and did full handshake
2. Server ACK'd ESNI and resumed PSK
3. Server ignored ESNI (thus used SNI) and did full handshake
4. Server ignored ESNI (thus used SNI) and resumed PSK

In particular, whether the server ACKs or ignores ESNI determines which name the server believes it is acting on behalf of. So we want that to be consistent with the identity the client has associated with the PSK, if any. (1) and (3) are vacuously self-consistent, so that's easy.

As for (2) and (4), presumably the client would not be offering a session good for the public name because it's actually trying to connect to the ESNI name. Thus we want to reject (4) and allow (2). (If we allow (4), then the verify-public-name section needs to decide whether this weird scenario counts as authenticating the replacement ESNI keys and that's weird.) That means the server needs to know not the produce (4).

Now, we could that, if a server ignores the ESNI extension due to key mismatch, it should also ignore the PSK extension. That works, except:
* We want to be robust to server rollbacks and partial deployments. Such a server won't know to ignore the PSK extension.
* Being able to send GREASE extensions is nice and a GREASE client wants to accept (4) since all the names are actually identical all around.

Our thinking was to instead take leaf out of EMS's (and new TLS versions in general) book and partitions the connection/resumption space by whether ESNI was used. We remember a bit in the session state for whether ESNI was used when establishing it. We then say that a client offering ESNI [for real, not GREASE] can only offer ESNI-ful sessions, and a server can only accept ESNI-ful sessions in ESNI-ful handshakes.

This gives all the properties we want:
* Clients offering ESNI will only offer an ESNI-ful session, which is invalid for the server to accept if it declined ESNI. Thus it can reject (4) above.
* This can work with ESNI-oblivious servers on rollback. The server makes the ESNI bit in the session state unparseable by older versions of itself. This is the same sort of thing the server would already have had to do for EMS, or TLS 1.3.
* Clients offering GREASE ESNI will almost always offer ESNI-less sessions (last time I talked to the server, it presumably didn't have ESNI either). The server will decline ESNI (since it's random noise), which means it's allowed to resume ESNI-less sessions.

There isn't an immediate need to constrain ESNI-less sessions, so I didn't in this version of the text. Though maybe a full partition is cleaner? It also occurs to me I have no idea what to do about external PSKs. We could say that if you really want ESNI + external PSK (strange considering that ESNI is usually about picking a certificate anyway...), you have to associate an ESNI bit with the EPSK?? EPSKs are a really underspecified use case that I don't know how to evaluate it.

Anyway, I've tweaked the text slightly and left the above questions as-is for now. Is the new text clearer? / Thoughts on the above?

draft-ietf-tls-esni.md
davidben: Rephrased. (See longer comment on resumption above.)

draft-ietf-tls-esni.md
chris-wood: I think the proposed text clears things up. @martinthomson @ekr I'll take this PR to the list to see if there are any objections.

draft-ietf-tls-esni.md
```diff
@@ -157,18 +158,17 @@ server.
 
 ## SNI Encryption
 
-The protocol designed in this document is quite straightforward.
+The protocol designed in this document is straightforward.
```
martinthomson: Maybe avoid making untestable claims:
```suggestion
```

More so this set of changes makes the design the very opposite of straightforward.

draft-ietf-tls-esni.md
```diff
@@ -347,14 +353,48 @@ For servers (in EncryptedExtensions), this extension contains the following
...
+       esni_accept(0),
+       esni_retry_request(1),
+   } ServerESNIResponseType;
+
    struct {
-       uint8 nonce[16];
+       ServerESNIResponseType response_type;
+       select (response_type) {
+           case esni_accept:        uint8 nonce[16];
+           case esni_retry_request: ESNIKeys retry_keys<1..2^16-1>;
```
martinthomson: We will evolve the format over time.  That means that this field is going to contain a specific version.  If that version is not supported, this design offers no means of recovery.  The client knows for certain that ESNI is supported, but there is no remedy.

Can this include a list of key structures?  That would be consistent with the DNS, which can have multiple alternative records with different versions.

draft-ietf-tls-esni.md
```diff
@@ -347,14 +353,48 @@ For servers (in EncryptedExtensions), this extension contains the following
...
+client when it offered an "encrypted_server_name" extension which was not
+accepted by the server.
+
+~~~~
+   enum {
+       esni_required(121),
+   } AlertDescription;
+~~~~
+
+Finally, implementations MUST track whether each PSK was established by a
```
martinthomson: I'm confused.  This is the first mention of a PSK.  This needs more context.

The point here is that if you get a PSK from a previous connection (presumably an external PSK can have its own rules), then you need to ensure that you mark that PSK as being ESNI or not.  If it is ESNI, then you never resume from an ESNI PSK without ESNI and you never resume from a non-ESNI PSK if you are adding ESNI.

Without context, this is impossible to decipher.  I'm not even sure if I agree with the latter part of the requirement, so explaining why is probably necessary.

Also, this sentence is a run-on, so I can't really parse it.

draft-ietf-tls-esni.md
```diff
@@ -446,33 +490,114 @@ to harmonize these to make sure that we maintain key separation.]]
...
-an "illegal_parameter" alert. Moreover, it MUST check that
-ClientESNIInner.nonce matches the value of the
-"encrypted_server_name" extension provided by the server,
-and otherwise abort the connection with an "illegal_parameter"
-alert.
-
-## Client-Facing Server Behavior
+The client MUST place the value of ESNIKeys.public_name in the "server_name"
+extension. (This is required for technical conformance with {{!RFC7540}};
+Section 9.2.)
```
martinthomson: Can public_name be empty?  If it is, can I omit SNI?

draft-ietf-tls-esni.md
```diff
@@ -446,33 +490,114 @@ to harmonize these to make sure that we maintain key separation.]]
...
+  with an "illegal_parameter" alert.
+
+If the server negotiates an earlier version of TLS, or if it does not
+provide an "encrypted_server_name" extension in EncryptedExtensions, the
+client proceeds with the handshake, verifying the certificate against
+ESNIKeys.public_name as described in {{verify-public-name}}. The client MUST
+NOT enable the False Start optimization {{RFC7918}}. If the handshake completes
+successfully, the client MUST abort the connection with an "esni_required" alert.
+It then SHOULD retry the handshake with a new transport connection and ESNI
+disabled.
```
martinthomson: The SHOULD here is either out of scope, or under-specified.

I think that the intent here is to say that if ESNI is not negotiated for any reason, but the connection is completed successfully with "public_name", then you aren't being attacked, but ESNI has simply been disabled.  A necessary condition for the "SHOULD retry" here is that the connection is successfully established with "public_name", but that connection is not explicit here.

You also want to add a condition here that if the server provides retry keys that the client cannot use, then it can regard ESNI as securely disabled.

If that is not the intent, then this stipulates policy: it's OK to fall back when you have a clear signal that ESNI is enabled, but the server doesn't do ESNI.  That's not a policy I would support.

draft-ietf-tls-esni.md
```diff
@@ -446,33 +490,114 @@ to harmonize these to make sure that we maintain key separation.]]
...
+### Verifying against the public name {#verify-public-name}
+
+When the server cannot decrypt or does not process the "encrypted_server_name"
+extension, it continues with the handshake using the cleartext "server_name"
+extension instead (see {{server-behavior}}). Clients that offer ESNI then
+verify the certificate with the public name, as follows:
+
+- If the server resumed a session or negotiated parameters not based on
+  certificates, the client MUST abort the connection with an illegal_parameter
+  alert. This case is invalid because {{client-behavior}} requires the client
```
martinthomson: This is right, but in the degenerate case it leads to three connections: one to fail resumption, one to get ESNI keys (because presumably you don't want the server that resumes to be able to supply retry keys - something that needs to be said more clearly in previous sections), and finally one to connect with the updated keys to the real host.  That isn't ideal, but I don't see how that could be shortened at all.

draft-ietf-tls-esni.md
```diff
@@ -446,33 +490,114 @@ to harmonize these to make sure that we maintain key separation.]]
...
+  "esni_retry_requested" and the "retry_keys" field set to one or more
+  ESNIKeys structures with up-to-date keys. Servers MAY supply multiple
+  ESNIKeys values of different versions. This allows a server to support
+  multiple versions at once.
+
+- The server MUST ignore all PSK identities in the ClientHello which correspond
+  to ESNI PSKs. ESNI PSKs offered by the client are associated with the ESNI
+  name. The server was unable to decrypt then ESNI name, so it should not resume
+  them when using the cleartext SNI name. This restriction allows a client to
+  reject resumptions in {{verify-public-name}}.
```
martinthomson: Why not prohibit resumption and PSK entirely for the case where the keys need replacing?  This only works for certificate-based server authentication anyway.  Are you really trying to salvage this for servers that authenticate with a PSK?

draft-ietf-tls-esni.md
```diff
@@ -509,8 +634,17 @@ if it were the "server_name" extension to finish the handshake. It
 SHOULD pad the Certificate message, via padding at the record layer,
 such that its length equals the size of the largest possible Certificate
 (message) covered by the same ESNI key. Moreover, the server MUST
-include the "encrypted_server_name" extension in EncryptedExtensions,
-and the value of this extension MUST match PaddedServerNameList.nonce.
+include the "encrypted_server_name" extension in EncryptedExtensions
+with the "response_type" field set to "esni_accept" and the "nonce"
+field set to PaddedServerNameList.nonce.
```
```suggestion
field set to the decrypted PaddedServerNameList.nonce value from the client
"encrypted_server_name" extension.
```

draft-ietf-tls-esni.md
```diff
@@ -521,57 +655,62 @@ PaddedServerNameList.sni and ClientESNIInner.nonce to the backend
...
 
-In general, this mechanism is designed only to be used with
-servers which have opted in, thus minimizing compatibility
-issues. However, there are two scenarios where that does not
-apply, as detailed below.
+Unlike most TLS extensions, placing the SNI value in an ESNI extension
+is not interoperable with existing servers, which expect the value in
+the existing cleartext extension. Thus server operators SHOULD ensure
+servers understand a given set of ESNI keys before advertising them in
+DNS. Additionally, servers SHOULD retain support for any
```
martinthomson: Remove "in DNS"

draft-ietf-tls-esni.md
```diff
@@ -521,57 +655,62 @@ PaddedServerNameList.sni and ClientESNIInner.nonce to the backend
...
+previously-advertised keys for the duration of their validity.
+
+However, in more complex deployment scenarios, this may be difficult
+to fully guarantee. Thus this protocol was designed to be robust in case
+of inconsistencies between DNS and servers, at the cost of extra
```
martinthomson: s/DNS/systems that advertise ESNI keys/

draft-ietf-tls-esni.md
```diff
@@ -521,57 +655,62 @@ PaddedServerNameList.sni and ClientESNIInner.nonce to the backend
...
+round-trips due to a retry. Two specific scenarios are detailed below.
+
+## Misconfiguration and Deployment Concerns {#misconfiguration}
 
-## Misconfiguration
+It is possible for DNS and servers to become inconsistent. This may occur, for instance,
```
martinthomson: s/DNS/advertisements/

draft-ietf-tls-esni.md
```diff
@@ -521,57 +655,62 @@ PaddedServerNameList.sni and ClientESNIInner.nonce to the backend
...
+from DNS misconfiguration, caching issues, or an incomplete rollout in a multi-server
+deployment. This may also occur if a server loses its ESNI keys, or an initial deployment
```
```suggestion
deployment. This may also occur if a server loses its ESNI keys, or if a deployment
```

draft-ietf-tls-esni.md
```diff
@@ -521,57 +655,62 @@ PaddedServerNameList.sni and ClientESNIInner.nonce to the backend
...
-server which is not prepared to receive ESNI, then the server will
-ignore the "encrypted_server_name" extension, as required by
-{{RFC8446}}; Section 4.1.2.  If the servers does not
-require SNI, it will complete the handshake with its default
-certificate. Most likely, this will cause a certificate name
-mismatch and thus handshake failure. Clients SHOULD NOT fall
-back to cleartext SNI, because that allows a network attacker
-to disclose the SNI. They MAY attempt to use another server
-from the DNS results, if one is provided.
+The retry mechanism repairs most such inconsistencies. If server and DNS keys mismatch,
```
martinthomson: s/DNS/advertised

draft-ietf-tls-esni.md
```diff
@@ -521,57 +655,62 @@ PaddedServerNameList.sni and ClientESNIInner.nonce to the backend
...
+the server will respond with esni_retry_requested. If the server does not understand the
+"encrypted_server_name" extension at all, it will ignore it as required by {{RFC8446}};
+Section 4.1.2. Provided the server can present a certificate valid for the public name,
+the client can safely retry with updated settings, as described in {{client-behavior}}.
 
+If the public name does not verify or the retry fails, the client SHOULD NOT
```
martinthomson: Again, you need to separate retries that provide usable keys and ones that don't.

draft-ietf-tls-esni.md
```diff
@@ -521,57 +655,62 @@ PaddedServerNameList.sni and ClientESNIInner.nonce to the backend
...
-but clients might opt to attempt captive portal detection
-to see if they are in the presence of a MITM proxy, and if
-so disable ESNI. Hopefully, the TLS 1.3 deployment experience
-has cleaned out most such proxies.
 
+A non-conformant MITM proxy which forwards the ESNI extension,
+substituting its own KeyShare value, will result in
+the client-facing server recognizing the key, but failing to decrypt
+the SNI. This causes a hard failure. Hopefully, the TLS 1.3 deployment
+experience has cleaned out most such proxies.
```
martinthomson: As much as I like and agree with this sentiment, the last sentence can go.

draft-ietf-tls-esni.md
```diff
@@ -521,57 +655,62 @@ PaddedServerNameList.sni and ClientESNIInner.nonce to the backend
...
+
+The public name, however, makes this protocol compatible with deployments that
+use correctly-implemented MITM proxies. If the client has cached an ESNIKey for
+the origin server, the MITM proxy will process the cleartext SNI field and
+terminate a connection to the public name instead. If the client is configured
+to trust the proxy's certificate, it will accept the connection as valid for the
+public name and retry with ESNI disabled. If the client is not configured to
+trust the proxy, it will reject this and not retry, meeting this protocol's and
+TLS's security requirements.
```
martinthomson: Editorial: avoid possessive for inanimate objects or abstract concepts.

draft-ietf-tls-esni.md
```diff
@@ -446,33 +490,114 @@ to harmonize these to make sure that we maintain key separation.]]
...
+  "encrypted_server_name" extension. If one of the values used a version known
+  to the client, the client SHOULD retry the handshake with a new transport
+  connection, using that value to encrypt the SNI. If no value is applicable,
+  the client SHOULD retry with ESNI disabled.
+
+  The client MUST NOT use the server-provided retry keys until the handshake
+  completes successfully. On success, it MUST NOT overwrite the DNS-provided
+  keys with the retry keys. It MUST use the retry keys at most once and
+  continue offering DNS-provided keys for subsequent connections. This avoids
+  introducing a tracking vector, should a malicious server present
```
kaduk: Should probably be clear about who can do this tracking (a network observer), since the server has other ways to track specific clients.

draft-ietf-tls-esni.md
```diff
@@ -446,33 +490,114 @@ to harmonize these to make sure that we maintain key separation.]]
...
+
+### Verifying against the public name {#verify-public-name}
+
+When the server cannot decrypt or does not process the "encrypted_server_name"
+extension, it continues with the handshake using the cleartext "server_name"
+extension instead (see {{server-behavior}}). Clients that offer ESNI then
+verify the certificate with the public name, as follows:
+
+- If the server resumed a session or negotiated parameters not based on
+  certificates, the client MUST abort the connection with an illegal_parameter
```
kaduk: "parameters not based on certificates" should probably be worded differently.  I assume the intent is that you can't negotiate a PSK-only mode and need to pick something where the server authenticates with a certificate?

draft-ietf-tls-esni.md
```diff
@@ -446,33 +490,114 @@ to harmonize these to make sure that we maintain key separation.]]
...
+  the server to decline ESNI-established sessions if it did not accept ESNI.
+
+- The client MUST verify that the certificate is valid for ESNIKeys.public_name.
+  If invalid, it MUST abort the connection with the appropriate alert.
+
+- If the server requests a client certificate, the client MUST respond with an
+  empty Certificate message, denoting no client certificate.
+
+Note that verifying a connection for the public name does not verify it for the
+origin. The TLS implementation MUST NOT surface such connections as successful to
```
kaduk: I'm not sure that this usage of "origin" is consistent with the other usages in this document; do we want to be talking about the "private domain" instead?

draft-ietf-tls-esni.md
davidben: Done

draft-ietf-tls-esni.md

draft-ietf-tls-esni.md
davidben: I don't quite follow. What's the concern here? This one says "If the public name does not verify". Note this is also existing text, but slightly adapted. From before this PR:

> If the servers does not
> require SNI, it will complete the handshake with its default
> certificate. Most likely, this will cause a certificate name	
> mismatch and thus handshake failure. Clients SHOULD NOT fall	
> back to cleartext SNI, because that allows a network attacker	
> to disclose the SNI. They MAY attempt to use another server	
> from the DNS results, if one is provided.

draft-ietf-tls-esni.md
davidben: This too is an existing sentence from before this PR. :-P Happy to remove it though. Removed.

draft-ietf-tls-esni.md

draft-ietf-tls-esni.md
davidben: Well, it's a little fussy. If we were worried about a malicious server colluding with a network observer, that malicious server could already just broadcast whatever tracking information it has to the clear somewhere else. :-P

But if the client is already not bothering with session resumption or scoping it for tracking reasons, a single-use retry means they don't *also* need to scope this. Honestly, saying that you must scope this at least as tightly as session resumption probably is sufficient? Though there's also thinking through accidental pinning if retry keys manage to change the public name.

I added "pinning concerns" to the text but I dunno what other text is needed given the above. Thoughts?

(Neither of these are obviously fatal. There is a corner of the design space around in-band provisioning of ESNI keys, if we're okay caching these and can stomach the complexity of an in-band retry. I know some folks on the list were interested in that. But I think it makes sense to focus on the simplest version of the idea first.)

draft-ietf-tls-esni.md
davidben: Yeah. Otherwise we don't have a public name to check. :-) How about "did not negotiate certificate-based authentication"?

draft-ietf-tls-esni.md
davidben: The word "origin" shows up elsewhere in the doc a bunch, but I only see "private domain" once. Though "origin" usually says "origin server" or "private origin" and not plain "origin". I think "private origin" fits better here than "origin server". What do you think?

davidben:
> Hopefully I didn't miss anything.

And, as I write this, I realize I forgot to deal with Stephen's comment on the thread about the middlebox stuff! It's quite late here, so I will reword that tomorrow morning.

draft-ietf-tls-esni.md
kaduk: I agree that "private origin" is better than "origin server"

draft-ietf-tls-esni.md
kaduk: I think we need to stay positive, "negotiated a session that did not use a certificate for authentication" -- otherwise we get confused if we have any sort of PSK+cert or other joint authentication scheme.

draft-ietf-tls-esni.md
martinthomson: Prose in support of this would be useful.

draft-ietf-tls-esni.md
martinthomson: This is a matter of not leaning on other text properly.  There are three cases and this says two:

The public name can not verify, which is clearly a terminal error condition.  
The server can indicate that ESNI is disabled, which triggers a retry without ESNI.
The server can provide (usable) alternative keys, which triggers a retry with those keys.  If those keys are no good, that's a terminal condition.

I think that you can simplify this to say that unless ESNI is disabled as a result of successfully establishing a connection to the public name, then a client MUST NOT attempt to connect without ESNI.  (not SHOULD NOT)

Yes, this is probably not your fault.  I'll let others sort that bit out.

draft-ietf-tls-esni.md
martinthomson: I see.  It's not very clear.  I'll make some suggestions.

draft-ietf-tls-esni.md
```diff
@@ -347,14 +353,41 @@ For servers (in EncryptedExtensions), this extension contains the following
...
 
+response_type
+: Indicates whether the server processed the client ESNI extension. (See
+{{client-behavior}} and {{server-behavior}}.}
+
 nonce
 : The contents of ClientESNIInner.nonce. (See {{client-behavior}}.)
 
+retry_keys
+: ESNIKeys structures containing the keys that the client should use on
```
```suggestion
: One or more ESNIKeys structures containing the keys that the client should use on
```

draft-ietf-tls-esni.md
```diff
@@ -446,33 +482,112 @@ to harmonize these to make sure that we maintain key separation.]]
...
+- If the value is "esni_retry_request", the client proceeds with the handshake,
+  verifying the certificate against ESNIKeys.public_name as described in
+  {{verify-public-name}}. If the handshake completes successfully, the client
+  MUST abort the connection with an "esni_required" alert.
+
+  The client then processes the "retry_keys" field from the server's
+  "encrypted_server_name" extension. If one of the values used a version known
+  to the client, the client SHOULD retry the handshake with a new transport
+  connection, using that value to encrypt the SNI. If no value is applicable,
+  the client SHOULD retry with ESNI disabled.
```
martinthomson: "If none of the values provided in "retry_keys" contains a version supported by the client, the client MUST consider the server to have disabled ESNI."

What the client does in response to an authenticated signal that disables ESNI is a matter for client policy, so the SHOULD is inappropriate.

BTW, this would argue for having a third option: esni_disabled.

draft-ietf-tls-esni.md
```diff
@@ -209,6 +208,11 @@ checksum
 of the ESNIKeys structure. For the purpose of computing the checksum, the
 value of the "checksum" field MUST be set to zero.
 
+public_name
+: The name of the entity trusted to update these encryption keys.
+This is used by the server in case there is a key mismatch or the
```
ghedo: This is used by the _client_?

draft-ietf-tls-esni.md
```diff
@@ -347,16 +351,53 @@ For servers (in EncryptedExtensions), this extension contains the following
...
+
+~~~~
+   enum {
+       esni_required(121),
+   } AlertDescription;
+~~~~
+
+Finally, requirements in {{client-behavior}} and {{server-behavior}} require
+implementations to track, alongside each PSK established by a previous
+connection, whether the connection which this extension with the "esni_accept"
```
ghedo: I think this sentence is missing something...

draft-ietf-tls-esni.md
```diff
@@ -446,33 +491,120 @@ to harmonize these to make sure that we maintain key separation.]]
...
+- If the value is "esni_accept", the client MUST check that the extension's
+  "nonce" field matches ClientESNIInner.nonce and otherwise abort the
+  connection with an "illegal_parameter" alert. The client then proceeds
+  with the connection as usual, verifying the certificate against the desired
+  name.
+
+- If the value is "esni_retry_request", the client proceeds with the handshake,
+  verifying the certificate against ESNIKeys.public_name as described in
+  {{verify-public-name}}. If verification or the handshake fails, the client
+  MUST return a failure to calling application. It MUST NOT use the retry keys
```
ghedo: Nit: ...to _the_ calling application.

draft-ietf-tls-esni.md
```diff
@@ -446,33 +491,120 @@ to harmonize these to make sure that we maintain key separation.]]
...
+  verified, the client MUST abort the connection with an "esni_required" alert.
+  The client can then regard the ESNI keys as securely replaced by the server.
+  It processes the "retry_keys" field from the server's "encrypted_server_name"
+  extension. If one of the values used a version known to the client, the client
+  SHOULD retry the handshake with a new transport connection, using that value
+  to encrypt the SNI. If no value is applicable, the client SHOULD retry with
+  ESNI disabled.
+
+  These retry keys may only be applied to the retry connection. The client MUST
+  continue to use the previously cached keys for subsequent connections. This
```
ghedo: Does "previously cached keys" refer to the keys fetched from DNS?

draft-ietf-tls-esni.md
```diff
@@ -446,33 +491,120 @@ to harmonize these to make sure that we maintain key separation.]]
...
+described below.
+
+Otherwise, when the handshake completes successfully with the public name
+verified, the client MUST abort the connection with an "esni_required" alert.
+The client can then regard ESNI as securely disabled by the server. It SHOULD
+retry the handshake with a new transport connection and ESNI disabled.
+
+[[TODO: Key replacement is significantly less scary than saying that ESNI-naive
+  servers bounce ESNI off. Is it worth defining a strict mode toggle in the ESNI
+  keys, for a deployment to indicate it is ready for that? ]]
```
ghedo: I'm not sure I understand this. Where would the toggle go, in DNS? And having the toggle turned on would signal to the client that it must not retry the connection with ESNI disabled?

draft-ietf-tls-esni.md
davidben: Er, I was thinking it's used by the server in that it's the identity the server assumes, but yeah conversely the client verifies it. I've reworded this to not reference either side.

draft-ietf-tls-esni.md
davidben: Oops! The mistake was s/which/negotiated/.

draft-ietf-tls-esni.md
davidben: Added the word non-empty to the public_name description.

draft-ietf-tls-esni.md
davidben: Done. Also removed "as described below". That could be read as "as described below, here's how you avoid using the retry keys", which is weird.

draft-ietf-tls-esni.md
davidben: @martinthomson left a bunch of comments that suggested he didn't want the specification text to talk about DNS keys. Thus, to answer your question, yes but not necessarily so we can't say that. :-P

On the list, someone thought a version of this text implied that you're supposed to keep track of all the retry keys and make sure they never repeat. (Honestly, I think that reading was tenuous to say the least.) So I've tried to avoid it saying MUST NOT use the old keys.

I've now switched it to "previously-advertised" which matches the verb used elsewhere. If you all still don't like it, please give me an alternative! :-)

draft-ietf-tls-esni.md
davidben: Some extension on the ESNIKeys yeah. This is a TODO, not spec text.

draft-ietf-tls-esni.md
davidben: Adopted your text with the MUST NOT.

draft-ietf-tls-esni.md
davidben: Oops. Missed this one somehow. (GitHub's UI is... not the best).

> "If none of the values provided in "retry_keys" contains a version supported by the client, the client MUST consider the server to have disabled ESNI."

I tweaked the wording a bit based on the first part. It seemed better to adjust the surrounding text too. Re the second part, I'm not sure what "[MUST consider](https://tools.ietf.org/html/rfc6919#section-2)" would mean as a normative requirement...

> What the client does in response to an authenticated signal that disables ESNI is a matter for client policy, so the SHOULD is inappropriate.

The text around unechoed extensions also uses SHOULD. It's true that there isn't a single "hook point" for client policy around the disabling bit. It sounds like you might prefer a more "abstracted" text where something first outputs "secure disable", or "secure replace" and then there's policy on that.

I haven't done that yet, in part because it's late on a Friday, but also I think we're well into wordsmithing weeds now. :-) I don't know what the right documentation organization would be for that. Also it's getting pretty long. Thoughts?

> BTW, this would argue for having a third option: esni_disabled.

It's already there by way of not echoing the extension. I think a separate explicit version would only make sense if there are clients who:
* are happy with `esni_retry_requested`
* would be happy with `esni_disabled`
* are *not* happy with the unechoed signal.

That preference is not entirely implausible because not echoing the extension is an implicit signal repurposing existing server behavior. However, that's also kind of the point, since it allows safe rollback. If server deployments cannot rely on that, the cost of deploying ESNI goes up dramatically.

ESNI, even more so than most TLS improvements, really wants a wide deployment, so I think it's good for the document to have a strong opinion here.

(Specialized clients and specialized servers, of course, would probably ignore disable signals, but they would ignore `esni_disabled` too, so that's more about the SHOULD. But it seems to me that's covered by "there may exist valid reasons in particular circumstances to ignore a particular item, but the full implications must be understood and carefully weighed before choosing a different course".)

davidben: Er, hit comment too soon. Will upload the corresponding git commit shortly.

davidben: (Uploaded.)

draft-ietf-tls-esni.md

draft-ietf-tls-esni.md
davidben: (Also missed this. I clearly don't know how to use GitHub...)

Done.

PR #128, opened 2019-01-19
Specify HRR behavior.

chris-wood:


chris-wood:
@kazuho @ekr please have a look.

davidben:
We also need to specify server-side HRR behavior, right? Specifically, the server MUST recheck the second ClientHello's ESNI extension before sending the extension, otherwise putting the key share into the AD field doesn't really mean anything. The attacker would be able to do a cut-and-paste thing.

(We could also incorporate the nonce into key schedule, but that upsets the document's split mode use case.)

chris-wood:
Yep! Thanks for bringing that up. I'll prepare text.

PR #136, opened 2019-02-27
Add support for "combined records" via an ESNIKeys extension

chris-wood:
This change does the following:

1. Add support for mandatory extensions, which are those that clients must process else reject the ESNIKeys record.
2. Add a mandatory extension carrying a set of v4/v6 addresses, which is effectively the combined record mode. 
3. Add text describing how clients ought to resolve ESNI and A/AAAA records, and use the results to initiate TLS connections.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -174,10 +174,59 @@ provider's public key. The provider can then decrypt the extension
 and either terminate the connection (in Shared Mode) or forward
 it to the backend server (in Split Mode).
 
-# Publishing the SNI Encryption Key {#publishing-key}
+# Publishing the SNI Encryption Key in the DNS {#publishing-key}
 
-SNI Encryption keys can be published in the DNS using the ESNIKeys
-structure, defined below.
+Publishing ESNI keys in the DNS requires care to ensure correct behavior.
+There are deployment environments in which a domain is served by multiple server
+operators who do not manage the ESNI Keys. Because ESNIKeys and A/AAAA lookup
+are independent, it is therefore possible to obtain an ESNIKeys record which does
+not match the A/AAAA records. (That is, the host to which an A or AAAA record
+refers is not in possession of the ESNI keys.) Naively using this record would
+result in handshake failure, and possibly require fallback to plaintext SNI.
+The design of the system must therefore allow clients to detect and recover
+from this situation.
+
+Servers operating in Split Mode SHOULD have DNS configured to return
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -306,6 +310,81 @@ servers to rotate the keys often and improve forward secrecy.
...
+
+~~~
+1. If an ESNIKeys response with an "address_set" extension arrives before an A or 
+AAAA response, clients SHOULD initiate TLS with ESNI to the provided address(es).
+
+2. If an A or AAAA response arrives before the ESNIKeys response, clients SHOULD wait up
+to CD milliseconds before initiating TLS to either address. (Clients may begin
+TCP connections in this time. QUIC connections should wait.) If an ESNIKeys
+response with an "address_set" extension arrives in this time, clients SHOULD 
+initiate TLS with ESNI to the provided address(es). If an ESNIKeys response
```
MikeBishop: So in this design, ESNIKeys containing address(es) causes the A/AAAA results to be entirely ignored?

chris-wood: Yep!

draft-ietf-tls-esni.md
```diff
@@ -236,13 +256,18 @@ extensions
 : A list of extensions that the client can take into consideration when
 generating a Client Hello message. The format is defined in
 {{RFC8446}}; Section 4.2. The purpose of the field is to
-provide room for additional features in the future; this document does
-not define any extension.
+provide room for additional features in the future. An extension 
+may be tagged as mandatory by using an extension type codepoint with 
+the high order bit set to 1. A client which receives a mandatory extension 
+they do not understand must reject the record.
```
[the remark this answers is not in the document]
chris-wood: I opted for redundancy here. Happy to remove the subsequent text if people feel that's best.

chris-wood:
> This is a substantial step in the right direction, but the fact that A/AAAA records are basically ignored makes me uneasy. 

That's fair. However, it seems to be the price to get this sort of simplicity. Moreover, as an extension, it's not pressed upon all clients for all providers.

> The fact that A/AAAA results are essentially fully embedded would imply that all logic involved in selectively returning those responses has to be duplicated for ESNIKeys, which is also sub-optimal.

Also true! I'm hoping that folks who are in the position and capable of vending these records can comment on this design's cost.

chris-wood:
Given that there was no serious objection to this change and that it works for at least one operator, I'm merging this PR. We can continue to iterate on #137 as needed in parallel.

PR #137, opened 2019-02-27
Add support for "host pointers" via an ESNIKeys extension

chris-wood:
This change does the following:

1. Add support for mandatory extensions, which are those that clients must process else reject the ESNIKeys record.
2. Add a mandatory extension carrying a "host pointer," which is a (probably poorly named) struct containing a "terminal CNAME" for the domain, along with address masks, which may be partial or complete. Clients may use complete addresses if present, else they may use masks to resolve ESNI+A/AAAA mismatches. When these fail, the terminal CNAME may be used to get an address of some host that has the corresponding ESNI key.
3. Add text describing how clients ought to resolve ESNI and A/AAAA records, and use the results to initiate TLS connections.

This is certainly more complex than #136.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -176,7 +176,7 @@ it to the backend server (in Split Mode).
 
 # Publishing the SNI Encryption Key in the DNS {#publishing-key}
 
-Publishing ESNI keys in DNS requires care to ensure correct behavior.
+Publishing ESNI keys in the DNS requires care to ensure correct behavior.
 There are deployment environments in which a domain is served by multiple server
 operators who do not manage the ESNI Keys. Because ESNIKeys and A/AAAA lookup
 are independent, it is therefore possible to obtain an ESNIKeys record which does
@@ -208,8 +208,8 @@ These "host pointers" are encoded using the following structure.
 
 ~~~~
     enum {
-        address_v4,
-        address_v6,
+        address_v4(4),
+        address_v6(6),
     } AddressType;
 
     struct {
@@ -355,7 +355,7 @@ RDLENGTH is only 16 bits {{RFC1035}}.
 
 ## Encrypted SNI DNS Resolution {#esni-resolution}
 
-Clients obtain ESNI records by querying DNS for ESNI-enabled server domains.
+Clients obtain ESNI records by querying the DNS for ESNI-enabled server domains.
 In cases where the domain of the A or AAAA records being resolved do not match the
 SNI Server Name, such as when {{!RFC7838}} is being used, the alternate domain should
 be used for querying the ESNI TXT record.
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -179,10 +179,8 @@ There are deployment environments in which a domain is served by multiple server
 operators who do not manage the ESNI Keys. Because ESNIKeys and A/AAAA lookup
 are independent, it is therefore possible to obtain an ESNIKeys record which does
 not match the A/AAAA records. (That is, the host to which an A or AAAA record
-refers is not in possession of the ESNI keys.) Naively using this record would
-result in handshake failure, and possibly require fallback to plaintext SNI.
-The design of the system must therefore allow clients to detect and recover
-from this situation.
+refers is not in possession of the ESNI keys.) The design of the system must 
+therefore allow clients to detect and recover from this situation.
 
 Servers operating in Split Mode SHOULD have DNS configured to return
 the same A (or AAAA) record for all ESNI-enabled servers they service. This yields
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -306,6 +310,118 @@ servers to rotate the keys often and improve forward secrecy.
...
+3. If an ESNIKeys response with a "host_pointer" extension arrives before this time, 
+and if the address netmask matches that in the corresponding A or AAAA response, 
+then clients SHOULD initiate TLS with ESNI to said address(es). (An ESNIKeys record 
+with a full address has a full netmask, and therefore the match must be exact.) 
+Clients MAY delay this check until both A and AAAA responses arrive, or if a 
+Resolution Delay timeout occurs, according to {{RFC8305}}.
+
+4. If the address netmask does not match the corresponding A or AAAA response,
+and the A or AAAA response yielded a terminal CNAME that matches host_pointer.name,
+then clients SHOULD initiate TLS using ESNI to the address(es) in the A or AAAA records.
```
MikeBishop: So the idea here is that, even if there's no overlap between the A/AAAA results and the ESNI record addresses, if the host_pointer chain includes (terminal might be too strong a statement) the indicated name, then you try it?  This is a bit of a leap of faith on two fronts:  First, that they have the appropriate keys without being in the list of addresses, but second that the caller will get the CNAME chain back rather than just the addresses.

Perhaps a reasonable optimization for those clients that can do it, though.

draft-ietf-tls-esni.md
```diff
@@ -306,6 +310,118 @@ servers to rotate the keys often and improve forward secrecy.
...
+initiate TLS with ESNI to the provided address(es).
+
+2. If an A or AAAA response arrives before the ESNIKeys response, clients SHOULD
+wait up to CD seconds before initiating TLS to either address. (Clients may begin
+TCP connections in this time. QUIC connections should wait.) If an ESNIKeys
+response does not arrive in this time, clients SHOULD initiate TLS without ESNI
+to the provided address(es).
+
+3. If an ESNIKeys response with a "host_pointer" extension arrives before this time, 
+and if the address netmask matches that in the corresponding A or AAAA response,
```
MikeBishop: "the address netmask" isn't defined, here or elsewhere.  I presume that this should read something like "...and if the intersection of the A or AAAA responses and the set of address ranges in address_set is not empty, then clients SHOULD initiate TLS with ESNI to addresses in that intersection."

draft-ietf-tls-esni.md
```diff
@@ -306,6 +310,118 @@ servers to rotate the keys often and improve forward secrecy.
...
+
+3. If an ESNIKeys response with a "host_pointer" extension arrives before this time, 
+and if the address netmask matches that in the corresponding A or AAAA response, 
+then clients SHOULD initiate TLS with ESNI to said address(es). (An ESNIKeys record 
+with a full address has a full netmask, and therefore the match must be exact.) 
+Clients MAY delay this check until both A and AAAA responses arrive, or if a 
+Resolution Delay timeout occurs, according to {{RFC8305}}.
+
+4. If the address netmask does not match the corresponding A or AAAA response,
+and the A or AAAA response yielded a terminal CNAME that matches host_pointer.name,
```
MikeBishop: Likewise, "If the intersection (...) is empty but the A/AAAA resolution included a CNAME which matches...."

draft-ietf-tls-esni.md
```diff
@@ -306,6 +310,118 @@ servers to rotate the keys often and improve forward secrecy.
...
+
+5. If the address netmask does not match the corresponding A or AAAA response,
+and neither the A or AAAA queries yielded a CNAME that matches host_pointer.name,
+clients SHOULD resolve the address of host_pointer.name using an
+address type that matches the A or AAAA response. Clients SHOULD initiate TLS
+using ENSI to the resulting address(es) when A or AAAA resolution completes.
+
+6. If the "host_pointer" address set is empty, clients SHOULD resolve the
+address(es) of host_pointer.name and initiate TLS with ESNI to the resulting 
+address(es).
```
MikeBishop: Isn't this just a subcase of the previous points?  If the address_set is empty, then by definition there's no overlap with the A/AAAA results.  So either the name occurs in the CNAME chain (4) or you resolve the name (5).

draft-ietf-tls-esni.md
chris-wood: Oh, hah, yes!

draft-ietf-tls-esni.md

draft-ietf-tls-esni.md
chris-wood: Yes, technically. I left it as is to be explicit. I can remove it if we think it's excessive.

chris-wood:
@MikeBishop Thanks for the feedback! I'll comment that, as an extension, this is compatible with #136, assuming of course that both extensions are not present. We could move forward with both.

draft-ietf-tls-esni.md
MikeBishop: Maybe instead of listing it as a separate step, just turn it into an unnumbered comment that an empty address set will always land in (4) or (5).

chris-wood:
Closing with #136.

PR #144, opened 2019-03-05
Use new RRType instead of TXT with a custom prefix

chris-wood:
Subsumes #73 and #108.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -244,27 +244,24 @@ be used to encrypt the SNI for the associated domain name.
 The cipher suite list is orthogonal to the
 list of keys, so each key may be used with any cipher suite.
 
-This structure is placed in the RRData section of a TXT record
-as a base64-encoded string. If this encoding exceeds the 255 octet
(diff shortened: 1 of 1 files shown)


mcmanus:
lgtm

ValZapod:
@chris-wood @kazuho @mcmanus 
What the... So all already working eSNI in Cloudflare is already outdated? Do you understand how difficult it will be to get new IN ESNI in Bind and Knot and even in dig utility? Do you understand that while IN txt is working in all OS, IN ESNI will not work in any of them (system resolver, I mean)? Do you understand that you use for ESNI RR 0xff9f (which is 65439), so it is reserved for private use in IANA?
Solution for this. Can you please use both of them?

mcmanus:
>   Internet-Drafts are draft documents valid for a maximum of six months
 >  and may be updated, replaced, or obsoleted by other documents at any
  > time.  It is inappropriate to use Internet-Drafts as reference
  > material or to cite them other than as "work in progress."

Viktor45:
this is good for wiretapping for easy to drop all ESNI with DPI for SNI downgrade

PR #147, opened 2019-03-07
Encrypt nonce for secure sharing of the true SNI with a third party

roelfdutoit:
Third party secret sharing solutions might be developed in future.  This PR enables sharing of the true SNI  while at the same time preventing the third party from impersonating the server.  The client encrypts the nonce with a key derived from Z, and only a server with access to Z will be able to acknowledge the decrypted nonce.


chris-wood:
@roelfdutoit what's the plan for this PR?

PR #153, opened 2019-04-14
Introduce ESNIRecord wrapper and move AddressSet into it

### LINKED ISSUE #150: record_digest should not cover address_set

ghedo:
Fixes #150.

---

Try 2. This turned out to be somewhat tricky due to existing text using "ESNI record" and "ESNIKeys record" (and similar) interchangeably, so I might have missed some instances.

An alternative design would have been to have a single ESNIRecord value per ESNI record, and then have multiple ESNIKeys values for each ESNIRecord (instead of multiple ESNIRecord values per ESNI and single ESNIKeys per ESNIRecord), though that will require more changes.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -176,11 +176,12 @@ it to the backend server (in Split Mode).
 
 Publishing ESNI keys in the DNS requires care to ensure correct behavior.
 There are deployment environments in which a domain is served by multiple server
-operators who do not manage the ESNI Keys. Because ESNIKeys and A/AAAA lookup
-are independent, it is therefore possible to obtain an ESNIKeys record which does
+operators who do not manage the ESNI keys. Because ESNI and A/AAAA lookups
+are independent, it is therefore possible to obtain an ESNI record which does
 not match the A/AAAA records. (That is, the host to which an A or AAAA record
 refers is not in possession of the ESNI keys.) The design of the system must 
-therefore allow clients to detect and recover from this situation.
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -215,8 +216,30 @@ SNI Encryption keys can be published using the following ESNIKeys structure.
...
+as well as some metadata related to those keys.
+
+dns_extensions
+: A list of extensions that the client can take into consideration when
+resolving the target DNS name. The format is defined in
+{{RFC8446}}; Section 4.2. The purpose of the field is to
+provide room for additional features in the future. An extension
+may be tagged as mandatory by using an extension type codepoint with
+the high order bit set to 1. A client which receives a mandatory extension
+they do not understand must reject the ESNIRecord values.
```
dvorak42: Might be useful to call out that these extensions aren't verified as part of the record_digest?

chris-wood: Agreed. I'll merge this and then make that change directly.

chris-wood:
Thanks, @ghedo!

PR #154, opened 2019-04-23
Switch to record_digest_mac to prevent GREASE detection.

dvorak42:
Currently an attacker can distinguish a real ESNI extension and
GREASE extension based on the error returned by replaying the
client's ClientHello with swapped out keyshares.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -440,7 +440,7 @@ used to derive the ESNI key.
 
 record_digest_mac
 : A HMAC value based on the cryptographic hash of the ESNIKeys and the client
-key shares, computed as described below.
+key shares, computed as described in {{record-digest-mac}}.
 
 encrypted_sni
 : The ClientESNIInner structure, AEAD-encrypted using cipher suite "suite" and
@@ -517,7 +517,8 @@ extension, not including the extension header.
 The additional HMAC in this construction prevents an attacker from echoing a
 ClientEncryptedSNI and checking whether the response indicates that this is a
 GREASE ESNI extension or a real ESNI extension. The attacker can only compare
-ESNI extensions against known ESNIKeys value.
+ESNI extensions against known ESNIKeys value. This is further discussed in
+{{do-not-stick-out}}.
 
 
 ## Client Behavior {#client-behavior}
@@ -964,7 +965,7 @@ extensions carrying valid digests. Thus, it is possible for an attacker to force
 decryption operations on the server. This attack is bound by the number of
 valid TCP connections an attacker can open.
 
-### Do not stick out
+### Do not stick out {#do-not-stick-out}
 
 As more clients enable ESNI support, e.g., as normal part of Web browser
 functionality, with keys supplied by shared hosting providers, the presence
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -977,6 +977,20 @@ Additionally, this specification allows for clients to send GREASE ESNI
...
+attacker from being able to distinguish between a real and GREASE ESNI
+extension. If the attacker is able to distinguish the two cases, they'd be able
+to prevent ESNI from being deployed by blocking real ESNI connections,
+disincentivizing server deployment as they would suffer additional breakage.
+When real and GREASE ESNI extensions are indistinguishable, an attacker wouldn't
+be able to block connections based on whether the server supports ESNI.
+
+This construction could be defeated by an attacker which has a list of ESNIKeys
+structures it wishes to block, as the attacker could just MAC each possibility
+and compare against what the client sent, however this is roughly equivalent to
```
chris-wood: s/, however/. However,

draft-ietf-tls-esni.md
```diff
@@ -977,6 +977,20 @@ Additionally, this specification allows for clients to send GREASE ESNI
...
+the attacker blocking the connection based on the IP it is connecting to.
```
chris-wood: nit: s/the IP it is connecting to./the destination IP address./

draft-ietf-tls-esni.md
```diff
@@ -977,6 +977,20 @@ Additionally, this specification allows for clients to send GREASE ESNI
 extensions (see {{grease-extensions}}), which helps ensure the ecosystem
 handles the values correctly.
 
+The additional construction in {{record-digest-mac}} is intended to prevent an
+attacker from being able to distinguish between a real and GREASE ESNI
+extension. If the attacker is able to distinguish the two cases, they'd be able
```
chris-wood: nit: I'd replace "the attacker Adv" with "an attacker", and "they'd" with "Adv"

draft-ietf-tls-esni.md
```diff
@@ -977,6 +977,20 @@ Additionally, this specification allows for clients to send GREASE ESNI
...
+to prevent ESNI from being deployed by blocking real ESNI connections,
+disincentivizing server deployment as they would suffer additional breakage.
```
chris-wood: nit: Perhaps say that this *may* disincentive deployment?

draft-ietf-tls-esni.md
```diff
@@ -495,6 +493,32 @@ connection, whether the connection negotiated this extension with the
...
+~~~~
+
+The HMAC uses the hash function associated with `suite`, and
+ClientHello.KeyShareClientHello is the body of the ClientHello "key_shares"
+extension, not including the extension header.
+
+The additional HMAC in this construction prevents an attacker from echoing a
+ClientEncryptedSNI and checking whether the response indicates that this is a
+GREASE ESNI extension or a real ESNI extension. The attacker can only compare
+ESNI extensions against known ESNIKeys value.
```
chris-wood: Perhaps drop a forward pointer to the security considerations blurb here?

draft-ietf-tls-esni.md
```diff
@@ -438,11 +438,9 @@ key_share
...
 used to derive the ESNI key.
 
-record_digest
-: A cryptographic hash of the ESNIKeys structure from which the ESNI
-key was obtained, i.e., from the first byte of "checksum" to the end
-of the structure.  This hash is computed using the hash function
-associated with `suite`.
+record_digest_mac
+: A HMAC value based on the cryptographic hash of the ESNIKeys and the client
+key shares, computed as described below.
```
chris-wood: Perhaps point to the section where it's defined?

draft-ietf-tls-esni.md
```diff
@@ -952,6 +978,20 @@ Additionally, this specification allows for clients to send GREASE ESNI
...
+extension. An attacker that is able to distinguish the two cases could block
+only real ESNI connections. This may disincentivize servers from deploying ESNI
+as they'd suffer additional breakage when they enable ESNI. When real and GREASE
+ESNI extensions are indistinguishable, an attacker wouldn't be able to block
+connections based on whether the server supports ESNI.
+
+This construction could be defeated by an attacker which has a list of ESNIKeys
+structures it wishes to block, as the attacker could just MAC each possibility
+and compare against what the client sent. However, this is roughly equivalent to
+the attacker blocking the connection based on the destination IP address.
```
ekr: I'm not sure how persuasive I find this. It seems like it's quite likely that the attacker can determine which IP addresses support ESNI, simply by doing reverse resolution and then forward resolution. Why does this help?

dvorak42: A way of avoiding that, though which might be problematic to deploy, is that the reverse resolution name and the public name don't have ESNI records, since you don't necessarily need ESNI when hitting those names, since they're already visible to an attacker.

chris-wood:
Closing in favor of #235!

PR #158, opened 2019-06-02
Clarify AAD for encrypted_sni and HKDF-Extract salt

### LINKED ISSUE #155: Clarification on AAD for ClientEncryptedSNI.encrypted_sni
It was initially not clear to me what should be included in the AAD for encrypting ClientESNIInner into ClientEncryptedSNI.encrypted_sni. The spec currently says:

> The ClientEncryptedSNI.encrypted_sni value is then computed using the usual TLS 1.3 AEAD:
>
>     encrypted_sni = AEAD-Encrypt(key, iv, ClientHello.KeyShareClientHello, ClientESNIInner)
>
> Where ClientHello.KeyShareClientHello is the body of the extension but not including the extension header.

ClientHello.KeyShareClientHello is not defined in this document. At first I thought it would cover ClientEncryptedSNI.key_share, but that seemed odd.

Later I found that it refers to the Key Share extension defined in [RFC 8446, Section 4.2.8. Key Share](https://tools.ietf.org/html/rfc8446#section-4.2.8). Perhaps this reference could be added?

On a related note, `HKDF-Extract(0, Z)` should probably become `HKDF-Extract("", Z)` or be explicit about the salt as is done in https://tools.ietf.org/html/rfc8446#section-7.1 (or https://tools.ietf.org/html/rfc5869#section-2.2).

Lekensteyn:
While at it, remove the redundant "ClientHello." prefix, the name
KeyShareClientHello implies this. Fixes #155
___
@sftcd I did not explicitly mention that "extension_data" is a list of key shares, but hope that the suggested phrase is strong enough to imply that. Do you think that is sufficient or should it become something like:

> Where ClientHello.KeyShareClientHello is the "extension_data" field of the
> "key_share" extension in a Client Hello (Section 4.2.8 of {{!RFC8446}})),
> namely a list of key shares.

(By the way, the list of client key shares could contain one or more entries, but it could also be empty if the client would like to trigger a HRR. That should still be OK since the key/nonce is derived from the client hello random.)

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -530,7 +530,8 @@ computed from Z as follows:
 ~~~~
 
 where ESNIContents is as specified below and Hash is the hash function
-associated with the HKDF instantiation.
+associated with the HKDF instantiation. The salt argument for HKDF-Extract is a
+string consisting of Hash.length bytes set to zeros.
 
 ~~~
    struct {
@@ -577,14 +578,13 @@ The ClientEncryptedSNI.encrypted_sni value is then computed using the usual
 TLS 1.3 AEAD:
 
 ~~~~
-    encrypted_sni = AEAD-Encrypt(key, iv, ClientHello.KeyShareClientHello, ClientESNIInner)
+    encrypted_sni = AEAD-Encrypt(key, iv, KeyShareClientHello, ClientESNIInner)
 ~~~~
 
-Where ClientHello.KeyShareClientHello is the body of the extension but
-not including the extension header. Including
-ClientHello.KeyShareClientHello in the AAD of AEAD-Encrypt binds the
-ClientEncryptedSNI value to the ClientHello and prevents cut-and-paste
-attacks.
+Where KeyShareClientHello is the "extension_data" field of the "key_share"
+extension in a Client Hello (Section 4.2.8 of {{!RFC8446}})). Including
+KeyShareClientHello in the AAD of AEAD-Encrypt binds the ClientEncryptedSNI
+value to the ClientHello and prevents cut-and-paste attacks.
 
 Note: future extensions may end up reusing the server's ESNIKeyShareEntry
 for other purposes within the same message (e.g., encrypting other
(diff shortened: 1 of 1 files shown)


sftcd:
Those changes seem okish though could maybe be better. My issue was that OpenSSL (afaics) previously ignored all the key shares except the one used, so handling two of 'em as the AAD meant adding some new state. No idea if other implementations are like that though. I guess it might help someone else if it explicitly said that there can be >1 key share to include in the AAD.

(BTW, on your BTW, I need to think more about ESNI and HRR handling, haven't tested that yet so not sure if there's a case where the encoding of 0 key shares would be used as ESNI AAD.)

Lekensteyn:
If this adds so much complication, why not drop the KeyShareClientHello completely? If the purpose is to bind this to the original Client Hello, then the Client Random should be sufficient.

An implementation that creates the encrypted_sni field is also responsible for generating a random client hello and a new ESNI key share. Assuming that these are always unique between Client Hellos, then removing the KeyShareClientHello extension should still preserve the binding property.

And even if for some reason the ESNI key share is reused (e.g. shared between different client hello, or reusing one from KeyShareClientHello), the derived key/nonce would still be different if Client Hello Random is unique.

@chris-wood what was the reason for mixing the client key shares list in addition to the existing client random (commit 134b15c87b5c4cde7f52500138662f64757e2e9c)?

ekr:
If you don't bind in the client key share, there is a trivial attack in which the attacker replaces the key share with their own, causing the server's certificate to be encrypted to them.

chris-wood:
> @chris-wood what was the reason for mixing the client key shares list in addition to the existing client random (commit 134b15c)?

What @ekr said above. :-)

chris-wood:
Thanks for the clarifying text!

Lekensteyn:
> If you don't bind in the client key share, there is a trivial attack in which the attacker replaces the key share with their own, causing the server's certificate to be encrypted to them.

Oh I see, the sender of the ESNI extension does not have to prove its knowledge about the ESNI private key, so an attacker could trivially copy the victim's ESNI extension and client random into the attacker's handshake, and recover the certificate if the client key shares were not included. Even if the attacker cannot decrypt SNI, they can still observe the side-effects. Thanks!

I think the text could be clarified to highlight this, I'll prepare a proposal for this.

PR #159, opened 2019-06-03
Clarify how anti-replay protection is achieved

Lekensteyn:
The SNI encryption key is not bound to the Client Hello, one could for
example "cut and paste" the ESNI value in a different Client Hello with
the same Random field, key_share extension and ESNI extension, but with
a different, say, supported_groups extension.

Remove the (EC)DH argument since it only ensures that an attacker cannot
compute the SNI encryption key, but actual anti-replay protection is
provided by binding to the key_share extension.


chris-wood:
This is likely overcome by events in #207. @Lekensteyn, can you please update this PR based on that text, or close it?

chris-wood:
Overcome by #207.

PR #162, opened 2019-06-19
Restrict padded_length values

bemasc:
See discussion at https://github.com/tlswg/draft-ietf-tls-esni/issues/134


ekr:
I am not a fan of this change. I don't like encoding this kind of restrictive choice in the protocol.

bemasc:
OK, I've removed the protocol change and made the restriction purely advisory.

Issue #146, opened 2019-03-05
User tracking via ESNI keys


kirsey:
If each client receives a unique ESNI key from a DNS provider, it is possible to link every usage of this key to the same client. It seems feasible that online services, as well as passive observers of the network traffic, can exploit such a practice for user tracking.

To enhance the privacy protections of this draft, we should describe these privacy risks in the draft. Furthermore, I recommend a clear restriction on the clients’ caching durations of ESNI keys. For example, clients MUST NOT cache ESNI keys longer than 24 hours and as privacy protection, it is RECOMMENDED to cache ESNI keys only for 10 minutes.

jb-wisemo:
To avoid downgrade attacks (by simply filtering out the key DNS response), clients should cache the fact that there is/was an ESNI key significantly longer, perhaps 14 days.  During these 14 days the server can authoritatively report that it no longer has a key, and it can change the key.  But the client would actively look for ESNI support before falling back to plain SNI for that server.

Bug #139 comment 3 contains a proposal that can implicitly handle the case "client thinks server has ESNI, but doesn't know the current keys or if it has been turned off again server side"

ekr:
I do not believe this will work. In multi-cdn scenarios, clients will sometimes just get an ESNIKeys record sometimes and no record other times and that's normal

mcmanus:
I think noting that this can track you within the same origin is a worthwhile thing to note - it gets added to a whole bucket of things with the same property (e.g. AAAA).

As its not really adding anything to that problem I'm not in favor of recommending particular cache ceilings which may or may not really address the issue anyhow..

kirsey:
I observe, that TLS has a tradition in providing upper limits for caching durations. I think this is a good practice as it restricts the harm of an extreme configuration.

Furthermore, I see added value in recommending a best practice configuration at least for a specific application type such as browsers because it leads to a description of the trade-offs at hand and facilitates the implementation.

kontaxis:
The document already sets the maximum TTL to the lifetime of the SNI key.
It also recommends that "servers rotate keys frequently" to reduce the harm from compromised keys.

So it seems that short-lived keys are good both for security and privacy.

A recommended value would be useful. Otherwise "frequently" could be interpreted as months.

whalehub:
@kontaxis I think 24h would be a good timeframe for rotating keys.

PR #166, opened 2019-07-03
Add note about optional record_digest and trial decryption alternative.

chris-wood:
Addresses #138.


chris-wood:
@huitema can you please take a look?

huitema:
Yes, adding a trial decryption is a good thing. Two suggestions:

1) in security section describe privacy attack by tracking the crypto ID.

2) also describe the attack by forcing trial decryption.

Maybe add an option to publish ESNI record with NULL checksum as a way to signal acceptance of trial decryption ?

chris-wood:
> 1. in security section describe privacy attack by tracking the crypto ID.

I plan to address this in a PR for #146.

> 2. also describe the attack by forcing trial decryption.

Good suggestion! I'll add that here.

chris-wood:
@huitema please let me know if the latest text works for you!

draft-ietf-tls-esni.md
```diff
@@ -739,10 +739,15 @@ for servers to proceed with the connection and rely on the client to abort if
...
+wherein trial decryption is a viable approach for matching ClientEncryptedSNI
+contents to a known ESNIKeys. (Some uses of ESNI, such as local discovery mode,
+may opt to omit the ClientEncryptedSNI.record_digest since it can be used as a
+tracking vector.)
 
-If the ClientEncryptedSNI.record_digest value does match the cryptographic
-hash of a known ESNIKeys, the server performs the following checks:
+If the ClientEncryptedSNI.record_digest value does match a known ESNIKeys, either by
+comparing against a known cryptographic hash or by trial decryption, the server
+performs the following checks:
```
davidben: To clarify, is the intent here is that servers outside those environments are not required to implement trial decryption?

The text reads a little unclear to me. The paragraph above just says an empty one "MAY be used in [some environments]", but that could mean by either the client or the server. This paragraph also could read as if the server needs to check the trial decryption, otherwise it doesn't know whether to follow these steps.

(Trial decryption would more fully solve the issue #154 is trying to address, but I worry about the performance cost of asymmetric trial decryption if you rotate keys decently often. Then again, the robustness fixes do mean it's okay to drop older keys sooner if only used by a tiny fraction of clients...)

huitema: The trial decryption should be opt-in. The typical use case is in local discovery, with DTLS or QUIC. Broadcast or multicast the first packet, and then all servers who are ready to be discovered do trial detection. Yes, this is trading off CPU for server privacy, which is only OK in some scenario. In a typical implementation, the server would have a "discovery" flag, which will enable trial decryption when "waiting to be discovered".

chris-wood: > To clarify, is the intent here is that servers outside those environments are not required to implement trial decryption?

Precisely. I see what you mean about it implying trial decryption is necessary. Perhaps: 

```
If the ClientEncryptedSNI.record_digest value does match a known ESNIKeys, the server...
```

And then later on...

```
If ClientEncryptedSNI.record_digest is non-empty, servers SHOULD compare it against cryptographic hashes of known ESNIKeys to find a match. If ClientEncryptedSNI.record_digest is empty, servers MAY use trial decryption to match to a known ESNIKeys.
```

chris-wood: @davidben @huitema: I made an attempt at clarifying the text. Suggestions for improvements are welcome!

draft-ietf-tls-esni.md
```diff
@@ -882,6 +893,14 @@ desired.
...
 without encryption of DNS queries in transit via DoH or DPRIVE mechanisms.
 
+## Optional Record Digests and Trial Decryption
+
+Supporting optional record digests and trial decryption opens oneself up to
+DoS attacks. Specifically, an adversary may send malicious ClientHello messages, i.e.,
+those which will not decrypt with any known ESNI key, in order to force
+decryption. Servers that support this feature should, for example, implement
+some form of rate limiting mechanism to limit the damage caused by such attacks.
+
```
huitema: Looking at this, I wonder whether there are other mitigations than "rate limiting". Take for example the case of a server that has published only one key: trial decryption then is no more expensive that decryption directed by a specific checksum, which the adversary can learn from the DNS record. So an obvious mitigation in case of stress is to only trial one of the published keys, possibly the latest, or a randomly chosen one.

Maybe we should point out the directed DOS attack -- for example, an adversary could bomb a forwarding server with repeated connection request. But then, that's not much more potent than repeated non-ESNI connection requests.

chris-wood: > I wonder whether there are other mitigations than "rate limiting".

There likely probably are, which is why I list this as only one example. :)

chris-wood:
@kazuho @ekr can you please take a look?

draft-ietf-tls-esni.md
```diff
@@ -719,10 +719,17 @@ Upon receiving an "encrypted_server_name" extension, the client-facing
...
-cryptographic hash of any known ESNIKeys structure, it MUST ignore the
-extension and proceed with the connection, with the following added
-behavior:
+The ClientEncryptedSNI value is said to match a known ESNIKeys if there exists
+an ESNIKeys that can be used to successfully decrypt ClientEncryptedSNI.encrypted_sni.
+This matching procedure should be performed as follows. If ClientEncryptedSNI.record_digest
+is non-empty, servers SHOULD compare it against cryptographic hashes of known ESNIKeys and
+choose the one that matches. Otherwise, if ClientEncryptedSNI.record_digest is empty,
+servers MAY use trial decryption to match to a known ESNIKeys. If both checks fail,
+there is no matching ESNIKeys.
```
ekr: This "otherwise" seems to forbid random record_digests (as opposed to empty) or to fall back to trial decryption.. Is that intended?

draft-ietf-tls-esni.md
```diff
@@ -743,10 +750,14 @@ for servers to proceed with the connection and rely on the client to abort if
 ESNI was required. In particular, the unrecognized value alone does not
 indicate a misconfigured ESNI advertisement ({{misconfiguration}}). Instead,
 servers can measure occurrences of the "esni_required" alert to detect this
-case.
+case. An empty ClientEncryptedSNI.record_digest value MAY be used in environments
+wherein trial decryption is a viable approach for matching ClientEncryptedSNI
+contents to a known ESNIKeys. (Some uses of ESNI, such as local discovery mode,
```
ekr: How would you know that this is possible.

draft-ietf-tls-esni.md
chris-wood: Hmm, good point. It's not intended. I'll remove "Otherwise."

draft-ietf-tls-esni.md
chris-wood: It'd be application or deployment-specific. In DNS-SD, for example, they'd likely configure the stack to omit the record digest.

ekr:
I think you should say that it has to be somehow externally configured.

chris-wood:
> I think you should say that it has to be somehow externally configured.

Good suggestion! Will fix.

draft-ietf-tls-esni.md
```diff
@@ -719,10 +719,23 @@ Upon receiving an "encrypted_server_name" extension, the client-facing
...
+The ClientEncryptedSNI value is said to match a known ESNIKeys if there exists
+an ESNIKeys that can be used to successfully decrypt ClientEncryptedSNI.encrypted_sni.
+This matching procedure should be done using one of the following two checks:
+
+1. Compare ClientEncryptedSNI.record_digest against cryptographic hashes of known ESNIKeys 
+and choose the one that matches. 
+2. Use trial decryption of ClientEncryptedSNI.encrypted_sni with known ESNIKeys and choose 
+the one that succeeds.
+
+Which check to use must be configured externally. Some uses of ESNI, such as local
```
ekr: I would write this as:
"Unless specified by the application using DTLS or externally configured on both sides, implementations MUST use the first method."

chris-wood: That's much better -- thanks!

PR #176, opened 2019-07-24
Clarify that ESNIContents is not transmitted

dmcardle:
I got a little tripped up on this point! Maybe you have a better way to explain it.


chris-wood:
Thanks, @dmcardle!

PR #185, opened 2019-10-16
Remove partial split mode

bemasc:
Based on comments in Montreal, I believe that the authors
no longer intend to document the partial-split configuration.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -149,14 +149,9 @@ Client <------------------------------------>|                     |
 
 In Split Mode, the provider is *not* the origin server for private
 domains. Rather the DNS records for private domains point to the provider,
-but the provider's server just relays the connection back to the
+and the provider's server relays the connection back to the
 backend server, which is the true origin server. The provider does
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -149,14 +149,9 @@ Client <------------------------------------>|                     |
...
-but the provider's server just relays the connection back to the
+and the provider's server relays the connection back to the
 backend server, which is the true origin server. The provider does
-not have access to the plaintext of the connection. In principle,
-the provider might not be the origin for any domains, but as
-a practical matter, it is probably the origin for a large set of
-innocuous domains, but is also providing protection for some private
-domains. Note that the backend server can be an unmodified TLS 1.3
-server.
+not have access to the plaintext of the connection.
```
chris-wood: Is this always true? Might the provider have access in shared mode?

bemasc: This paragraph is only about split mode (and this sentence is unchanged).

chris-wood: Oh, hah, yes. That’s what I get for skimming.

PR #186, opened 2019-10-23
Proposal: Use a hash instead of padding

bemasc:
This is (almost) entirely compatible with wildcard names
and avoids the need for padding.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -216,8 +216,8 @@ SNI Encryption keys can be published using the following ESNIRecord structure.
         opaque public_name<1..2^16-1>;
         KeyShareEntry keys<4..2^16-1>;
         CipherSuite cipher_suites<2..2^16-2>;
-        uint16 padded_length;
         Extension extensions<0..2^16-1>;
+        uint8 label_limit;
     } ESNIKeys;
 
     struct {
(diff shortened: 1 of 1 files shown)


davidben:
I do not think this change is worth the complexity. First, it does not avoid the need for padding. The server certificate needs to be padded regardless, so the server already needs to make decisions based on the length distributions of known hostnames.

Second, it dramatically increases the complexity of deploying ESNI. In the existing design, the actual certificate lookup on the server is relatively unchanged. The TLS library could even abstract much of the name lookup, public name fallback, etc., from the application by simply passing the name along the same interface as before and leave that portion unchanged. This design forces the server to have a new digest-based path, which extends throughout the entire configuration pipeline.

The `label_limit` notion for dealing with wildcards is also another source of stateful deployment problems with configuration mismatches. Perhaps I used to use a wildcard cert, but now I'd like to switch to individual subdomains. This transition is difficult under this PR. For more fun, the folks managing the names may be different from the folks who implemented ESNI, so the deployment concern will go unnoticed.

The wildcard cert scheme is also messy with protocols that do not have their own in-band vhosting mechanism and just use TLS. HTTP doesn't care about HTTP has a Host header, so it doesn't actually care about the SNI. (I will observe, however, that the Host header should probably trigger padding too, so we're still back to square one. :-) )

bemasc:
Yep, deployment is more complex in this mode.  I'm going to close this to avoid creating ongoing confusion; I just wanted to remind people that there are other options beyond debates about padding length, and we shouldn't be too quick to adopt a leaky padding scheme for the sake of handshake size.

Issue #179, opened 2019-07-25
Decouple ESNIKeys and the envelope DNS record


chris-wood:
This will let us adopt another delivery mechanism (HTTPSVC?) if needed later on.

ghedo:
We already have `ESNIKeys` and `ESNIRecord` as different structures, wouldn't a different delivery mechanism be able to simply wrap `ESNIKeys` in its own transport, without the DNS record? I guess this point could be clarified though.

chris-wood:
This is mostly an editorial change to make it easy to remove `ESNIRecord`, if needed, without changing any other text.

chris-wood:
Fixed by #187.

Issue #139, opened 2019-03-01
Associate the ESNI record with the client-facing server IP


jb-wisemo:
To maximize the cover range for clients and simplify management of the ESNI records, it is more useful to associate the ESNI records with the server rather than the query domain name.  This has the additional benefit of allowing different client-facing servers to use different rotating private DH keys.

This eliminates the issues with CNAMEs, individual query domain holders having to make changes, DNS namespace issues etc.

For example a large hoster wishing to provide cover for some clients by having their primary / CDN front ends offer ESNI for all requests can then simply enable the code on their server then add the ESNI records.  They can even do this gradually over a large pool of servers, thus having some servers with different (or no) ESNI support during each careful rollout of ESNI software upgrades.

In practice the following conventions would be used:

1. Any ESNI server IP must have exactly one PTR record (typically pointing to the hoster domain, such as a typical systematic CDN node DNS name).  Lets call this name ptrnam.dom

2. The ESNI records would be under _esni.ptrnam.dom

3. As with most PTR records ptrnam.dom is expected to have exactly one AAAA or A record pointing back at the same IP (if the starting IP is IPv6 only AAAA need to satisfy this, if IPv4 only A, the same ptrnam.dom can be used for one IPv4 and one IPv6 address if the ESNI records would be the same).

4. For address-less server types such as tor hidden services, the ESNI records would be published via whatever DNS or DNS-like mechanism is used for that transport, or queried by sending a DNS query for the record to the target, allowing the target to respond within that transport.

Benefits:

- B1. Large client-facing servers such as CDNs and shared IP hosters can deploy ESNI without having to negotiate DNS changes with all their customers (in fact with any of their customers).

- B2. Operators of large pools of client-facing servers such as CDNs can roll out ESNI support and ESNI implementation upgrades in a piecemeal and controlled manner similar to how other system updates are rolled out.  Simply because each IP address will have its own ESNI records.

- B3. ESNI key rotation becomes much simpler as there is no need to synchronize it between different servers in a pool (except where a pure layer 1/2/3 load balancer shares a single IP among multiple physical/virtual servers, such as the layer 2 HALinux scheme).

- B4. ESNI private key protection can be much stronger when it is not shared among all the servers.  The private key may be kept in protected hardware and/or in volatile RAM.  The latter case would require clients to fall back to a different A or AAAA record until the DNS entries for now forgotten keys to have expired, but only in case of a server crash or similar event.

[...]

PR #187, opened 2019-10-25
Replace ESNIKeys with ESNIConfig, and delegate the publishing container to HTTPSVC.

chris-wood:
DNSOP adopted HTTPSSVC recently [1]. This is a generalized Resource Record that supports the same functionality as the custom ESNI RRType, including direct IP address specification and ENSI key delivery. This change drops the custom DNS bits from the ESNI document and delegates this to HTTPSSVC, where it probably belongs. This change also renames ESNIKeys to ESNIConfig to more accurately capture its intended purpose, i.e., as a struct for conveying ESNI keys and metadata.

This should address #180, #179, #139, and #110.

[1] https://mailarchive.ietf.org/arch/msg/dnsop/9zCxhCfIhDzA3Cv3D2k4NF_nj4s

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -164,7 +164,7 @@ SNI encryption requires that each provider publish a public key and metadata whi
 is used for SNI encryption for all the domains for which it serves directly or
 indirectly (via Split Mode). This document defines the format of the SNI encryption
 public key and metadata, referred to as an ESNI configuration, and delegates DNS
-publication details to {{!I-D.nygren-dnsop-svcb-httpssvc}}, though other delivery
+publication details to {{!HTTPSSVC=I-D.nygren-dnsop-svcb-httpssvc}}, though other delivery
 mechanisms are possible. In particular, if some of the clients of a private
 server are applications rather than Web browsers, those applications might have
 the public key and metadata preconfigured.
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -158,13 +158,14 @@ innocuous domains, but is also providing protection for some private
 domains. Note that the backend server can be an unmodified TLS 1.3
 server.
 
-
 ## SNI Encryption
 
-First, the provider publishes a public key and metadata which is used for SNI
-encryption for all the domains for which it serves directly or indirectly (via
-Split Mode). This document defines a publication mechanism using DNS, but other
-mechanisms are also possible. In particular, if some of the clients of a private
+SNI encryption requires that each provider publish a public key and metadata which
+is used for SNI encryption for all the domains for which it serves directly or
+indirectly (via Split Mode). This document defines the format of the SNI encryption
+public key and metadata, referred to as an ESNI configuration, and delegates DNS
+publication details to {{!I-D.nygren-dnsop-svcb-httpssvc}}, though other delivery
+mechanisms are possible. In particular, if some of the clients of a private
 server are applications rather than Web browsers, those applications might have
 the public key and metadata preconfigured.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -158,13 +158,14 @@ innocuous domains, but is also providing protection for some private
...
 
-First, the provider publishes a public key and metadata which is used for SNI
-encryption for all the domains for which it serves directly or indirectly (via
-Split Mode). This document defines a publication mechanism using DNS, but other
-mechanisms are also possible. In particular, if some of the clients of a private
+SNI encryption requires that each provider publish a public key and metadata which
+is used for SNI encryption for all the domains for which it serves directly or
+indirectly (via Split Mode). This document defines the format of the SNI encryption
+public key and metadata, referred to as an ESNI configuration, and delegates DNS
+publication details to {{!I-D.nygren-dnsop-svcb-httpssvc}}, though other delivery
```
```suggestion
publication details to {{!HTTPSSVC=I-D.nygren-dnsop-svcb-httpssvc}}, though other delivery
```

ekr:
I'm not sure that this is going to work with HTTPSSVC as currently defined. The current ESNIKeys record specifies the IP address explicitly and you are supposed to favor that. However, assuming I am reading HTTPSSVC S 6.4 correctly, ipv[46]hint are not in fact favored but are actually disfavored.

  "The "ipv4hint" and "ipv6hint" keys represent IP address hints for the
   service.  If A and AAAA records for SvcDomainName are locally
   available, the client SHOULD ignore these hints.  Otherwise, clients
   MUST perform A and/or AAAA queries for SvcDomainName as in Section 3,
   and clients SHOULD switch to an IP address in those records as soon
   as possible."

This is counter to what we agreed upon for ESNI and resurrects the A/AAAA vs. ESNIKeys mismatch problem. 

Unless I am misunderstanding this, we should first fix HTTPSSVC before making this change.

davidben:
I think the intent with HTTPSSVC is that, if you're using the HTTPSSVC record, you'd look up A/AAAA from the SvcDomainName not the origin name you queried. That way there's still a binding between A/AAAA and ESNIKeys.

This does, however, cost an extra DNS round-trip, so this plan does require the recursive resolvers follow HTTPSSVC for you like they do CNAME, otherwise it would not meet the performance requirements. DoT/DoH are an ecosystem reset, so maybe that's plausible? I think this goes back to how sacred you believe A and AAAA are and whether you believe it's rude to get IP addresses elsewhere. The HTTPSSVC record evidently has different views on this point than the ESNI record.

draft-ietf-tls-esni.md
```diff
@@ -438,7 +294,7 @@ structure:
        ServerESNIResponseType response_type;
        select (response_type) {
            case esni_accept:        uint8 nonce[16];
-           case esni_retry_request: ESNIKeys retry_keys<1..2^16-1>;
+           case esni_retry_request: ESNIConfig retry_keys<1..2^16-1>;
```
davidben: Nit: s/retry_keys/retry_configs/

davidben: Benefit of the ESNIConfig rename: really `retry_keys` should have been `retry_keyses`, but now we don't have to pretend to be Gollum. :-)

chris-wood:
> This is counter to what we agreed upon for ESNI and resurrects the A/AAAA vs. ESNIKeys mismatch problem.

I would choose to interpret their "SHOULD ignore" advice as optional to ensure the ESNIRecord functionality is maintained. @bemasc, perhaps we can clarify this in the HTTPSVC draft?

bemasc:
> I think the intent with HTTPSSVC is that, if you're using the HTTPSSVC record, you'd look up A/AAAA from the SvcDomainName not the origin name you queried. That way there's still a binding between A/AAAA and ESNIKeys.

Yep, that's exactly right.

> I would choose to interpret their "SHOULD ignore" advice as optional to ensure the ESNIRecord functionality is maintained.

I'm not sure I understand you.  The purpose of "SHOULD ignore" is to ensure that DNS-based load balancing of A/AAAA on SvcDomainName continues to work correctly.  Whether the client uses an ipv6hint or a AAAA record for SvcDomainName, it will be contacting a server that has the corresponding ESNI private key; there's no possibility of mismatch either way.

> perhaps we can clarify this in the HTTPSVC draft?

I think the confusion here is that in HTTPSSVC, clients use the IP of SvcDomainName, not the IP of the origin hostname.  There's no way for a client to accidentally combine the ESNIConfig for the SvcDomainName with the IP of the origin hostname (or of a different SvcDomainName).

I think the draft is pretty clear on this point (e.g. the [intro example](https://tools.ietf.org/html/draft-nygren-dnsop-svcb-httpssvc-00#section-1.1)), but the authors always think that :).  Suggestions are welcome.

ekr:
Well, so as @davidben points out, this requires a second resolution, which is a performance problem.

bemasc:
Yes, waiting for a second resolution is the performance problem that ipv6hint solves, by giving the client an IP address along with the ESNIConfig in the same response.

ekr:
@bemasc: then I just don't see the point of doing the second resolution. It seems disruptive and doesn't add anything. Once I have connected to the server, it can Alt-Svc me itself.

ekr:
I should add, I'm not quite sure how I would implement this in Firefox. Say I resolve example.com and get 1.2.3.4 for ipv4 hints. Now I am supposed to in parallel connect to the server *and* try to resolve SvcDomainName? But what then? I've already started the connection so I'm not going to abort it if I get a 5.6.7.8....

bemasc:
> I just don't see the point of doing the second resolution. It seems disruptive and doesn't add anything.

Consider a CDN customer with their own (old-fashioned, static) DNS.  The customer publishes an HTTPSSVC record whose SvcDomainName, ESNIConfig, and IP hint are the CDN's domain, ESNIConfig, and IP.  Clients connect to the IP from the hint, use this ESNIConfig, and everything works ... but the customer has broken the CDN's load balancing.  All of their traffic is going to a single  CDN IP.

The point of the second resolution is to move customer traffic onto the CDN's preferred IP.

> Once I have connected to the server, it can Alt-Svc me itself.

True, but HTTPSSVC doesn't require clients to support Alt-Svc.  (Chrome, for example, does not support this feature of Alt-Svc.)

> Say I resolve example.com and get 1.2.3.4 for ipv4 hints. Now I am supposed to in parallel connect to the server and try to resolve SvcDomainName? But what then? I've already started the connection so I'm not going to abort it if I get a 5.6.7.8....

The draft says clients "SHOULD switch to an IP address in those records as soon as possible".  An aggressive implementation might start connecting to 5.6.7.8, and close the connection to 1.2.3.4 once it is connected.  (This is roughly what Chrome does for QUIC upgrade.)  A less aggressive implementation might just let the A record sit in the local DNS cache, so that any subsequent connections to this origin can go to the CDN's preferred IP.

ekr:
OK, so I think that this text needs a bunch of work before I would be comfortable with it, as I do not think your "less aggressive implementation" would be conformant with this SHOULD

chris-wood:
@ekr if that’s the case, can you please suggest concrete changes to the text?

ekr:
@bemasc is there some place I can send a PR?

bemasc:
@ekr yes, https://github.com/MikeBishop/dns-alt-svc/pulls

ekr:
https://github.com/MikeBishop/dns-alt-svc/pull/69

ekr:
OK, the new text has been merged, so I am now OK with this PR.

PR #196, opened 2019-11-05
Tunnel version

ekr:


davidben:
I haven't looked at this in full detail yet, but one initial comment is that encrypting an entire ClientHello means doubling the size, and [post-quantum key shares](https://www.imperialviolet.org/2019/10/30/pqsivssl.html) are looking to be quite large. I believe [one of our experiments](https://www.imperialviolet.org/2018/04/11/pqconftls.html) suggested the additional cost is roughly linear to the number of bytes.

We could encrypt some sort of "delta ClientHello", either saying "copy these extensions from the outside" or something hilarious scheme where you specify a sequence of new bytes and references into your outer ClientHello... (I think the former is easier to send while the latter is easier to receive.)

Another thought, though it requires a side channel from frontend to backend, is the `ClientHelloInner` omits `key_shares` in favor of a new `key_shares_hashed` extension where you insert `SHA256(ClientHelloOuter.key_shares)` for transcript purposes. The receiver then asserts the hash matches and pretends you included `ClientHelloOuter.key_shares` in `ClientHelloInner` for handshake purposes.

ekr:
Yeah, one big drawback here is clearly the size, and yeah, PQ makes that worse.

With that said, it seems like you could potentially ameliorate this by only including classical KeyShares in ClientHelloOuter. The rationale here would be that the only purpose of ClientHelloOuter is to enable the fallback handshake, and so those keys are just protecting the SNI (and ESNIConfig, but that's just there to protect SNI), and we could potentially live with that being non-quantum-safe, especially if we're not going to use PQ algorithms for ESNI encryption. And of course if we *do* use PQ algorithms for ESNI encryption, then the overhead of the second CH is only about 1/2 as opposed to double.

davidben:
Wouldn't that would negate the point of GREASE? ClientHellos advertising ESNI would have large ciphertext and classical-only outer key shares, while ClientHellos not advertising ESNI would include the PQ outer key shares (and probably send a small GREASE extension ton compensate).

ekr:
Yes, that's true. I think I'd be willing to live with that tradeoff but
perhaps others would not.

martinthomson:
The size-doubling issue could be addressed through the use of compression.  If the inner and outer values are the same, then the inner could just reference the outer (and the removal of encryption could also include expansion).

kazuho:
By compression (or reusing stuff), doesn't the proposal become closer to the ESNI transformation design that we have been talking in https://github.com/kazuho/draft-rescorla-tls-esni/pull/1?

davidben:
@martinthomson Yeah, compression was what I was referring with "delta ClientHello". You could imagine a ClientHello-specific compression that knows how to toss extensions around (in which case, yeah, it starts getting closer to ESNI transformation), or some more generic delta encoding. Here's an ad hoc goofy thing:

```
enum { bytes(0), reference(1) } FragmentType;
 
struct {
  uint24 offset;
  uint24 length;
} ReferenceFragment;
 
opaque BytesFragment<1..2^24-1>;
 
struct {
  FragmentType type;
  select (Fragment.type) {
    // Append these bytes.
    case bytes: BytesFragment;
    // Append the referenced fragment of the outer ClientHello.
    case reference: ReferenceFragment;
  }
} Fragment;
 
struct {
  // Included so preallocating the buffer and enforcing size limits is easy.
  uint24 clienthello_length;
  // Walk over all the fragments to reconstruct the ClientHello.
  Fragment clienthello_fragments<0..2^24-1>;
} ESNIPlaintext;
```

A ClientHello-specific encoding is easier to produce since senders know how their ClientHellos look like and we can tailor things what fields should and shouldn't be copied. But it's harder to decode since you need to add checks for silly questions like "what order do I insert extensions?" or "what happens if the placeholder is missing" or "what happens if there are two placeholders?" or "what happens if the outer ClientHello is missing the extension to copy?".

Conversely, though it may take more spec text to describe, a bytes-based encoding is easier to decode since you just blindly evaluate the two opcodes and only need to include a couple of checks. But it's harder to encode since you need to account for length prefixes when chopping things up.

davidben:
@ekr pointed out that delta-encoding tricks should include the hash of the reconstructed ClientHello, otherwise the ciphertext tag does not cover the key shares, etc., so the above needs one more field.

draft-ietf-tls-esni.md
```diff
@@ -272,35 +280,21 @@ of the structure.  This hash is computed using the hash function
...
-For servers (in EncryptedExtensions), this extension contains the following
-structure:
 
-~~~
-   enum {
-       esni_accept(0),
-       esni_retry_request(1),
-   } ServerESNIResponseType;
+If the server accepts ESNI, then does not send this extension.
+If it rejects ESNI, then it sends the following structure in
```
kazuho: This paragraph talks about sending a ESNIKeys that the client should use for retrying.

Therefore, maybe s/If it rejects ESNI/If the client sent an ESNI "encrypted_client_hello" extension using a key not available to the server/.

draft-ietf-tls-esni.md
```diff
@@ -707,7 +678,7 @@ is not interoperable with existing servers, which expect the value in
 the existing cleartext extension. Thus server operators SHOULD ensure
 servers understand a given set of ESNI keys before advertising them.
 Additionally, servers SHOULD retain support for any
-previously-advertised keys for the duration of their validity.
+previously-advertised keys for the duration of their validity
```
kazuho: Have you dropped the period by an accident?

draft-ietf-tls-esni.md
```diff
@@ -347,148 +408,92 @@ computed from Z as follows:
...
+to harmonize these to make sure that we maintain key separation.]]
 
-dns_name
-: The true SNI DNS name, that is, the HostName value that would have been sent in the
-plaintext "server_name" extension. (NameType values other than "host_name" are
-unsupported since SNI extensibility failed {{SNIExtensibilityFailed}}).
+Finally, the client MUST generate a ClientHelloOuter message
+containing the "encrypted_client_hello" extension with the values as
+indicated above. The client MUST place the value of
+ESNIConfig.public_name in the "server_name" extension. The remaining
```
sayrer: The draft -02 protocol seems to work without an unencrypted SNI. Is this adding a required unencrypted SNI?

draft-ietf-tls-esni.md
```diff
@@ -347,148 +408,92 @@ computed from Z as follows:
...
-   struct {
-       opaque dns_name<1..2^16-1>;
-       opaque zeros[ESNIConfig.padded_length - length(dns_name)];
-   } PaddedServerNameList;
-
-   struct {
-       uint8 nonce[16];
-       PaddedServerNameList realSNI;
-   } ClientESNIInner;
+    encrypted_sni = AEAD-Encrypt(key, iv, "", ClientHelloIInner)
```
sayrer: Double capital I in "ClientHelloIInner"

draft-ietf-tls-esni.md
```diff
@@ -155,21 +155,30 @@ not have access to the plaintext of the connection.
...
-server are applications rather than Web browsers, those applications might have
-the public key and metadata preconfigured.
+SNI encryption requires that each provider publish a public key and
+metadata which is used for SNI encryption for all the domains for
+which it serves directly or indirectly (via Split Mode). This document
+defines the format of the SNI encryption public key and metadata,
+referred to as an ESNI configuration, and delegates DNS publication
+details to {{!HTTPSSVC=I-D.nygren-dnsop-svcb-httpssvc}}, though other
+delivery mechanisms are possible. In particular, if some of the
+clients of a private server are applications rather than Web browsers,
```
sayrer: nit: Both the AP style guide and https://tools.ietf.org/html/rfc7322 no longer capitalize "web"

draft-ietf-tls-esni.md
```diff
@@ -210,13 +219,12 @@ keys
...
-expects to support rounded up the nearest multiple of 16. If the
-server supports arbitrary wildcard names, it SHOULD set this value to
-260. Clients SHOULD reject ESNIConfig as invalid if padded_length is
-greater than 260.
+maximum_name_length
+: the largest name the server expects to support
+rounded up the nearest multiple of 16. If the server supports
+arbitrary wildcard names, it SHOULD set this value to
+256. Clients SHOULD reject ESNIConfig as invalid if maximum_name_length is
+greater than 256.
```
sayrer: "reject an ESNIConfig"?

I'm still not clear why there's server-controlled padding here. I can see using this figure as a hint, though.

draft-ietf-tls-esni.md
```diff
@@ -309,33 +303,100 @@ This protocol also defines the "esni_required" alert, which is sent by the
...
 Otherwise, it is a "non-ESNI PSK". This may be implemented by adding a new field
 to client and server session states.
 
-## Client Behavior {#client-behavior}
+## Incorporating Outer Extensions {#outer-extensions}
+
+Some TLS 1.3 extensions can be quite large
+and having them both in the inner and outer ClientHello wil lead to
+a very large overall size. One particularly pathological example
+is "key_share" with post-quantum algorithms. In order to reduce
```
sayrer: May want to cite draft-hoffman-c2pq or a successor draft here.

draft-ietf-tls-esni.md
```diff
@@ -309,33 +303,100 @@ This protocol also defines the "esni_required" alert, which is sent by the
...
+(this is a violation of normal TLS rules, but the resulting ClientHelloInner
+is never processed directly). However, there MUST NOT be
+multiple "outer_extension" extensions with the same extension code point.
+
+
+# Client Behavior {#client-behavior}
+
+## Sending an encrypted ClientHello {#send-esni}
+
+In order to send an encrypted SNI, the client MUST first generate its
```
sayrer: I've noticed that this draft uses "encrypted SNI" and "ESNI" interchangeably, but unpredictably. Would recommend taking a pass over all of these acronyms once the naming is settled ("ECHO" etc...)

draft-ietf-tls-esni.md
```diff
@@ -347,148 +408,92 @@ computed from Z as follows:
...
-       Random client_hello_random;
-   } ESNIContents;
-~~~
-
-record_digest
-: Same value as ClientEncryptedSNI.record_digest.
-
-esni_key_share
-: Same value as ClientEncryptedSNI.key_share.
+The client MAY replace any large, duplicated, extensions in ClientHelloInner
```
sayrer: I think the commas are off here. I think the best fix might be to leave out "large" and write "duplicated extensions".

draft-ietf-tls-esni.md
```diff
@@ -552,11 +546,32 @@ trigger retries, as described in {{handle-server-response}}. This may be
 implemented, for instance, by reporting a failed connection with a dedicated
 error code.
 
-### GREASE extensions {#grease-extensions}
+### HelloRetryRequest
+
+If the server sends a HelloRetryRequest in response to the ClientHello
+and the client can send a second updated ClientHello per the rules in
+{{RFC8446}}. At this point, the client does not know whether the
```
sayrer: Incomplete sentence.

draft-ietf-tls-esni.md
sayrer: Just to add to the seemingly endless traffic identification via packet size: https://isc.sans.edu/diary/25616

sftcd:
While waiting for this or some similar PR to be merged, I played about a bit with implementing guesses as to how it might be done in a [branch](https://github.com/sftcd/openssl/tree/encch) of my ESNI-enabled OpenSSL fork. (If interested here's the [diff](https://github.com/sftcd/openssl/compare/master...sftcd:encch))

Doing that was fairly messy but not that hard, the main thoughts/questions I was left with were:
- it'd maybe be easier to have the ESNI-specific padding in the esni-nonce extension directly as that'd mean less change to OpenSSL API behaviour (applications can set callbacks to do padding themselves too)
 - I took a stab at using my OpenSSL code for [HPKE](https://github.com/sftcd/happykey) instead of the ESNI-specific stuff that's been in earlier drafts and that wasn't a problem (though I didn't spend time thinking about how exactly to use HPKE - I just set the TLS session key share as the AAD)
- I'm still very unclear as to what rules should/must apply to the relationship between values in the inner and outer CH's - but the simpler we can make that the better (I don't mean compression here, just what rules are to apply when we have two versions of a thing, like the TLS session key share or ALPN for example)

And just in case: I didn't try any compression and this code is likely flakey as hell in error cases, so don't anyone go using it for anything:-) I plan to ditch it anyway once draft-06 is out.

chris-wood:
Closing in favor of #207.

PR #199, opened 2019-11-05
Unify version/extension and key share negotiation mechanisms

davidben:
[This PR sits on top of #198. Ignore the first commit to look at it in isolation.]

We currently have two levels of mechanisms for servers to advertise multiple sets of parameters. First, each ESNIConfig has multiple keys in it, so a server can suport multiple groups at once. These keys are stapled together so a single HTTPSSVC record, etc., delivers the entire configuration.

At a layer above that, we have a version/extension-based mechanism. The server may advertise multiple ESNIConfigs to cover multiple versions of ESNI (hopefully we won't need this mechanism after standardization, but it's quite useful for draft versions) or multiple sets of mandatory extensions. Unlike the group advertisement, these are each in separate configuration units, so, e.g., they may be in multiple HTTPSSVC records.

This PR flattens the two mechanisms into one. Each ESNIConfig now has a single KeyShareEntry. Multiple groups and multiple versions are expressed in the same way: multiple ESNIConfig structures. It then introduces a wrapper ESNIConfigs structure which is the new unit of ESNI configuration delivered in HTTPSSVC, etc.

This should be paired with a change to draft-nygren-dnsop-svcb-httpssvc to carry an ESNIConfigs, not an ESNIConfig (formerly ESNIKeys). (@bemasc FYI)

This fixes a number of oddities:

- Client and server need to maintain two levels of configuration. On the server, the structure has two layers of lists, something like list(tuple(ESNIConfig, list(PrivateKey))). Now the server maintains list(tuple(ESNIConfig, PrivateKey)). Conversely, the client needed two negotiation loops whereas now it only requires one.

- Logic across different layers of the client stack now more reliably have the full ESNI configuration available in one unit.

  In the old design, a single ESNIConfig, rather than a set of them, would be correlated with an IP address or CNAME, by way of the HTTPSSVC record. A server which supportsed multiple ESNI versions or mandatory extension sets would advertise several HTTPSSVC records, potentially with different IPs. That means the DNS and transport connection logic must know about TLS capabitilies to pick the right HTTPSSVC record.

  In the new design, the server includes all its variations at once, so the DNS and transport logic can reasonably pick an HTTPSSVC purely based on whether the TLS stack supports ESNI, rather than the full set of capabilities. That is, the text in [0] does not work right now, and this PR fixes it.

The downside is that, in the likely case that the server only supports one version and one set of extensions, the non-key ESNI config bits (notably the public name; the rest is small) are repeated for every named group. However, the public name should not be that large and servers should only need one or two groups[1].

This also does make ESNIConfigs and ESNIConfig separate structures, which is slightly goofy. Alternate names if folks prefer:

- ESNIConfig/ESNIConfigs => ESNIConfig/ESNIBundle
- ESNIConfig/ESNIConfigs => ESNIKey/ESNIConfig

[0] https://tools.ietf.org/html/draft-nygren-dnsop-svcb-httpssvc-00#section-8.1.1
[1] https://www.imperialviolet.org/2016/05/16/agility.html

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -185,12 +185,16 @@ ESNIConfig structure.
 
     struct {
         uint16 version;
+        opaque contents<1..2^16-1>;
+    } ESNIConfig;
+
+    struct {
         opaque public_name<1..2^16-1>;
         KeyShareEntry keys<4..2^16-1>;
         CipherSuite cipher_suites<2..2^16-2>;
         uint16 padded_length;
         Extension extensions<0..2^16-1>;
-    } ESNIConfig;
+    } ESNIConfigContents;
 ~~~~
 
 The ESNIConfig structure contains the following fields:
@@ -200,6 +204,12 @@ version
 SHALL be 0xff03. Clients MUST ignore any ESNIConfig structure with a
 version they do not understand.
 [[NOTE: This means that the RFC will presumably have a nonzero value.]]
+
+contents
+: An opaque byte string whose contents depend on the version of the structure.
+For this specification, the contents are an ESNIConfigContents structure.
+
+The ESNIConfigContents structure contains the following fields:
 
 public_name
 : The non-empty name of the entity trusted to update these encryption keys.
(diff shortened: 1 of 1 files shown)


ekr:
It seems to me that there are two changes here:

1. Bundling all the ESNIConfigs into a single HTTPSVC record. This seems like a good idea for the reasons you lay out.

2. Flattening the structure. I'm less enthusiastic about this: it requires duplicating data and is a new opportunity to have inconsistencies as well as bloat.

davidben:
That's a good point. It's late here, so I will see about splitting this into two PRs tomorrow.

davidben:
Split this into #200 and #201.

PR #200, opened 2019-11-06
Advertise multiple ESNIConfigs as a unit

davidben:
[This was split out of #199 as the hopefully less controversial idea. Also it sits on top of #198. Ignore the first commit to look at it in isolation.]

We have two levels of mechanisms for servers to advertise parameters. Each ESNIConfig has multiple keys, so a server can suport multiple groups at once. These keys are stapled together so a single HTTPSSVC record delivers the entire configuration.

Then we have a versions/extensions mechanism. A server supporting multiple versions of ESNI or multiple sets of mandatory extensions must advertise multiple ESNIConfigs. Unlike the group advertisement, these require separate HTTPSSVC records. This means the full ESNI config is not reliably available to logic across different layers of the client stack. In particular, the HTTPSSVC spec text in [0] does not work. (@bemasc FYI)

Currently, a single ESNIConfig, rather than a set of them, is correlated with an IP address or CNAME, by way of the HTTPSSVC record. A server which supports multiple ESNI versions or mandatory extension sets must advertise several HTTPSSVC records, potentially with different IPs. DNS and transport logic must know TLS capabilities to pick the right HTTPSSVC record.

This PR introduces an ESNIConfigs structure which advertises a list of ESNIConfig structures. The ESNIConfigs is the new unit of ESNI configuration. This should be paired with a change to draft-nygren-dnsop-svcb-httpssvc to carry an ESNIConfigs, not an ESNIConfig (formerly ESNIKeys).

Now the server can include the full config at once, so the DNS and transport logic can reasonably pick an HTTPSSVC purely based on whether the TLS stack supports ESNI, rather than the full set of capabilities. This fixes the text in [0].

That ESNIConfigs and ESNIConfig are separate structures is slightly goofy naming. Alternate names if folks prefer:

- ESNIConfig/ESNIConfigs => ESNIConfig/ESNIBundle

- ESNIConfig/ESNIConfigs => ESNIKeys/ESNIConfig
  (Nuisance: an ESNIConfig is a list of ESNIKeyses, which brings back the double plural.)

(While I'm here, fix the remnants of ESNIKeyShares. I'll split that out if this PR is controversial, but GitHub is really bad at dependent PRs, so I've left it together for now.)

[0] https://tools.ietf.org/html/draft-nygren-dnsop-svcb-httpssvc-00#section-8.1.1

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -185,12 +185,16 @@ ESNIConfig structure.
 
     struct {
         uint16 version;
+        opaque contents<1..2^16-1>;
+    } ESNIConfig;
+
+    struct {
         opaque public_name<1..2^16-1>;
         KeyShareEntry keys<4..2^16-1>;
         CipherSuite cipher_suites<2..2^16-2>;
         uint16 padded_length;
         Extension extensions<0..2^16-1>;
-    } ESNIConfig;
+    } ESNIConfigContents;
 ~~~~
 
 The ESNIConfig structure contains the following fields:
@@ -200,6 +204,12 @@ version
 SHALL be 0xff03. Clients MUST ignore any ESNIConfig structure with a
 version they do not understand.
 [[NOTE: This means that the RFC will presumably have a nonzero value.]]
+
+contents
+: An opaque byte string whose contents depend on the version of the structure.
+For this specification, the contents are an ESNIConfigContents structure.
+
+The ESNIConfigContents structure contains the following fields:
 
 public_name
 : The non-empty name of the entity trusted to update these encryption keys.
@@ -390,7 +400,7 @@ The client then creates a ClientESNIInner structure:
 ~~~~
    struct {
        opaque dns_name<1..2^16-1>;
-       opaque zeros[ESNIConfig.padded_length - length(dns_name)];
+       opaque zeros[ESNIConfigContents.padded_length - length(dns_name)];
    } PaddedServerNameList;
 
    struct {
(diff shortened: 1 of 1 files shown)


bemasc:
> This means the full ESNI config is not reliably available to logic across different layers of the client stack. In particular, the HTTPSSVC spec text in [0] does not work. (@bemasc FYI)

Why doesn't it work?  The transport side tries the highest-priority option, and the TLS library is free to reject it, in which case the transport side will try the next option, etc.  All the transport side needs to know is whether the TLS library supports ESNI.

Are you thinking about a case where the client only supports ESNI version 1, and the server only offers ESNI version 2?

davidben:
That presumes your retry loop is integrated up and down the layers. It also means that, unless you shortcircuit things and check ESNI ahead of time, you pay for a TCP roundtrip to the unsupported ESNIConfig addresses. It also doesn't account for cases when the client doesn't support any of the ESNI versions. In that case, it should connect without ESNI.

bemasc:
> That presumes your retry loop is integrated up and down the layers. It also means that, unless you shortcircuit things and check ESNI ahead of time, you pay for a TCP roundtrip to the unsupported ESNIConfig addresses.

OK, I agree with all of this.  I'm happy to reduce the required level of integration, if we can find a way.

> It also doesn't account for cases when the client doesn't support any of the ESNI versions. In that case, it should connect without ESNI (presuming it would be okay doing so without ESNI configs).

Are you proposing that the TLS library should silently ignore an ESNIConfigs with all unsupported versions/ciphers, and send the SNI in cleartext?  That seems like a recommendation that should go in this draft.  Or are you proposing a fallback check that goes back to the transport layer?

chris-wood:
Discussed in Singapore; no objections. Will merge.

PR #208, opened 2020-03-12
Align client GREASE section with ClientEncryptedCH definition.

dmcardle:
Also define the length of the randomly-generated "encrypted_ch" field.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -624,10 +624,7 @@ If the client attempts to connect to a server and does not have an ECHOConfig
 structure available for the server, it SHOULD send a GREASE
 {{I-D.ietf-tls-grease}} "encrypted_client_hello" extension as follows:
 
-- Select a supported cipher suite, named group, and padded_length
-  value. The padded_length value SHOULD be 260 (sum of the maximum DNS name
-  length and TLS encoding overhead) or a multiple of 16 less than 260.
-  Set the "suite" field  to the selected cipher suite. These selections
+- Set the "suite" field  to a supported cipher suite. These selections
   SHOULD vary to exercise all supported configurations, but MAY be held constant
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -631,15 +631,15 @@ structure available for the server, it SHOULD send a GREASE
...
   bytes, where hash_length is the length of the hash function associated with
   the chosen cipher suite.
 
-- Set the "encrypted_client_hello" field to a randomly-generated string of
-  [TODO] bytes.
+- Set the "enc" field to a randomly-generated valid encapsulated public key
+  output by the HPKE KEM.
+
+- Set the "encrypted_ch" field to a randomly-generated string of padded_length
+  bytes.
```
dmcardle: Is `padded_length` supposed to be the length for "encrypted_ch"? I was unclear on this.

dmcardle:
@davidben @dvorak42

draft-ietf-tls-esni.md
chris-wood: We should probably table this until we figure out the padding strategy. (Apparently that's still up in the air!) It also occurs to me that the first bullet in this list is obsolete -- we don't have a named group anymore or padded_length anymore:

```
      struct {
          CipherSuite suite;
          opaque record_digest<0..2^16-1>;
          opaque enc<1..2^16-1>;
          opaque encrypted_ch<1..2^16-1>;
      } ClientEncryptedCH;
```

Can we just say something like the following? 

"Set the "suite" field to a supported cipher suite. This selection SHOULD vary to exercise all supported configurations, but MAY be constant for successive connections to the same server in the same session."

dmcardle: Ah, ok. I'll bring the [TODO] back for the length of encrypted_ch.

PR #209, opened 2020-03-14
padding scheme text

sftcd:
I've tried to capture what I think is a reasonable way to handle padding now we've changed from ESNI->ECHO.  If we could do something like this, that'd be better than what's in -06.

Note that I'd be even happier if we entirely eliminated ECHOConfig.minimum_inner_length - I don't think it actually adds any value and it represents yet another way to get a configuration wrong.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -460,7 +460,7 @@ For each field in the inner ClientHello, clients need to determine how much to
 pad given the semantics of that field.  For example, if a client can propose
 various ALPN values, it could add padding to round up to the longest of those.
 
-For most fields in a ClientHello this can be determined without server help.
+The target padding length of most ClientHello extensions can be determined without server help.
 For the server_name, however, if ECHOConfig.maximum_name_length is longer than
 the actual server_name then clients SHOULD add padding to make up that
 difference. If the maximum_name_length is zero or less than the length of the
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -204,7 +204,7 @@ following ECHOConfigs structure.
         HkpeKemId kem_id;
         CipherSuite cipher_suites<2..2^16-2>;
 
-        uint16 maximum_name_length;
+        uint16 minimum_inner_length;
         Extension extensions<0..2^16-1>;
     } ECHOConfigContents;
 
@@ -254,10 +254,11 @@ cipher_suites
 See {{hpke-map}} for information on how a cipher suite maps to corresponding
 HPKE algorithm identifiers.
 
-maximum_name_length
-:  the largest name the server expects to support. If the server supports arbitrary
-wildcard names, it SHOULD set this value to 256. Clients SHOULD reject ESNIConfig as
-invalid if maximum_name_length is greater than 256.
+minimum_inner_length
+:  The minimum (encoded) length ClientHelloInner recommended by the server.
+Sending a ClientHelloInner that is shorter than this will likely stand out.
+Clients can use a padding extension within the ClientHelloInner to ensure
+the plaintext is at leat this long.
 
 extensions
 : A list of extensions that the client can take into consideration when
@@ -444,10 +445,33 @@ values, ClientHelloInner MUST also contain:
  - an "echo_nonce" extension
  - TLS padding {{!RFC7685}}
 
-Padding SHOULD be P = L - D bytes, where
+An encoded ClientHelloInner can be around 300 octets long, but variations in
+the length of the ciphertext version of that could defeat the entire purpose of
+ECHO, if for example those expose the chosen server_name field in the
+ClientHelloInner.  The padding ClientHello extension, if well used, can ensure
+that length information does not expose what is contained in the
+ClientHelloInner. Clients MUST add padding to the ClientHelloInner to try meet
+this requirement.
 
-- L = ECHOConfig.maximum_name_length, rounded up to the nearest multiple of 16
-- D = len(dns_name), where dns_name is the DNS name in the ClientHelloInner "server_name" extension
+Given that new extensions may be defined in future that contain sensitive
+length fields, we cannot describe all the ways in which length information
+could expose sensitive content. Implementers ought therefore be aware that they
+might have to change their padding scheme as the set of supported extensions
+changes. In addition to padding the ClientHelloInner, clients and servers MUST
+both pad all other handshake messages that have sensitive length fields. For
+example, if the client proposes ALPN values in the ClientHelloInner, the
+server-selected value will be returned in an EncryptedExtension, so that
+handshake message needs to be padded with TLS record layer padding. 
+
+One approach that may be relative safe and simple is once ECHO is in use:
+
+- When constructing the ClientHelloInner, use padding bytes so that the length
+  of the plaintext erncoding of ClienHelloInner is greater than
+ECHOConfig.minimum_inner_length and is a multiple of 16
+- Both clients and servers pad all handshake messages so that plaintexts are a
+  multiple of 16 octets long.
+- Clients or servers MAY send additional padding bytes. This MUST NOT be
+  treated as an error.
 
 When offering an encrypted ClientHello, the client MUST NOT offer to resume any
 non-ECHO PSKs. It additionally MUST NOT offer to resume any sessions for TLS 1.2
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -466,7 +466,7 @@ handshake message needs to be padded with TLS record layer padding.
 One approach that may be relative safe and simple is once ECHO is in use:
 
 - When constructing the ClientHelloInner, use padding bytes so that the length
-  of the plaintext erncoding of ClienHelloInner is greater than
+  of the plaintext encoding of ClienHelloInner is greater than
 ECHOConfig.minimum_inner_length and is a multiple of 16
 - Both clients and servers pad all handshake messages so that plaintexts are a
   multiple of 16 octets long.
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -256,10 +256,10 @@ HPKE algorithm identifiers.
 
 maximum_name_length  
 : The largest name the server expects to support, if known.
-If this value is not known (e.g. if wildcard names are in use, or names can be
-added or removed from the anonymity set during the lifetime of a particular
-resource record value), then this value can be set to zero, in which case
-clients SHOULD use the inner ClientHello padding scheme described below.
+If this value is not known it can be set to zero, in which case clients SHOULD
+use the inner ClientHello padding scheme described below.  That could happen if
+wildcard names are in use, or if names can be added or removed from the
+anonymity set during the lifetime of a particular resource record value.
 
 extensions
 : A list of extensions that the client can take into consideration when
@@ -969,12 +969,12 @@ SNI uniformly?]]
 
 ## Padding
 
-Variations in the length of the ciphertext version of the ClientHelloInner
-could defeat the purpose of ECHO, if those expose the chosen server_name field
-in the ClientHelloInner. The padding ClientHello extension, if well used, can
-ensure that length information does not expose what is contained in the
-ClientHelloInner. Clients SHOULD add padding to the ClientHelloInner to meet this
-requirement.
+Variations in the length of the ClientHelloInner ciphertext could defeat the
+purpose of ECHO, if those expose the chosen server_name field in the
+ClientHelloInner. The padding ClientHello extension, if well used, can ensure
+that length information does not expose what is contained in the
+ClientHelloInner. Clients SHOULD add padding to the ClientHelloInner to meet
+this requirement.
 
 [[TODO: once the padding scheme is final, revisit the above.]]
 
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -476,11 +476,11 @@ all per-extension padding values, rounded up to the nearest multiple of 32.
 (This additional rounding step aims to hide variance across different client
 implementations.)
 
-In addition to padding the ClientHelloInner, clients and servers will also need
+In addition to padding ClientHelloInner, clients and servers will also need
 to pad all other handshake messages that have sensitive-length fields. For
-example, if a client proposes ALPN values in the ClientHelloInner, the
+example, if a client proposes ALPN values in ClientHelloInner, the
 server-selected value will be returned in an EncryptedExtension, so that
-handshake message also needs to be padded using TLS record layer padding. 
+handshake message also needs to be padded using TLS record layer padding.
 
 When offering an encrypted ClientHello, the client MUST NOT offer to resume any
 non-ECHO PSKs. It additionally MUST NOT offer to resume any sessions for TLS 1.2
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -446,38 +446,40 @@ values, ClientHelloInner MUST also contain:
  - an "echo_nonce" extension
  - TLS padding {{!RFC7685}}
 
-An encoded ClientHelloInner can be around 300 octets long, but variations in
-the length of the ciphertext version of that could defeat the entire purpose of
-ECHO, if for example those expose the chosen server_name field in the
-ClientHelloInner.  The padding ClientHello extension, if well used, can ensure
-that length information does not expose what is contained in the
-ClientHelloInner. Clients add padding to the ClientHelloInner to try meet
-this requirement.
+Variations in the length of the ciphertext version of the ClientHelloInner
+could defeat the purpose of ECHO, if those expose the chosen server_name field
+in the ClientHelloInner. The padding ClientHello extension, if well used, can
+ensure that length information does not expose what is contained in the
+ClientHelloInner. Clients SHOULD add padding to the ClientHelloInner to meet this
+requirement.
 
 Given that extensions could be defined in the future that reveal sensitive information
 through their length, we cannot describe all the ways in which length information
 could expose sensitive content. Implementers ought therefore be aware that they
 might have to change their padding scheme as the set of supported extensions
-changes. In addition to padding the ClientHelloInner, clients and servers will
-both need to pad all other handshake messages that have sensitive-length fields. For
-example, if the client proposes ALPN values in the ClientHelloInner, the
+changes. 
+
+For each field in the inner ClientHello, clients need to determine how much to
+pad given the semantics of that field.  For example, if a client can propose
+various ALPN values, it could add padding to round up to the longest of those.
+
+For most fields in a ClientHello this can be determined without server help.
+For the server_name, however, if ECHOConfig.maximum_name_length is longer than
+the actual server_name then clients SHOULD add padding to make up that
+difference. If the maximum_name_length is zero or less than the length of the
+actual server_name then round the server_name up to a multiple of 32 octets and
+randomly add another 32 octets 50% of the time and then include that amount
+of additional padding.
+
+Sum all the padding together, then in order to reduce entropy across different
+client implementationss, round up so that the inner ClientHello encoding length 
+is a multiple of 32.
+
+In addition to padding the ClientHelloInner, clients and servers will also need
+to pad all other handshake messages that have sensitive-length fields. For
+example, if a client proposes ALPN values in the ClientHelloInner, the
 server-selected value will be returned in an EncryptedExtension, so that
-handshake message needs to be padded with TLS record layer padding. 
-
-One approach that may be relative safe and simple is once ECHO is in use is for
-each field, determine how much it makes sense to pad given what kinds of things
-it sends. If it varies ALPN, maybe round up to the largest of those. Most of
-this can be determined without server help.  For fields where server help is
-useful, like the server_name, apply that help if available.  
-
-If the maximum_name_length is longer than the actual server_name then clients
-should add padding to make up the difference. If the maximum_name_length is
-zero or less than the length of the actual server_name then round the
-server_name up to a multiple of 32 octets and randomly add another 32 octets
-50% of the time.
-
-Sum all that padding together, then in order to reduce entropy across different
-kinds of clients, round up to a multiple of 32.
+handshake message also needs to be padded using TLS record layer padding. 
 
 When offering an encrypted ClientHello, the client MUST NOT offer to resume any
 non-ECHO PSKs. It additionally MUST NOT offer to resume any sessions for TLS 1.2
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -461,11 +461,14 @@ pad given the semantics of that field.  For example, if a client can propose
 various ALPN values, it could add padding to round up to the longest of those.
 
 The target padding length of most ClientHello extensions can be determined without server help.
-For the server_name, however, if ECHOConfig.maximum_name_length is longer than
-the actual server_name then clients SHOULD add padding to make up that
-difference. If the maximum_name_length is zero or less than the length of the
-actual server_name then round the server_name up to a multiple of 32 octets and
-randomly add another 32 octets 50% of the time and then include that amount
+However, the "server_name" extension could benefit from server input (ECHOConfig.maximum_name_length). 
+Clients SHOULD compute the padding for this extension as follows:
+
+1. If ECHOConfig.maximum_name_length is longer than the actual server_name then 
+clients SHOULD add padding to make up that difference.
+2. Otherwise, if ECHOConfig.maximum_name_length is zero or less than the length 
+of the actual server_name then round the server_name up to a multiple of 32 octets 
+and randomly add another 32 octets 50% of the time and then include that amount
 of additional padding.
 
 Sum all the padding together, then in order to reduce entropy across different
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -257,7 +257,7 @@ HPKE algorithm identifiers.
 maximum_name_length  
 : The largest name the server expects to support, if known.
 If this value is not known (e.g. if wildcard names are in use, or names can be
-added or removed from tne anonymity set during the lifetime of a particular
+added or removed from the anonymity set during the lifetime of a particular
 resource record value), then this value can be set to zero, in which case
 clients SHOULD use the inner ClientHello padding scheme described below.
 
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -471,9 +471,10 @@ of the actual server_name then round the server_name up to a multiple of 32 octe
 and randomly add another 32 octets 50% of the time and then include that amount
 of additional padding.
 
-Sum all the padding together, then in order to reduce entropy across different
-client implementationss, round up so that the inner ClientHello encoding length 
-is a multiple of 32.
+The amount of padding applied to the ClientHello is then computed as the sum of 
+all per-extension padding values, rounded up to the nearest multiple of 32. 
+(This additional rounding step aims to hide variance across different client
+implementations.)
 
 In addition to padding the ClientHelloInner, clients and servers will also need
 to pad all other handshake messages that have sensitive-length fields. For
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -258,7 +258,7 @@ minimum_inner_length
 :  The minimum (encoded) length ClientHelloInner recommended by the server.
 Sending a ClientHelloInner that is shorter than this will likely stand out.
 Clients can use a padding extension within the ClientHelloInner to ensure
-the plaintext is at leat this long.
+the plaintext is at least this long.
 
 extensions
 : A list of extensions that the client can take into consideration when
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -444,14 +444,36 @@ The client then generates a ClientHelloInner value. In addition to the normal
 values, ClientHelloInner MUST also contain:
 
  - an "echo_nonce" extension
- - TLS padding {{!RFC7685}}
+ - TLS padding {{!RFC7685}} (see section {{!padding}})
 
-Variations in the length of the ciphertext version of the ClientHelloInner
-could defeat the purpose of ECHO, if those expose the chosen server_name field
-in the ClientHelloInner. The padding ClientHello extension, if well used, can
-ensure that length information does not expose what is contained in the
-ClientHelloInner. Clients SHOULD add padding to the ClientHelloInner to meet this
-requirement.
+When offering an encrypted ClientHello, the client MUST NOT offer to resume any
+non-ECHO PSKs. It additionally MUST NOT offer to resume any sessions for TLS 1.2
+or below.
+
+The encrypted ClientHello value is then computed as:
+
+~~~~
+    encrypted_ch = context.Seal("", ClientHelloInner)
+~~~~
+
+Finally, the client MUST generate a ClientHelloOuter message
+containing the "encrypted_client_hello" extension with the values as
+indicated above. In particular,
+
+- suite contains the client's chosen ciphersuite;
+- record_digest contains the digest of the corresponding ECHOConfig structure;
+- enc contains the encapsulated key as output by SetupBaseS; and
+- encrypted_ch contains the HPKE encapsulated key (enc) and the ClientHelloInner ciphertext (encrypted_ch_inner).
+
+The client MUST place the value of ECHOConfig.public_name in the
+ClientHelloOuter "server_name" extension. The remaining
+contents of the ClientHelloOuter MAY be identical to those in
+ClientHelloInner but MAY also differ.  The ClientHelloOuter MUST NOT
+contain a "cached_info" extension {{!RFC7924}} with a CachedObject
+entry whose CachedInformationType is "cert", since this indication
+would divulge the true server name.
+
+## Recommended Padding Scheme {#padding}
 
 Future extensions could reveal sensitive information through their length. Consequently, 
 padding should be flexible and support arbitrary extension changes. 
@@ -481,33 +503,6 @@ to pad all other handshake messages that have sensitive-length fields. For
 example, if a client proposes ALPN values in ClientHelloInner, the
 server-selected value will be returned in an EncryptedExtension, so that
 handshake message also needs to be padded using TLS record layer padding.
-
-When offering an encrypted ClientHello, the client MUST NOT offer to resume any
-non-ECHO PSKs. It additionally MUST NOT offer to resume any sessions for TLS 1.2
-or below.
-
-The encrypted ClientHello value is then computed as:
-
-~~~~
-    encrypted_ch = context.Seal("", ClientHelloInner)
-~~~~
-
-Finally, the client MUST generate a ClientHelloOuter message
-containing the "encrypted_client_hello" extension with the values as
-indicated above. In particular,
-
-- suite contains the client's chosen ciphersuite;
-- record_digest contains the digest of the corresponding ECHOConfig structure;
-- enc contains the encapsulated key as output by SetupBaseS; and
-- encrypted_ch contains the HPKE encapsulated key (enc) and the ClientHelloInner ciphertext (encrypted_ch_inner).
-
-The client MUST place the value of ECHOConfig.public_name in the
-ClientHelloOuter "server_name" extension. The remaining
-contents of the ClientHelloOuter MAY be identical to those in
-ClientHelloInner but MAY also differ.  The ClientHelloOuter MUST NOT
-contain a "cached_info" extension {{!RFC7924}} with a CachedObject
-entry whose CachedInformationType is "cert", since this indication
-would divulge the true server name.
 
 ## Handling the server response {#handle-server-response}
 
@@ -971,6 +966,17 @@ verifying the server's identity in its certificate.
 [[TODO: Some more analysis needed in this case, as it is a little
 odd, and probably some precise rules about handling ECHO and no
 SNI uniformly?]]
+
+## Padding
+
+Variations in the length of the ciphertext version of the ClientHelloInner
+could defeat the purpose of ECHO, if those expose the chosen server_name field
+in the ClientHelloInner. The padding ClientHello extension, if well used, can
+ensure that length information does not expose what is contained in the
+ClientHelloInner. Clients SHOULD add padding to the ClientHelloInner to meet this
+requirement.
+
+[[TODO: once the padding scheme is final, revisit the above.]]
 
 # IANA Considerations
 
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -453,11 +453,8 @@ ensure that length information does not expose what is contained in the
 ClientHelloInner. Clients SHOULD add padding to the ClientHelloInner to meet this
 requirement.
 
-Given that extensions could be defined in the future that reveal sensitive information
-through their length, we cannot describe all the ways in which length information
-could expose sensitive content. Implementers ought therefore be aware that they
-might have to change their padding scheme as the set of supported extensions
-changes. 
+Future extensions could reveal sensitive information through their length. Consequently, 
+padding should be flexible and support arbitrary extension changes. 
 
 For each field in the inner ClientHello, clients need to determine how much to
 pad given the semantics of that field.  For example, if a client can propose
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -453,8 +453,8 @@ that length information does not expose what is contained in the
 ClientHelloInner. Clients MUST add padding to the ClientHelloInner to try meet
 this requirement.
 
-Given that new extensions may be defined in future that contain sensitive
-length fields, we cannot describe all the ways in which length information
+Given that extensions could be defined in the future that reveal sensitive information
+through their length, we cannot describe all the ways in which length information
 could expose sensitive content. Implementers ought therefore be aware that they
 might have to change their padding scheme as the set of supported extensions
 changes. In addition to padding the ClientHelloInner, clients and servers MUST
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -204,7 +204,7 @@ following ECHOConfigs structure.
         HkpeKemId kem_id;
         CipherSuite cipher_suites<2..2^16-2>;
 
-        uint16 minimum_inner_length;
+        uint16 maximum_name_length;
         Extension extensions<0..2^16-1>;
     } ECHOConfigContents;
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -254,10 +254,11 @@ cipher_suites
...
 
-maximum_name_length
-:  the largest name the server expects to support. If the server supports arbitrary
-wildcard names, it SHOULD set this value to 256. Clients SHOULD reject ESNIConfig as
-invalid if maximum_name_length is greater than 256.
+minimum_inner_length
+:  The minimum (encoded) length ClientHelloInner recommended by the server.
+Sending a ClientHelloInner that is shorter than this will likely stand out.
+Clients can use a padding extension within the ClientHelloInner to ensure
+the plaintext is at leat this long.
```
```suggestion
the plaintext is at least this long.
```

draft-ietf-tls-esni.md
```diff
@@ -444,10 +445,33 @@ values, ClientHelloInner MUST also contain:
...
+An encoded ClientHelloInner can be around 300 octets long, but variations in
+the length of the ciphertext version of that could defeat the entire purpose of
+ECHO, if for example those expose the chosen server_name field in the
+ClientHelloInner.  The padding ClientHello extension, if well used, can ensure
+that length information does not expose what is contained in the
+ClientHelloInner. Clients MUST add padding to the ClientHelloInner to try meet
+this requirement.
+
+Given that new extensions may be defined in future that contain sensitive
+length fields, we cannot describe all the ways in which length information
```
```suggestion
Given that extensions could be defined in the future that reveal sensitive information
through their length, we cannot describe all the ways in which length information
```

draft-ietf-tls-esni.md
```diff
@@ -444,10 +445,33 @@ values, ClientHelloInner MUST also contain:
...
+changes. In addition to padding the ClientHelloInner, clients and servers MUST
+both pad all other handshake messages that have sensitive length fields. For
+example, if the client proposes ALPN values in the ClientHelloInner, the
+server-selected value will be returned in an EncryptedExtension, so that
+handshake message needs to be padded with TLS record layer padding. 
+
+One approach that may be relative safe and simple is once ECHO is in use:
+
+- When constructing the ClientHelloInner, use padding bytes so that the length
+  of the plaintext erncoding of ClienHelloInner is greater than
```
```suggestion
  of the plaintext encoding of ClienHelloInner is greater than
```

draft-ietf-tls-esni.md
```diff
@@ -444,10 +445,33 @@ values, ClientHelloInner MUST also contain:
...
+ECHOConfig.minimum_inner_length and is a multiple of 16
+- Both clients and servers pad all handshake messages so that plaintexts are a
+  multiple of 16 octets long.
+- Clients or servers MAY send additional padding bytes. This MUST NOT be
+  treated as an error.
```
martinthomson: What about the converse?  Is it an error to send something that is not padded to the minimum length?

sftcd: I think I'd say no, that's maybe better not treated as an error. But I could be convinced otherwise.

sftcd:
Thanks for the fixes. 
On the general point, I think that *if* we keep the flexibility for the client to vary the inner/outer CH however it chooses, then we will also have to depend on clients to figure out how to pad well for whatever they do. If we constrain the inner/outer variance a lot, then we could do better.

davidben:
I don't think this PR works. The server does not have enough information to report a useful value, and the value the server reports is not useful to maintain good anonymity sets.

The server does not know that the client supports some large extension or key share type (or lots of cipher suites, ALPN protocols, etc). It also doesn't know which extensions the client will compress from the outer ClientHello and which it will reencode.

Conversely, the value the server sets isn't useful. If the client implementation has a smaller baseline compression and extensions, we waste bytes without much benefit. Worse, if the client implementation has a larger baseline compression and extensions, this PR regresses anonymity. The ClientHello will then exceed `minimum_inner_length` and we don't hide the lengths of the sensitive bits.

Stepping back, we want to hide the distribution of ClientHelloInner lengths within some anonymity set. Most of the contributors to that length come from local client configuration: how many ciphers you support, what ALPN protocols, etc. The client is the one that knows, e.g., it sometimes asks servers for protocols A and B and sometimes for A and C. It largely has enough information to pad itself.

The main exception is the server name. Realistically we need to trim anonymity sets down to colocated services, and the client doesn't know that distribution of names. Thus, we want servers to report the value in ECHOConfig. Maybe we'll need more such statistics later, which is what extensions are for. It's annoying this is a bit complicated, but we've signing up to encrypt far more things now.

I should note this presumes an anonymity set of similarly-configured clients. I think, realistically, that's all we can say rigorous things about. It's nice to reduce visible artifacts of configuration, which is why we *additionally* round up to multiples of 16, but that's something we can burn into the protocol without ECHOConfig support.

sftcd:
Hiya,

On 16/03/2020 16:02, David Benjamin wrote:
> I don't think this PR works. The server does not have enough
> information to report a useful value, and the value the server
> reports is not useful to maintain good anonymity sets.
> 
> The server does not know that the client supports some large
> extension or key share type (or lots of cipher suites, ALPN
> protocols, etc). It also doesn't know which extensions the client
> will compress from the outer ClientHello and which it will reencode.
> 
> Conversely, the value the server sets isn't useful. If the client
> implementation has a smaller baseline compression and extensions, we
> waste bytes without much benefit. Worse, if the client implementation
> has a larger baseline compression and extensions, this PR regresses
> anonymity. The ClientHello will then exceed `minimum_inner_length`
> and we don't hide the lengths of the sensitive bits.

Sorry, not seeing that last. The PR just says (or was
meant to say) to pad to some multiple of 16 in that
case which I agree isn't much guidance, but does allow
doing something that works. (As well as allowing things
that don't work well;-)

> 
> Stepping back, we want to hide the distribution of ClientHelloInner
> lengths within some anonymity set. Most of the contributors to that
> length come from local client configuration: how many ciphers you
> support, what ALPN protocols, etc. The client is the one that knows,
> e.g., it sometimes asks servers for protocols A and B and sometimes
> for A and C. It largely has enough information to pad itself.

I agree that all this really needs to be driven by the
client.

> The main exception is the server name. Realistically we need to trim
> anonymity sets down to colocated services, and the client doesn't
> know that distribution of names. 

Well, yes and no. I don't believe a server config is
useful for that, (as I've argued before and still do;-)
ISTM the max server name value will almost always be
set to 254. May as well tell the client to pad as do
that IMO.

> Thus, we want servers to report the
> value in ECHOConfig. Maybe we'll need more such statistics later,
> which is what extensions are for. It's annoying this is a bit
> complicated, but we've signing up to encrypt far more things now.
> 
> I should note this presumes an anonymity set of similarly-configured
> clients. I think, realistically, that's all we can say rigorous
> things about. It's nice to reduce visible artifacts of configuration,
> which is why we *additionally* round up to multiples of 16, but
> that's something we can burn into the protocol without ECHOConfig
> support.

Not quite sure if I agree with that last para or not TBH!

S.


>

davidben:
> > Conversely, the value the server sets isn't useful. If the client
> > implementation has a smaller baseline compression and extensions, we
> >  waste bytes without much benefit. Worse, if the client implementation
> > has a larger baseline compression and extensions, this PR regresses
> > anonymity. The ClientHello will then exceed `minimum_inner_length`
> > and we don't hide the lengths of the sensitive bits.
>
> Sorry, not seeing that last. The PR just says (or was
> meant to say) to pad to some multiple of 16 in that
> case which I agree isn't much guidance, but does allow
> doing something that works. (As well as allowing things
> that don't work well;-)

Well, if we were satisfied just padding names to a multiple of 16, we wouldn't need any of this. We'd just say to pad to a multiple 16 and move on. :-) In the current name-focused spelling, the client is always able to apply a smarter padding decision based on the server's knowledge of the name distribution. With this PR, clients whose local configuration exceeds or is near `minimum_inner_length` lose this and hit the baseline multiple of 16 behavior. Clients whose local configuration is decently below `minimum_inner_length` do hide the name distribution, but using many more bytes than they would with the current text.

> > The main exception is the server name. Realistically we need to trim
> > anonymity sets down to colocated services, and the client doesn't
> > know that distribution of names.
>
> Well, yes and no. I don't believe a server config is
> useful for that, (as I've argued before and still do;-)
> ISTM the max server name value will almost always be
> set to 254. May as well tell the client to pad as do
> that IMO.

We have some metrics in Chrome on the distribution of DNS names. It's nowhere near 254. But, regardless, if you don't believe a server config is even useful for the name and agree with me that other extensions ought to be client-driven, I'm confused, why this PR? This PR seems to go in the opposite direction.

> > I should note this presumes an anonymity set of similarly-configured
> > clients. I think, realistically, that's all we can say rigorous
> > things about. It's nice to reduce visible artifacts of configuration,
> > which is why we *additionally* round up to multiples of 16, but
> > that's something we can burn into the protocol without ECHOConfig
> > support.
>
> Not quite sure if I agree with that last para or not TBH!

Hehe. Which part?

sftcd:
HIya,

On 16/03/2020 22:13, David Benjamin wrote:
>>> Conversely, the value the server sets isn't useful. If the
>>> client implementation has a smaller baseline compression and
>>> extensions, we waste bytes without much benefit. Worse, if the
>>> client implementation has a larger baseline compression and
>>> extensions, this PR regresses anonymity. The ClientHello will
>>> then exceed `minimum_inner_length` and we don't hide the lengths
>>> of the sensitive bits.
>> 
>> Sorry, not seeing that last. The PR just says (or was meant to say)
>> to pad to some multiple of 16 in that case which I agree isn't much
>> guidance, but does allow doing something that works. (As well as
>> allowing things that don't work well;-)
> 
> Well, if we were satisfied just padding names to a multiple of 16, we
> wouldn't need any of this. We'd just say to pad to a multiple 16 and
> move on. :-) In the current name-focused spelling, the client is
> always able to apply a smarter padding decision based on the server's
> knowledge of the name distribution. With this PR, clients whose local
> configuration exceeds or is near `minimum_inner_length` lose this and
> hit the baseline multiple of 16 behavior. Clients whose local
> configuration is decently below `minimum_inner_length` do hide the
> name distribution, but using many more bytes than they would with the
> current text.

Yes. That's IMO the downside of having any server config.
(To answer your question below, I prefer not having any
server config at all but proposed this as I think it's
better than draft-06.)

> 
>>> The main exception is the server name. Realistically we need to
>>> trim anonymity sets down to colocated services, and the client
>>> doesn't know that distribution of names.
>> 
>> Well, yes and no. I don't believe a server config is useful for
>> that, (as I've argued before and still do;-) ISTM the max server
>> name value will almost always be set to 254. May as well tell the
>> client to pad as do that IMO.
> 
> We have some metrics in Chrome on the distribution of DNS names. It's
> nowhere near 254. 

Having also looked at some passive DNS stuff, I fully
agree. But ISTM for many scenarios the entity that
generates the ECHOConfig won't know the distribution
of names in use (e.g. whoever configures OpenSSL won't,
same with Apache etc.) so it's likely they'll have to
go for whatever is the max (i.e. 254). Same is true
for anyone with a wild-card server cert in play
somewhere.

> But, regardless, if you don't believe a server
> config is even useful for the name and agree with me that other
> extensions ought to be client-driven, I'm confused, why this PR? This
> PR seems to go in the opposite direction.

See above. I'm trying for either a modest improvement
over draft-06 or for us all to arrive at a real
improvement by ditching the server config entirely:-)

> 
>>> I should note this presumes an anonymity set of
>>> similarly-configured clients. I think, realistically, that's all
>>> we can say rigorous things about. It's nice to reduce visible
>>> artifacts of configuration, which is why we *additionally* round
>>> up to multiples of 16, but that's something we can burn into the
>>> protocol without ECHOConfig support.
>> 
>> Not quite sure if I agree with that last para or not TBH!
> 
> Hehe. Which part?

I like the last sentence. I think I dislike the 1st:-)

Cheers,
S.

>

davidben:
> > [...] With this PR, clients whose local configuration exceeds or is near `minimum_inner_length` lose this and hit the baseline multiple of 16 behavior.  [...trimmed for brevity, but please let me know if I've misinterpreted what 'That' referred to!]
>
> Yes. That's IMO the downside of having any server config.
> (To answer your question below, I prefer not having any
> server config at all but proposed this as I think it's
> better than draft-06.)

It's not a downside of the draft-06 formulation, is it? The client learns something targeted to the name and can pad accordingly.

> >  We have some metrics in Chrome on the distribution of DNS names. It's nowhere near 254.
>
> Having also looked at some passive DNS stuff, I fully
> agree. But ISTM for many scenarios the entity that
> generates the ECHOConfig won't know the distribution
> of names in use (e.g. whoever configures OpenSSL won't,
> same with Apache etc.) so it's likely they'll have to
> go for whatever is the max (i.e. 254). Same is true
> for anyone with a wild-card server cert in play
> somewhere.

Even with a wildcard server cert, the server should still know the distribution. For instance, `*.github.io` isn't going to see anything terribly long most of the time. (Looks like the registration limit is 39 characters.) For reference, this is how long a 254-character hostname is:

aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa

In order to enjoy wide adoption, ECHO can't be too much of a waste of resources.  I think it thus makes sense to target a different maximum here.

> See above. I'm trying for either a modest improvement
> over draft-06 or for us all to arrive at a real
> improvement by ditching the server config entirely:-)

I guess my view is that, because of this issue above, this is a regression over the draft-06 formulation, even though it looks more general.

sftcd:
Hiya,

On 17/03/2020 19:32, David Benjamin wrote:
>>> [...] With this PR, clients whose local configuration exceeds or
>>> is near `minimum_inner_length` lose this and hit the baseline
>>> multiple of 16 behavior.  [...trimmed for brevity, but please let
>>> me know if I've misinterpreted what 'That' referred to!]
>> 
>> Yes. That's IMO the downside of having any server config. (To
>> answer your question below, I prefer not having any server config
>> at all but proposed this as I think it's better than draft-06.)
> 
> It's not a downside of the draft-06 formulation, is it? The client
> learns something targeted to the name and can pad accordingly.

Draft-06 assumes that no other inner CH content has the
property that server guidance will help. My claim is that
that guidance is useless in practice, and therefore
unnecessary. I think the server config'd length in the PR
is a little less useless:-)

> 
>>> We have some metrics in Chrome on the distribution of DNS names.
>>> It's nowhere near 254.
>> 
>> Having also looked at some passive DNS stuff, I fully agree. But
>> ISTM for many scenarios the entity that generates the ECHOConfig
>> won't know the distribution of names in use (e.g. whoever
>> configures OpenSSL won't, same with Apache etc.) so it's likely
>> they'll have to go for whatever is the max (i.e. 254). Same is
>> true for anyone with a wild-card server cert in play somewhere.
> 
> Even with a wildcard server cert, the server should still know the
> distribution. 

Some servers will. Some will not. Many (most?) server
ECHOConfig instances will make use of default settings.
Those'll pick 254 because adding a new VirtualHost to
an apache instance (or similar) won't be co-ordinated
with making a new ECHOConfig in general. And even it
it were, (but it won't!) we get a TTL's worth of mismatch.
If we don't provide a way for clients to handle the case
where the server config is too short, again servers will
publish 254.

> For instance, `*.github.io` isn't going to see anything
> terribly long most of the time. 

I am more interested in random smaller web sites/hosters
myself. I think ECHO should work for those as well as we
can make it, (albeit with smaller anonymity sets), and
this tie between web server config and ECHOConfig is just
a bad plan for such.

> (Looks like the registration limit is
> 39 characters.) For reference, this is how long a 254-character
> hostname is:
> 
> aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
>
>  In order to enjoy wide adoption, ECHO can't be too much of a waste
> of resources. 

AFAIK, all current deployments do just that. To be fair,
that's largely driven by CF's draft-02 deployment, but
I think that backs up my argument that if there's a max,
that'll almost always be chosen, which implies it's a
useless field that leads to the inefficiency you note.

> I think it thus makes sense to target a different
> maximum here.

Then it can't be the max:-) The 254 value matches the
max for server_name IIRC.

> 
>> See above. I'm trying for either a modest improvement over draft-06
>> or for us all to arrive at a real improvement by ditching the
>> server config entirely:-)
> 
> I guess my view is that, because of this issue above, this is a
> regression over the draft-06 formulation, even though it looks more
> general.

Ok. We disagree. But what do you think the client should
do? Draft-06 is IMO plainly incorrect in that it says to
pad in a way that'd expose other lengths should those be
sensitive. If your answer were pad the SNI to at least
ECHOConfig.maximum_name_length and then make the overall
a multiple of 16, with as many added 16 octets blocks as
the client chooses, then a) that's != draft-06 and b) is
quite close to this PR and c) just failing if the actual
SNI in the inner CH is longer than the config setting
seems quite wrong to me and d) hardcodes the assumption
that SNI is the only thing where the server may know
better.

Cheers,
S.


>

davidben:
> Draft-06 assumes that no other inner CH content has the
> property that server guidance will help. My claim is that
> that guidance is useless in practice, and therefore
> unnecessary. I think the server config'd length in the PR
> is a little less useless:-)

If we come up with other places where server guidance is useful, that can be addressed with extensions. But, gotcha, I suppose that is where we disagree. I think server guidance on the overall ClientHello size is entirely useless, while server guidance on the name is may be useful.

> I am more interested in random smaller web sites/hosters
> myself. I think ECHO should work for those as well as we
> can make it, (albeit with smaller anonymity sets), and
> this tie between web server config and ECHOConfig is just
> a bad plan for such.

Agreed that it should also work for smaller things. I don't follow how this PR helps in that regard. It seems they still need to come up with a random value or a default, but now it's even less clear how to set it well.

>> I think it thus makes sense to target a different
>> maximum here.
>
> Then it can't be the max:-) The 254 value matches the
> max for server_name IIRC.

Er, yes, sorry I meant to say target a different public length. That is, 254 is so far from the maximum domain name in practice (As it should be! Hard length limits ought to be comfortably away from the actual requirement.), so padding up to it is wasteful. I think it would be better to pad up to a tighter value and, if we exceed it, accept that we only have the fallback multiple of 16 padding. (Or maybe a different strategy.) It does result in smaller anonymity sets in edge cases, but I think that a reasonable default tradeoff given other desires like broader deployment and not wasting too much of users' data.

> Ok. We disagree. But what do you think the client should
> do? Draft-06 is IMO plainly incorrect in that it says to
> pad in a way that'd expose other lengths should those be
> sensitive. If your answer were pad the SNI to at least
> ECHOConfig.maximum_name_length and then make the overall
> a multiple of 16, with as many added 16 octets blocks as
> the client chooses, then a) that's != draft-06 and b) is
> quite close to this PR and c) just failing if the actual
> SNI in the inner CH is longer than the config setting
> seems quite wrong to me and d) hardcodes the assumption
> that SNI is the only thing where the server may know
> better.

I'm thinking something along the lines of:

1. For each field, determine how much it makes sense to pad given what kinds of things it sends. If it varies ALPN, maybe round up to the largest of those. Most of this can be determined without server help.
2. For fields where server help is useful, like the name, apply that help. Right now it's simply a `max` function. I'm also open to other strategies.
3. Sum all that padding together.
4. In order to generally reduce entropy across different kinds of clients, maybe apply some additional overall padding, like rounding up to a multiple of 16, which cuts down the state space by 16x and doesn't cost much. (Or maybe we should round up to the nearest `int(64 * 1.1**i)`? That avoids higher resolution on longer values while bounding the fraction of bytes wasted by this step.) That said, it's worth keeping in mind that clients already vary quite a bit in the cleartext portions of the ClientHello. There's a bit of tension between wanting to look like the other clients and wanting to ship new security improvements.

This is different from this PR because this PR asks the server to report a value including things it doesn't know anything about (how large the client-controlled parameters are). I don't quite follow (c). It seems this PR and the above proposal has roughly the same failure modes when something is too long. The difference is that `maximum_name_length`'s failure mode is only tripped on particular long names (something the server knows the distribution of), while this PR's equivalent failure mode is based on factors outside the server control, yet the server is responsible for the value.

davidben:
> For each field

Clarification: this is for each field where the client doesn't reference the outer ClientHello. If the client wishes to do that, it means the client doesn't consider that field secret so there isn't anything to pad.

chris-wood:
I like @davidben's suggested algorithm, as an ideal padding function requires input from both the client and server. (Ideally, servers would send padding recommendations for each extension that might vary specifically for them, such as the server name. They don't have insight into anything else, so clients need to make padding decisions based on local configuration.)

@sftcd, what do you think?

sftcd:
On 18/03/2020 15:26, Christopher Wood wrote:
> I like @davidben's suggested algorithm, as an ideal padding function
> requires input from both the client and server. (Ideally, servers
> would send padding recommendations for each extension that might vary
> specifically for them, such as the server name. They don't have
> insight into anything else, so clients need to make padding decisions
> based on local configuration.)
> 
> @sftcd, what do you think?

Could live with it, but still prefer no server config.
I'm looking a bit at some numbers and will try make a
concrete suggestion later today. If I don't get that
done, it seems fair to accept @davidben's algo and move
on.

Cheers,
S.

>

davidben:
If we can get away with no server config, I'm also entirely happy with that. Fewer moving parts is always great. We certainly could do that by saying names are padded up to 254, but I think that's too wasteful. But if we're happy with some global set of buckets that makes everyone happy (be it multiples of 16, some funny exponential thing), great!

sftcd:
(trimming violently and responding slowly:-)

On 17/03/2020 22:02, David Benjamin wrote:
> I'm thinking something along the lines of:

So your steps 1-6 is a better scheme than draft-06.

I still think step 2 would be better as something like:

2. ECHOConfig contains max_name_len as before but we
RECOMMEND it be zero and a) if the name to be padded is
longer than max_name_len, then names will be padded as
follows: P=32-(len(name)%32)+heads_or_tails*32, or b)
if the name to be padded is shorter than max_name_len
then we just do as the server asked and pad to that
length.

That (hopefully:-) means a default of "pad out to the
next multiple of 32, then toss a coin and add another
32 if the coin landed as heads."

If we landed somewhere there, I'd be happy to make this
PR say that or to make a new one or whatever.

According to some data I have the CDFs for names and
for these padded lengths would be as follows:

Len, CDF, Padded-CDF
32, 81.28, 40.64
64, 97.87, 89.57
96, 99.31, 98.59
128, 99.87, 99.59

That means that 81.28 of names are <32 octets long
and 40.64% of name+name_padding will be 32 octets
long. 97.87% of names are <64 octets and 89.57% of
name+name_padding are 64 octets long.

I think we should RECOMMEND setting max_name_len to
zero unless an ESNI key generator has a specific
reason to set it to a non-zero value. ISTM the use-case
for non-zero values is where the anonymity set has only
a few very long names that with length differences of
>32 octets for which it is considered important to use
ECHO. I'm basically fine that we default to not
serving that use-case wellmby default, as I don't
believe it's real. (But I could be wrong;-)


Cheers,
S.

sayrer:
Could the sender always pad their packets to at least 1200 bytes or something, using an encrypted extension in the ClientHello?

chris-wood:
@sftcd what's the value in adding zero or one 32B blocks after rounding? Why not just stick with what was done, say, in RFC8467?

chris-wood:
> I don't understand the focus on server input to a padding function that would ideally fill out one packet.

Per [David's comment](https://github.com/tlswg/draft-ietf-tls-esni/pull/209#issuecomment-600322366) above, clients *could* use help from servers in choosing appropriate padding lengths. (Only servers know their name anonymity set.)

sftcd:
Hiya,

On 22/03/2020 16:32, Christopher Wood wrote:
> @sftcd what's the value in adding zero or one 32B blocks after
> rounding? Why not just stick with what was done, say, in RFC8467?
>

Say if the anonymity set has lots of names <32 length
(~81% of names in the data I have) but a small number
between 32 and 64 (~16% in my data). Then this should
disguise the use of the longer names at the expense
of sending more octets. Would be v. happy to know if
that made sense to others.

Cheers,
S.

sftcd:
A friendly amendment below (but one I think is
important)...

On 22/03/2020 16:34, Christopher Wood wrote:
>> I don't understand the focus on server input to a padding function
>> that would ideally fill out one packet.
> 
> Per [David's
> comment](https://github.com/tlswg/draft-ietf-tls-esni/pull/209#issuecomment-600322366)
> above, clients *could* use help from servers in choosing appropriate
> padding lengths. (Only servers know their name anonymity set.)
> 

I think that last really needs to be "Only servers can know
their name anonymity set. But many servers will not know
that, either due to wildcard certs or because names/vhosts
are added and removed all the time independent of whatever
is in the DNS."

Cheers,
S.

chris-wood:
> Would be v. happy to know if that made sense to others.

It makes sense, I'm not not sure it's worth the cost. That said, this will always be an imperfect solution, so I'm happy either way.

sftcd:
On 22/03/2020 16:44, Christopher Wood wrote:
>> Would be v. happy to know if that made sense to others.
> 
> It makes sense, I'm just not not sure it's worth the cost. 

Yep. Me neither. It is cheaper than padding to 254 octets
for everyone though;-)

Partly, I'm not that sympathetic towards anyone who
wants to hide a crazily-long server_name. OTOH, I guess
there may be a few who can't easily change names so we
probably ought try do something for 'em.

Cheers,
S.

> That said,
> this will always be an imperfect solution, so I'm happy either way.
> 
>

sayrer:
> > I don't understand the focus on server input to a padding function that would ideally fill out one packet.
> 
> Per [David's comment](https://github.com/tlswg/draft-ietf-tls-esni/pull/209#issuecomment-600322366) above, clients _could_ use help from servers in choosing appropriate padding lengths. (Only servers know their name anonymity set.)

Agree that only servers could know their name anonymity set. But what is the cost of always padding out the ClientHello so it approaches the MTU? Just trying to understand why the spec attempts to economize on the number of bytes in the ClientHello packet.

chris-wood:
> But what is the cost of always padding out the ClientHello so it approaches the MTU?

As far as I know, there's no reason other than folks might not want to always "waste" these bytes.

davidben:
Padding the ClientHello up to an MTU isn't free. See the numbers here on what a 400-byte increase costs.
https://www.imperialviolet.org/2018/04/11/pqconftls.html

Likewise, also from that post, post-quantum key shares will likely exceed a packet anyway. I think our padding scheme should be designed with that expectation in mind (e.g., the `minimum_inner_length` mechanism fails this because a PQ key share in the inner ClientHello will blow past the built-in server assumptions on client behavior). Canary builds of Chrome already have code for such a thing behind a flag.

sayrer:
> Padding the ClientHello up to an MTU isn't free. See the numbers here on what a 400-byte increase costs.
> https://www.imperialviolet.org/2018/04/11/pqconftls.html

It's not quite clear what's going on in those results. For example, the mobile results are close to zero. Without seeing confidence intervals, and the distributions of the full packet sizes, I don't think this experiment settles this issue (although it could, with more detail). Also, is there a theory on /why/ 400 bytes might add latency?

> Likewise, also from that post and a [later experiment](https://www.imperialviolet.org/2019/10/30/pqsivssl.html), post-quantum key shares will likely exceed a packet anyway. 

If we're going to design the padding mechanism with PQ key shares in mind, but not conventional ones, that should be in the document.

davidben:
> It's not quite clear what's going on in those results. For example, the mobile results are close to zero. Without seeing confidence intervals, and the distributions of the full packet sizes, I don't think this experiment settles this issue (although it could, with more detail). Also, is there a theory on /why/ 400 bytes might add latency?

@agl would know the details of that experiment, but I don't think the idea that sending more bytes takes more time would be terribly surprising! :-P

> If we're going to design the padding mechanism with PQ key shares in mind, but not conventional ones, that should be in the document.

I don't think anyone's suggested designing without conventional key shares in mind. They are what's deployed today, after all. However, we shouldn't design *only* for a small unpadded ClientHello. PQ key shares will need larger ones. Session tickets also go in the ClientHello, and it's common for TLS servers to stick client certificates inside tickets, so you can already get large ones today already.

chris-wood:
@sftcd will you be able to update this as above?

sayrer:
would the spec allow me to put an arbitrarily-sized extension in the encrypted part of the ClientHello? if so, I don't care about the padding schemes.

chris-wood:
> would the spec allow me to put an arbitrarily-sized extension in the encrypted part of the ClientHello? if so, I don't care about the padding schemes.

Yes, of course! The padding policies described here are guidance at best.

sftcd:
I've updated the PR to try reflect the discussion above. Please take a peek and see if it seems useful now.

draft-ietf-tls-esni.md
```diff
@@ -444,10 +446,40 @@ values, ClientHelloInner MUST also contain:
...
+pad given the semantics of that field.  For example, if a client can propose
+various ALPN values, it could add padding to round up to the longest of those.
+
+For most fields in a ClientHello this can be determined without server help.
+For the server_name, however, if ECHOConfig.maximum_name_length is longer than
+the actual server_name then clients SHOULD add padding to make up that
+difference. If the maximum_name_length is zero or less than the length of the
+actual server_name then round the server_name up to a multiple of 32 octets and
+randomly add another 32 octets 50% of the time and then include that amount
+of additional padding.
```
chris-wood, line 472:
```suggestion
However, the "server_name" extension could benefit from server input (ECHOConfig.maximum_name_length). 
Clients SHOULD compute the padding for this extension as follows:

1. If ECHOConfig.maximum_name_length is longer than the actual server_name then 
clients SHOULD add padding to make up that difference.
2. Otherwise, if ECHOConfig.maximum_name_length is zero or less than the length 
of the actual server_name then round the server_name up to a multiple of 32 octets 
and randomly add another 32 octets 50% of the time and then include that amount
of additional padding.
```

draft-ietf-tls-esni.md
```diff
@@ -444,10 +446,40 @@ values, ClientHelloInner MUST also contain:
...
+
+Sum all the padding together, then in order to reduce entropy across different
+client implementationss, round up so that the inner ClientHello encoding length 
+is a multiple of 32.
```
chris-wood, line 476:
```suggestion
The amount of padding applied to the ClientHello is then computed as the sum of 
all per-extension padding values, rounded up to the nearest multiple of 32. 
(This additional rounding step aims to hide variance across different client
implementations.)
```

draft-ietf-tls-esni.md
```diff
@@ -444,10 +446,40 @@ values, ClientHelloInner MUST also contain:
...
+
+In addition to padding the ClientHelloInner, clients and servers will also need
+to pad all other handshake messages that have sensitive-length fields. For
+example, if a client proposes ALPN values in the ClientHelloInner, the
+server-selected value will be returned in an EncryptedExtension, so that
+handshake message also needs to be padded using TLS record layer padding.
```
chris-wood, line 482:
```suggestion
In addition to padding ClientHelloInner, clients and servers will also need
to pad all other handshake messages that have sensitive-length fields. For
example, if a client proposes ALPN values in ClientHelloInner, the
server-selected value will be returned in an EncryptedExtension, so that
handshake message also needs to be padded using TLS record layer padding.
```

draft-ietf-tls-esni.md
```diff
@@ -254,10 +254,12 @@ cipher_suites
...
 HPKE algorithm identifiers.
 
-maximum_name_length
-:  the largest name the server expects to support. If the server supports arbitrary
-wildcard names, it SHOULD set this value to 256. Clients SHOULD reject ESNIConfig as
-invalid if maximum_name_length is greater than 256.
+maximum_name_length  
+: The largest name the server expects to support, if known.
+If this value is not known (e.g. if wildcard names are in use, or names can be
+added or removed from tne anonymity set during the lifetime of a particular
```
chris-wood, line 260:
```suggestion
added or removed from the anonymity set during the lifetime of a particular
```

draft-ietf-tls-esni.md
```diff
@@ -444,10 +446,40 @@ values, ClientHelloInner MUST also contain:
...
+through their length, we cannot describe all the ways in which length information
+could expose sensitive content. Implementers ought therefore be aware that they
+might have to change their padding scheme as the set of supported extensions
+changes. 
+
+For each field in the inner ClientHello, clients need to determine how much to
+pad given the semantics of that field.  For example, if a client can propose
+various ALPN values, it could add padding to round up to the longest of those.
+
+For most fields in a ClientHello this can be determined without server help.
```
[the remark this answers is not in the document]
chris-wood: Can we also merge this paragraph with the preceding one?

sayrer:
This looks like the right idea. I'll review this again as well, but it's probably better to wait for @chris-wood's suggestions to be addressed.

sftcd:
Thanks for all the changes Chris. I think I've actioned all those as requested. If not, I tried, but failed:-)

draft-ietf-tls-esni.md
```diff
@@ -476,6 +473,37 @@ contain a "cached_info" extension {{!RFC7924}} with a CachedObject
...
 would divulge the true server name.
 
+## Recommended Padding Scheme {#padding}
+
+Future extensions could reveal sensitive information through their length. Consequently, 
+padding should be flexible and support arbitrary extension changes. 
+
+For each field in the inner ClientHello, clients need to determine how much to
+pad given the semantics of that field.  For example, if a client can propose
+various ALPN values, it could add padding to round up to the longest of those.
```
sayrer, line 483: Is this part right? Do clients need to consider padding for each field, or just the entire encrypted ClientHelloInner?

draft-ietf-tls-esni.md
```diff
@@ -476,6 +473,37 @@ contain a "cached_info" extension {{!RFC7924}} with a CachedObject
...
+The target padding length of most ClientHello extensions can be determined without server help.
+However, the "server_name" extension could benefit from server input (ECHOConfig.maximum_name_length). 
+Clients SHOULD compute the padding for this extension as follows:
+
+1. If ECHOConfig.maximum_name_length is longer than the actual server_name then 
+clients SHOULD add padding to make up that difference.
+2. Otherwise, if ECHOConfig.maximum_name_length is zero or less than the length 
+of the actual server_name then round the server_name up to a multiple of 32 octets 
+and randomly add another 32 octets 50% of the time and then include that amount
+of additional padding.
```
sayrer, line 494: Suggest: "add at least another 32 octets".

draft-ietf-tls-esni.md
```diff
@@ -939,6 +967,17 @@ verifying the server's identity in its certificate.
 odd, and probably some precise rules about handling ECHO and no
 SNI uniformly?]]
 
+## Padding
+
+Variations in the length of the ciphertext version of the ClientHelloInner
```
sayrer, line 972: Suggest: "Variations in the length of the ClientHelloInner ciphertext..."

draft-ietf-tls-esni.md
```diff
@@ -476,6 +473,37 @@ contain a "cached_info" extension {{!RFC7924}} with a CachedObject
...
+The amount of padding applied to the ClientHello is then computed as the sum of 
+all per-extension padding values, rounded up to the nearest multiple of 32. 
+(This additional rounding step aims to hide variance across different client
+implementations.)
+
+In addition to padding ClientHelloInner, clients and servers will also need
+to pad all other handshake messages that have sensitive-length fields. For
+example, if a client proposes ALPN values in ClientHelloInner, the
+server-selected value will be returned in an EncryptedExtension, so that
+handshake message also needs to be padded using TLS record layer padding.
```
sayrer, line 505: Is there a good reference for "TLS record layer padding"? I think I know what this means, but it could be clearer.

chris-wood: It's defined in Section 5.4 of RFC8446: https://tools.ietf.org/html/rfc8446#section-5.4

draft-ietf-tls-esni.md
chris-wood: Yep -- this is what was discussed way up above in this PR.

sayrer: Ok, I will re-read this PR, since I don't get it.

But, even so, I think this section needs a rationale. It says "clients need to determine how much to pad given the semantics of that field", but there is no reason given.

draft-ietf-tls-esni.md
sayrer: Maybe that reference suggests a terminology issue. It says "All encrypted TLS records can be padded to inflate the size of the TLSCiphertext." Does ECHO define things in terms of TLSCiphertext?

I don't necessarily think there is a problem here, but in the two code bases I've read (rustls and NSS), the encrypted SNI or ClientHello haven't fit in all that cleanly.

draft-ietf-tls-esni.md
jhoyla: Or "add one or more blocks of 32 octets" just to exclude behaviours that undo the effect of rounding the padding to a multiple of 32 octets.

draft-ietf-tls-esni.md
```diff
@@ -963,6 +967,17 @@ verifying the server's identity in its certificate.
 odd, and probably some precise rules about handling ECHO and no
 SNI uniformly?]]
 
+## Padding
+
+Variations in the length of the ciphertext version of the ClientHelloInner
+could defeat the purpose of ECHO, if those expose the chosen server_name field
```
jhoyla, line 973: Do we need to update GREASE if we are rounding the packet sizes to multiples of 32 octets? If we want GREASE to look like ECHO then it should have the same external properties.

draft-ietf-tls-esni.md
chris-wood: 👍 to N >= 1 blocks. Clients can choose the distribution from which they want to sample N.

draft-ietf-tls-esni.md
chris-wood: Yep -- there's a TODO in the GREASE section.

chris-wood:
@sftcd do you need any help moving this forward?

sftcd:
On 07/04/2020 14:47, Christopher Wood wrote:
> @sftcd do you need any help moving this forward?
>
Doing other stuff today but feel free to take the text
and then do more edits. Or I can look at it tomorrow.
Either's fine by me.

S.

draft-ietf-tls-esni.md
jhoyla: On second thoughts, we should probably match what's done in encrypted DNS, because otherwise we could create a potential leak from correlating the DNS request length with the ECHO length.

sayrer:
> On 07/04/2020 14:47, Christopher Wood wrote: @sftcd do you need any help moving this forward?
> Doing other stuff today but feel free to take the text and then do more edits. 

Thanks for taking the time. None of my comments are strongly-held opinions (meaning they only need to be addressed at all if others agree that the issues are important).

chris-wood:
Strong +1 to @sayrer -- thank you for the work here, @sftcd!

draft-ietf-tls-esni.md
sftcd: I moved that rationale down to the security considerations section as suggested earlier. It could be moved back up, but basically anything where the length could be sensitive may need padding.

draft-ietf-tls-esni.md
sftcd: Don't think copying DNS is correct. RFC 8467 calls for queries to be padded to a multiple of 128 octets and could be that the server-name is not the only sensitive name. 

I'm not sure there's a point in recommending against zero additional blocks of 32 - so long as you don't do zero too often it'd be ok. If you look back to the name lengths seen in the data I have it may be that zero additional blocks 50% of the time is ok. 

But mainly, I'd say we may be better to wait 'till we have a few implementations and then see what makes best sense.

chris-wood: > Don't think copying DNS is correct. RFC 8467 calls for queries to be padded to a multiple of 128 octets and could be that the server-name is not the only sensitive name.

Can you show that there is no information leakage if the two aren't in sync?

> I'm not sure there's a point in recommending against zero additional blocks of 32 - so long as you don't do zero too often it'd be ok. If you look back to the name lengths seen in the data I have it may be that zero additional blocks 50% of the time is ok.

The proposal above suggests N >= 1, not zero. In any case, as you suggest, spinning our wheels on something for which there is no backing data or analysis (yet) seems premature.

chris-wood:
@sftcd can you please resolve conflicts? @ekr @kazuho can you please review?

draft-ietf-tls-esni.md
chris-wood: Hmm, is this resolved?

ekr: I actually don't think we should recommend the random padding. Let's do the minimum here.

sftcd:
On 24/04/2020 16:07, ekr wrote:
> @ekr commented on this pull request.
> 
> I am generally fine with this, with the exception of the random
> padding, which I think should be omitted. It's hard to analyze and
> the value is unclear. In particular, many clients will retry if they
> receive TCP errors, and so the attacker can learn information about
> the true minimum value by forging TCP RSTs and looking at the new
> CH.

Fair point. The argument I'd make for keeping this
is that if an anonymity set has ~80% of names <32
octets, then if we do nothing, passive observation
will probably reveal when the server name is in
the longest ~20%. And I'd guess that only a few
names in that set would be popular/common, so we
could be giving away the game whenever an ECHO
is 32 octets longer than the usual. The random
padding could mitigate that passive attack but
yes would be vulnerable to an active reset based
counter.

I'd also be fine with just living with the issue
in which case we might want to note that using
ECHO with uncommonly long names is a bad idea.

> 
>> +and randomly add another 32 octets 50% of the time and then
>> include that amount
> +of additional padding.
> 
> I actually don't think we should recommend the random padding. Let's
> do the minimum here.
>

chris-wood:
Let's just drop the random addition. Servers don't (can't) do a padding check anyway, so clients can always choose to make this longer if desired.

chris-wood:
@sftcd can you update this PR and resolve the conflicts?

ekr:
"Fair point. The argument I'd make for keeping this
is that if an anonymity set has ~80% of names <32
octets, then if we do nothing, passive observation
will probably reveal when the server name is in
the longest ~20%. "

Which is precisely why we want the server to provide the longest name, so the client can pad to it.

chris-wood:
As discussed on today's call, let's do the following:

1. Resolve merge conflicts.
2. Drop random padding for the SNI. And add text which says that if the client's name is larger than max_name_length, it rounds up to the nearest 32B or 64B boundary.

@sftcd do you need, or would you like, any help with this?

PR #212, opened 2020-03-23
Bind CHOuter to CHInner

chris-wood:
This makes it so that changes to the outer CH do not affect the inner. If anything does change that might muck with the binder or decryption, then ECHO is ignored. (This should help GREASE.)

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -718,9 +718,12 @@ Otherwise, if there is no "echo_nonce" extension, or if its value does not
 match the derived echo_nonce, the server MUST abort the connection with a
 "decrypt_error" alert. Next, the server MUST scan ClientHelloInner for any
 "outer_extension" extensions and substitute their values with the values in
-ClientHelloOuter. It MUST first verify that the hash found in the extension
-matches the hash of the extension to be interpolated in and if it does not,
-abort the connections with a "decrypt_error" alert.
+ClientHelloOuter.
+
+If ClientHelloInner contains a resumption PSK binder, the server SHOULD process it
+as per the rules in {{!RFC8446}}. Note that this requires servers process more than
+one binder when resuming an ECHO connection: one in ClientHelloOuter and another
+in ClientHelloInner.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -392,18 +391,14 @@ ClientHelloInner.
 When sending ClientHello, the client first computes ClientHelloInner,
 including any PSK binders, and then MAY substitute extensions which
 it knows will be duplicated in ClientHelloOuter. To do so, the client
-computes a hash H of the entire ClientHelloInner message with the same
-hash as for the KDF used to encrypt ClienHelloInner. Then, the client
 removes and and replaces extensions from ClientHelloInner with a single
```
dmcardle: Just noticed this "and and"

chris-wood: Thanks :)

draft-ietf-tls-esni.md
```diff
@@ -476,6 +473,10 @@ contain a "cached_info" extension {{!RFC7924}} with a CachedObject
 entry whose CachedInformationType is "cert", since this indication
 would divulge the true server name.
 
+Finally, client computes a PSK binder using outer_binder_key with the identity
+outer_binder_identity. This binds the encryption of ClientHelloInner to
+ClientHelloOuter.
```
davidben: This probably wants an extra sentence that you add the identity and binder to the PSK extension. At first I was confused what you did with the binder.

draft-ietf-tls-esni.md
```diff
@@ -431,6 +426,8 @@ pkR = HPKE.KEM.Unmarshal(ECHOConfig.public_key)
 enc, context = SetupBaseS(pkR, "tls13-echo")
 echo_nonce = context.Export("tls13-echo-nonce", 16)
 echo_hrr_key = context.Export("tls13-echo-hrr-key", 16)
+outer_binder_identity = context.Export("tls13-echo-binder-identity", TODO)
```
davidben: (Hah. Nice!)

Issue #218, opened 2020-04-12
Use HPKE codepoints for all HPKE-relevant operations


sftcd:
HPKE has it's own cipher suite code points, as well as the KEM id. I don't see why we're using TLS codepoints (ECHOConfigContents.suites and ClientEncryptedCH.suite) and yet also use one HPKE code point (ECHOConfigContents.kem_id). But maybe I'm missing something?

If not, I think one of two things would make sense:

1. Change HPKE to use TLS code points. (That could be done but'd involve a bit of nose-holding, however it might be better in the longer term.)
2. Only use HPKE code points in ECHO - I don't see that TLS ciphersuites are actually needed for ECHO, so this seems easiest. (That'd mean s/suite/aead_id/g I guess.)

ekr:
I tend to agree with Stephen here. ISTM that probably (1) is better, actually.

chris-wood:
I don't think we want HPKE to depend on TLS code points. Algorithms that come down the pipe may not be suitable for TLS. I'd prefer option (2).

bifurcation:
There's also a third option here:

3. Use TLS code points (or something else) and define a mapping to HPKE code points

Like @chris-wood, I am disinclined to tie HPKE to TLS, since it is a more general tool.

It seems like the answer here depends on what the negotiation in ECHO is being used for.  If it's just for the HPKE, use HPKE code points.  If it's used in TLS further down the line, use TLS code points and map to HPKE for the HPKE bits.

ekr:
How is (3) different than the current design?

chris-wood:
Per the call today, we'll use this issue to track moving everything to HPKE codepoints. (We can remove the mapping table at the same time).

PR #224, opened 2020-04-28
Replace ECHOConfig TLS ciphersuites with HPKE code points.

chris-wood:
Addresses #218.

A given ECHOConfig now lists a single HPKE public key, with the corresponding
KEM identifier, and a list of acceptable AEAD and KDF algorithm identifiers.
Clients may then choose their preferred AEAD and KDF algorithms for ClientHello
encryption.

We may need to provide some guidance, either here or in the HPKE document, about acceptable (KEM, KDF, AEAD) tuples. 

cc @bifurcation

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -195,14 +195,21 @@ following ECHOConfigs structure.
 
 ~~~~
     opaque HpkePublicKey<1..2^16-1>;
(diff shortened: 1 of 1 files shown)


ekr:
I hadn't realized these were separate IDs. Oh well.

ocheron:
HpkeCipherSuite should also be used in ClientEncryptedCH.

chris-wood:
Oops -- thanks, @ocheron!

PR #225, opened 2020-05-02
Deterministic padding

chris-wood:
This builds on #209, but removes the probabilistic pieces. I created a fresh PR to make review easier. 

cc @sftcd

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -487,9 +487,9 @@ length. Thus, each extension in the inner ClientHello may require different amou
 of padding. This padding may be fully determined by the client's configuration or
 may require server input.
 
-By way of example, a client's set of ALPN values is typically fixed. Sensible padding
-for this extension should therefore round up to the longest of those values. The
-target padding length of most ClientHello extensions can be computed in this way.
+By way of example, a client's set of ALPN values is typically fixed. Clients SHOULD
+pad this extension by rounding up to the longest of those values. The target padding
+length of most ClientHello extensions can be computed in this way.
 
 In contrast, clients do not know the longest SNI value in the client-facing server's
 anonymity set without server input. For the "server_name" extension with length D,
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -496,8 +496,10 @@ anonymity set without server input. For the "server_name" extension with length 
 clients SHOULD use the server's length hint L (ECHOCOnfig.maximum_name_length) when
 computing the padding as follows:
 
-1. If L > D, add L - D bytes of padding.
-2. Otherwise, add 32 - (D % 32) bytes of padding.
+1. If L > D, add L - D bytes of padding. This rounds to the server's advertised
+hint, i.e., ECHOCOnfig.maximum_name_length.
+2. Otherwise, add 32 - (D % 32) bytes of padding. This rounds D up to the nearest
+multiple of 32 bytes.
 
 The amount of padding applied to the ClientHello is computed as the sum of
 all per-extension padding values, rounded up to the nearest multiple of 32.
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -482,6 +469,39 @@ would divulge the true server name. The remaining
...
+
+This section describes a deterministic padding mechanism based on the following
+observation: individual extensions can reveal sensitive information through their
+length. Thus, each extension in the inner ClientHello may require different amounts
+of padding. This padding may be fully determined by the client's configuration or
+may require server input.
+
+By way of example, a client's set of ALPN values is typically fixed. Clients SHOULD
+pad this extension by rounding up to the longest of those values. The target padding
+length of most ClientHello extensions can be computed in this way.
```
ekr, line 482: This text seems a bit confusing because clients typically offer a fixed set of ALPNs for a given application.  Perhaps:

"Clients typically support a small number of application profiles. For instance, a browser might support HTTP with ALPN values ["http/1.1, "h2"] and WebRTC media with ALPNs ["webrtc", "c-webrtc"]. Clients SHOULD pad this extension by rounding up to the total size of the longest ALPN extension across all application profiles".

draft-ietf-tls-esni.md
```diff
@@ -482,6 +469,39 @@ would divulge the true server name. The remaining
...
+
+1. If L > D, add L - D bytes of padding. This rounds to the server's advertised
+hint, i.e., ECHOCOnfig.maximum_name_length.
+2. Otherwise, add 32 - (D % 32) bytes of padding. This rounds D up to the nearest
+multiple of 32 bytes.
+
+The amount of padding applied to the ClientHello is computed as the sum of
+all per-extension padding values, rounded up to the nearest multiple of 32.
+(This additional rounding step aims to hide variance across different client
+implementations.)
```
ekr, line 497: Say we just padded SNI and L = 0 and D = 10. We then compute the per-extension padding value as 22, but because of this paragraph we pad by 32? If so, I don't think this works, because if D was 11, we would also pad by 32, so now we are leaking the length of D.

chris-wood: Discussed offline -- will remove and consider later on, as this exists to mask fixed-length values across different stacks, and fingerprinting stacks is already pretty trivial.

draft-ietf-tls-esni.md
martinthomson: This might be a little narrow a view.  In a browser, we have a bunch of things we might do.  But that is because the browser is platform.  But the anonymity set on a different platform (say you used the system TLS stack) is potentially wider (all apps) and cannot make the same claim about application profiles.

draft-ietf-tls-esni.md
```diff
@@ -482,6 +469,37 @@ would divulge the true server name. The remaining
...
+profiles. The target padding length of most ClientHello extensions can be computed
+in this way.
+
+In contrast, clients do not know the longest SNI value in the client-facing server's
+anonymity set without server input. For the "server_name" extension with length D,
+clients SHOULD use the server's length hint L (ECHOCOnfig.maximum_name_length) when
+computing the padding as follows:
+
+1. If L > D, add L - D bytes of padding. This rounds to the server's advertised
+hint, i.e., ECHOCOnfig.maximum_name_length.
```
martinthomson, line 493:
```suggestion
hint, i.e., ECHOConfig.maximum_name_length.
```

draft-ietf-tls-esni.md
chris-wood: The point is that the "platform," be it the stack used in the browser or in the OS, typically defines the set of values here and should pad accordingly. I'm fine with using the browser as an example here, as it's just a specific instance of that platform.

Issue #216, opened 2020-04-12
Clarify ECHOConfig (HTTPSSVC) extension purpose


sftcd:
ECHOConfig.public_name provides information that goes in the outer CH. HTTPSSVC.alpn provides information to go in the inner CH (I assume). We don't specify a way for the server to  explicitly say what alpn to use in the outer CH.  I think it'd be worth re-considering where servers/admins might want to put that kind of information and how much of the inner/outer CH they may want to describe. There could be a range of potential outcomes if we re-visit this - at one extreme HTTPSSVC or ECHOConfig might provide a template for the entire CH (possibly via some template naming/numbering scheme), at the other extreme we might decide that ECHOConfig should only contain the keying information and just define alpn, public_alpn and public_name as optional fields for inclusion in HTTPSSVC. But it seems a bit broken for the only two parts of the CH the server specifies to be handled so differently,

davidben:
> HTTPSSVC.alpn provides information to go in the inner CH (I assume).

It doesn't. It's important for downgrade protection that the DNS (not secured) *not* impact the client's ALPN preferences. The interaction between HTTPSSVC and ALPN is... rather a mess and the text around this still needs work.

sftcd:
On 13/04/2020 15:54, David Benjamin wrote:
>> HTTPSSVC.alpn provides information to go in the inner CH (I
>> assume).
> 
> It doesn't. It's important for downgrade protection that the DNS (not
> secured) *not* impact the client's ALPN preferences. The interaction
> between HTTPSSVC and ALPN is... rather a mess and the text around
> this still needs work.

Ah. Fair enough. Though I guess ECHO will likely only be
used via DoH or something equivalent which could change
things maybe.

But regardless... doesn't the point still stand that we'd
be better off with some consistency in how we manage the
server-name in the outer CH, the ALPNs in both inner and outer, and any
other CH content where the server wants to
publish some hints?

I'm not particularly wedded to a specific approach (though
I'd likely be more biased to simplifying the ECHOConfig).

davidben:
Even DoH won't authenticate the origin, only the resolver. One hopes the resolver is reasonably trusted, but I'd still prefer not giving it new powers. (Today, only the origin server pick HTTP/1.1 vs HTTP/2. That's useful given HTTP/1.1's safety against truncation attacks is... complicated.)

[...]

PR #226, opened 2020-05-02
Provide guidance on ECHOConfig extension usage.

chris-wood:
This addresses #216.

cc @davidben, @bemasc, @mikebishop, @sftcd, @enygren

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -268,8 +268,8 @@ invalid if maximum_name_length is greater than 256.
 extensions
 : A list of extensions that the client can take into consideration when
 generating a ClientHello message. The purpose of the field is to provide room
-for additional features in the future. See {{config-extensions}} for guidance
-on what type of extensions are appropriate for this structure.
+for additional functionality in the future. See {{config-extensions}} for 
+guidance on what type of extensions are appropriate for this structure.
 
 The format is defined in {{RFC8446}}, Section 4.2. The same interpretation rules
 apply: extensions MAY appear in any order, but there MUST NOT be more than one
@@ -940,13 +940,11 @@ existing registry for Alerts (defined in {{!RFC8446}}), with the
 # ECHOConfig Extension Guidance {#config-extensions}
 
 Any future information or hints that influence the outer ClientHello SHOULD be
-specified as ECHOConfig extensions. This is primarily because the outer ClientHello
-exists only in support of ECHO. Namely, it is both an envelope for the encrypted
-inner ClientHello and enabler for authenticated key mismatch signals
-(see {{server-behavior}}). In contrast, the inner ClientHello is the true ClientHello
-used upon ECHO negotiation. Any information that influences the inner ClientHello,
-if needed, therefore belongs elsewhere, such as the corresponding HTTPSSVC
-record {{!HTTPSSVC}}.
+specified as ECHOConfig extensions, or in an entirely new version of ECHOConfig. 
+This is primarily because the outer ClientHello exists only in support of ECHO. 
+Namely, it is both an envelope for the encrypted inner ClientHello and enabler for 
+authenticated key mismatch signals (see {{server-behavior}}). In contrast, the inner 
+ClientHello is the true ClientHello used upon ECHO negotiation. 
 
 --- back
 
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -267,17 +267,20 @@ invalid if maximum_name_length is greater than 256.
 
 extensions
 : A list of extensions that the client can take into consideration when
-generating a Client Hello message. The purpose of the field is to provide room
-for additional features in the future. The format is defined in {{RFC8446}};
-Section 4.2. The same interpretation rules apply: extensions MAY appear in any
-order, but there MUST NOT be more than one extension of the same type in the
-extensions block. An extension may be tagged as mandatory by using an extension
-type codepoint with the high order bit set to 1. A client which receives a
-mandatory extension they do not understand must reject the ECHOConfig content.
+generating a ClientHello message. The purpose of the field is to provide room
+for additional features in the future. See {{config-extensions}} for guidance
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -267,17 +267,20 @@ invalid if maximum_name_length is greater than 256.
...
-order, but there MUST NOT be more than one extension of the same type in the
-extensions block. An extension may be tagged as mandatory by using an extension
-type codepoint with the high order bit set to 1. A client which receives a
-mandatory extension they do not understand must reject the ECHOConfig content.
-
-Clients MUST parse the extension list and check for unsupported
-mandatory extensions. If an unsupported mandatory extension is
-present, clients MUST reject the ECHOConfig value.
+generating a ClientHello message. The purpose of the field is to provide room
+for additional features in the future. See {{config-extensions}} for guidance
```
martinthomson, line 271: features or protections

draft-ietf-tls-esni.md
```diff
@@ -267,17 +267,20 @@ invalid if maximum_name_length is greater than 256.
...
+on what type of extensions are appropriate for this structure.
+
+The format is defined in {{RFC8446}}; Section 4.2. The same interpretation rules
```
martinthomson, line 274: Comma

draft-ietf-tls-esni.md
```diff
@@ -934,6 +937,17 @@ IANA is requested to create an entry, echo_required(121) in the
 existing registry for Alerts (defined in {{!RFC8446}}), with the
 "DTLS-OK" column being set to "Y".
 
+# ECHOConfig Extension Guidance {#config-extensions}
+
+Any future information or hints that influence the outer ClientHello should be
```
martinthomson, line 942: SHOULD?

draft-ietf-tls-esni.md
```diff
@@ -934,6 +937,17 @@ IANA is requested to create an entry, echo_required(121) in the
...
+specified as ECHOConfig extensions. This is primarily because the outer ClientHello
+exists only in support of ECHO. Namely, it is both an envelope for the encrypted
+inner ClientHello and enabler for authenticated key mismatch signals
+(see {{server-behavior}}). In contrast, the inner ClientHello is the true ClientHello
+used upon ECHO negotiation. Any information that influences the inner ClientHello,
```
martinthomson, line 947: Except, as we have discussed, the length, though I guess that is a part of the outer, too.

chris-wood: Yep, I'm considering that part of the outer.

draft-ietf-tls-esni.md
chris-wood: Isn't protection a feature?

draft-ietf-tls-esni.md
chris-wood: Will fix here and elsewhere!

draft-ietf-tls-esni.md
chris-wood: I read them both the same, so that works for me.

MikeBishop:
This looks fine, though I'll note that since an HTTPSSVC record is public and the outer ClientHello record is public, it's not unreasonable to pull values from there to put in the public fields since the observer already knows (or can get) them.  Restricting them to come only from the ECHOConfig seems unnecessary.

I think the key observation in this PR is that the outer ClientHello isn't _really_ offering to establish a TLS connection -- it's a not-crazy-looking wrapper for the actual ClientHello inside the ECHO extension, and the vehicle for the recovery flow.  The only things a server can legitimately do are respond to the inner ClientHello or do a recovery handshake.  Thus, most of the things you might set on the outer ClientHello are irrelevant; they can be chosen to look innocuous, chosen at random, or copied from the inner ClientHello as the implementation chooses.

chris-wood:
@davidben, does this seem reasonable to you?

draft-ietf-tls-esni.md
ekr: functionality?

draft-ietf-tls-esni.md
```diff
@@ -934,6 +937,17 @@ IANA is requested to create an entry, echo_required(121) in the
 existing registry for Alerts (defined in {{!RFC8446}}), with the
 "DTLS-OK" column being set to "Y".
 
+# ECHOConfig Extension Guidance {#config-extensions}
+
+Any future information or hints that influence the outer ClientHello SHOULD be
+specified as ECHOConfig extensions. This is primarily because the outer ClientHello
```
ekr, line 943: Or, I guess, as ECHOConfigV2, right?

draft-ietf-tls-esni.md
```diff
@@ -934,6 +937,17 @@ IANA is requested to create an entry, echo_required(121) in the
...
+exists only in support of ECHO. Namely, it is both an envelope for the encrypted
+inner ClientHello and enabler for authenticated key mismatch signals
+(see {{server-behavior}}). In contrast, the inner ClientHello is the true ClientHello
+used upon ECHO negotiation. Any information that influences the inner ClientHello,
+if needed, therefore belongs elsewhere, such as the corresponding HTTPSSVC
+record {{!HTTPSSVC}}.
+
```
ekr, line 950: I'm not sure there is a need to be this directive here. Why not leave this for future work.

Also, hypothetically, suppose that the extension was like "make the nonce twice as long"

draft-ietf-tls-esni.md
chris-wood: Functionality is a bit more general -- let's go with that!

draft-ietf-tls-esni.md
chris-wood: Yeah, I'll add that: "as ECHOConfig extensions, or an entirely new version of ECHOConfig."

draft-ietf-tls-esni.md
chris-wood: That's fair (and a good point). It is a bit... restrictive. I just removed the sentence.

PR #230, opened 2020-05-02
Add information about per-client tracking via unique ECHOConfig structures

chris-wood:
Addresses #146.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -814,7 +814,7 @@ way by observing TLS connection attempts.
 The cost of this type of attack scales linearly with the desired number of target
 clients. Moreover, DNS caching behavior makes targeting individual users for extended
 periods of time, e.g., using per-client ECHOConfig structures delivered via HTTPSSVC
-RRs with high TTLs, challenging. Also, clients can help mitigate this problem by
+RRs with high TTLs, challenging. Clients can help mitigate this problem by
 flushing any DNS or ECHOConfig state upon changing networks.
 
 ## Optional Record Digests and Trial Decryption
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -809,11 +809,13 @@ less useful without encryption of DNS queries in transit via DoH or DPRIVE mecha
 A malicious client-facing server could distribute unique, per-client ECHOConfig
 structures as a way of tracking clients across subsequent connections. On-path
 adversaries which know about these unique keys could also track clients in this
-way by observing TLS connection attempts. This type of attack scales linearly
-with the desired number of target clients. Also, DNS caching behavior makes
-distribution of per-client ECHOConfig structures via HTTPSSVC RRs with high TTLs
-challenging. Moreover, given the existence of other tracking vectors, such as
-IPv6 addresses, this does not pose an immediate threat to clients.
+way by observing TLS connection attempts.
+
+The cost of this type of attack scales linearly with the desired number of target
+clients. Moreover, DNS caching behavior makes targeting individual users for extended
+periods of time, e.g., using per-client ECHOConfig structures delivered via HTTPSSVC
+RRs with high TTLs, challenging. Also, clients can help mitigate this problem by
+flushing any ECHOConfig state upon changing networks.
 
 ## Optional Record Digests and Trial Decryption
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -804,6 +804,17 @@ by attackers on the local network, which is a common case where ClientHello and
 encryption are desired. Moreover, as noted in the introduction, SNI encryption is
 less useful without encryption of DNS queries in transit via DoH or DPRIVE mechanisms.
 
+## Client Tracking
+
+A malicious client-facing server could distribute unique, per-client ECHOConfig
+structures as a way of tracking clients across subsequent connections. On-path
+adversaries which know about these unique keys could also track clients in this
+way by observing TLS connection attempts. This type of attack scales linearly
```
martinthomson, line 812: A paragraph break here might be good. You jump from description of the attack to a description of mitigating factors without segue.

draft-ietf-tls-esni.md
```diff
@@ -804,6 +804,17 @@ by attackers on the local network, which is a common case where ClientHello and
...
+with the desired number of target clients. Also, DNS caching behavior makes
+distribution of per-client ECHOConfig structures via HTTPSSVC RRs with high TTLs
+challenging. Moreover, given the existence of other tracking vectors, such as
+IPv6 addresses, this does not pose an immediate threat to clients.
```
martinthomson, line 816: I don't think the v6 example is good. And none of these are especially strong, except perhaps the DNS one. 

For me, the question is how targeted the delivery could be, and whether the client can then do anything about that. For instance, a client might decide to flush its ECHO config cache if it moves and wants to be unlinkable.

chris-wood: @martinthomson I'll remove the last sentence, not because of v6 (which I do think is a good example), but because it can probably bite us in the future when someone does turn it into a threat. :)

On the latter points, how about the following:

~~~
The cost of this type of attack scales linearly with the desired number of target clients. DNS caching behavior makes targeting individual users for extended periods of time, e.g., using per-client ECHOConfig structures delivered via HTTPSSVC RRs with high TTLs, challenging. Moreover, clients can flush any ECHOConfig state upon changing networks to mitigate potential linkability problems.
~~~

draft-ietf-tls-esni.md
```diff
@@ -804,6 +804,19 @@ by attackers on the local network, which is a common case where ClientHello and
...
+A malicious client-facing server could distribute unique, per-client ECHOConfig
+structures as a way of tracking clients across subsequent connections. On-path
+adversaries which know about these unique keys could also track clients in this
+way by observing TLS connection attempts.
+
+The cost of this type of attack scales linearly with the desired number of target
+clients. Moreover, DNS caching behavior makes targeting individual users for extended
+periods of time, e.g., using per-client ECHOConfig structures delivered via HTTPSSVC
+RRs with high TTLs, challenging. Also, clients can help mitigate this problem by
+flushing any ECHOConfig state upon changing networks.
```
davidben, line 818: Is the thinking here that the client might have some ECHOConfig state outside the DNS cache, or is this just saying the client could flush its DNS cache on network change? An attacker that can distribute per-client ECHOConfigs in DNS could just as easily distribute per-client A/AAAA records in DNS. (Well, A may be hard due to address scarcity, but AAAA should be doable.) So if you're concerned about this attack, you probably want to treat this as a DNS-wide issue and drop the entire DNS cache.

chris-wood: Indeed -- in practice it'll likely be the entire cache. I didn't want to include other state (V4 or V6) to make any assumption on what the cache might contain, be it the DNS cache or something else. Is this just too generic for your taste?

davidben: Mostly a clarifying question. Though perhaps we should say "flushing any DNS state" instead of flushing any ECHOConfig state". Or perhaps "flushing any DNS or ECHOConfig state". Or maybe we leave it along because too many words. :-)

chris-wood: How about: "flushing any ECHOConfig state, which may the include DNS cache"?

davidben: SGTM

draft-ietf-tls-esni.md
```diff
@@ -804,6 +804,19 @@ by attackers on the local network, which is a common case where ClientHello and
...
+
+A malicious client-facing server could distribute unique, per-client ECHOConfig
+structures as a way of tracking clients across subsequent connections. On-path
+adversaries which know about these unique keys could also track clients in this
+way by observing TLS connection attempts.
+
+The cost of this type of attack scales linearly with the desired number of target
+clients. Moreover, DNS caching behavior makes targeting individual users for extended
+periods of time, e.g., using per-client ECHOConfig structures delivered via HTTPSSVC
+RRs with high TTLs, challenging. Also, clients can help mitigate this problem by
```
ekr, line 817:
```suggestion
RRs with high TTLs, challenging. Clients can help mitigate this problem by
```

Issue #229, opened 2020-05-02
Describe HRR cut-and-paste 1 in Security Considerations.


chris-wood:
<img width="909" alt="Screen Shot 2020-05-02 at 3 33 05 PM" src="https://user-images.githubusercontent.com/2172048/80893782-3fae8980-8c8a-11ea-8bca-cb5fc1e33685.png">

chris-wood:
Addressed in #234.

PR #234, opened 2020-05-12
Describe known active attacks

chris-wood:
Addresses #229.

This is for motivating some of our design decisions (and documenting what we know along the way). I meant to do this a while ago, but never got around to it. In particular, it describes why we have the following three bindings:

1. Transcript to ECHO secret (to prevent client reaction attacks).
2. CH1 to CH2 (to prevent HRR oracles).
3. ClientHello to resumption PSK binder (to prevent resumption oracles).

I'm *sure* the language and presentation can be improved, so suggestions to that effect are very much welcome!

cc @martinthomson, @davidben, @dmcardle, @jhoyla, @kazuho, @karthikbhargavan, @grittygrease

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -341,9 +341,7 @@ When using ECHO, the client MUST also add an extension of type
 "echo_nonce" to the ClientHelloInner (but not to the outer
 ClientHello). This nonce ensures that the server's encrypted
 Certificate can only be read by the entity which sent this
-ClientHello. [[TODO: Describe HRR cut-and-paste 1 in
-Security Considerations.]]
-This extension is defined as follows:
+ClientHello. This extension is defined as follows:
 
 ~~~
    enum {
@@ -621,7 +619,8 @@ echo_nonce = context.Export("tls13-echo-hrr-nonce", 16)
 ~~~
 
 Clients then encrypt the second ClientHelloInner using this new HPKE context.
-In doing so, the encrypted value is also authenticated by echo_hrr_key.
+In doing so, the encrypted value is also authenticated by echo_hrr_key. The rationale
+for this is described in {{flow-hrr-hijack}}.
 
 Client-facing servers perform the corresponding process when decrypting second
 ClientHelloInner messages. In particular, upon receipt of a second ClientHello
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -963,6 +962,135 @@ Variations in the length of the ClientHelloInner ciphertext could leak informati
...
+                 <------
+   Alert
+                 ------>
+~~~
+{: #flow-diagram-client-reaction title="Client reaction attack"}
+
+The "echo_nonce" extension in the inner ClientHello prevents this attack. In particular,
+since the attacker does not have access to this value, it cannot produce the right transcript
+and handshake keys needed for encrypting the Certificate message. Thus, the client will fail
+to decrypt before any verification check occurs.
```
ekr, line 1006:
```suggestion
to decrypt the Certificate and abort the connection.
```

draft-ietf-tls-esni.md
```diff
@@ -963,6 +962,135 @@ Variations in the length of the ClientHelloInner ciphertext could leak informati
...
+
+### HelloRetryRequest Hijack Mitigation {#flow-hrr-hijack}
+
+This attack aims to exploit server HRR state management to recover information about
+a legitimate ClientHello using its own adversarially-controlled ClientHello.
+To begin, the attacker intercepts and forwards a legitimate ClientHello with an
+"encrypted_client_hello" (echo) extension to the server, which triggers a HelloRetryRequest
+in return. Rather than forward the retry to the client, the attacker, attempts to
```
ekr, line 1014: I think you need to say something about how this had to legitimately be an HRR case. Or can the attacker force it?

draft-ietf-tls-esni.md
```diff
@@ -963,6 +962,135 @@ Variations in the length of the ClientHelloInner ciphertext could leak informati
...
+generates its own ClientHello in response based on the contents of the first ClientHello
+and HelloRetryRequest exchange.
```
ekr, line 1016: Note that this only works if the server (arguably incorrectly) stores the SNI from CH1.

ekr, line 1016:
```suggestion
and HelloRetryRequest exchange with the result that the server encrypts the certificate to the attacker.
```

draft-ietf-tls-esni.md
```diff
@@ -963,6 +962,135 @@ Variations in the length of the ClientHelloInner ciphertext could leak informati
...
+This attack relies on how clients react to SNI and certificate mismatch as an oracle.
+In particular, the attacker intercepts a legitimate ClientHello and replies with a
+ServerHello, Certificate, CertificateVerify, and Finished messages, wherein the Certificate
+message contains a "test" certificate for the domain name it wishes to query. If the client
+decrypted the Certificate and failed verification (or leaked information about its verification
+process by a timing side channel), the attacker learns that its test certificate name
+was incorrect. As an example, suppose the client's SNI value in its inner ClientHello is
+"example.com," and the attacker replied with a Certificate for "test.com". If the client
+produces a verification failure alert because of the mismatch faster than it would due to
+the Certificate signature validation, information about the name leaks.
```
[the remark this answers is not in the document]
chris-wood: That seems like a different variation, so I'll just note it. Thanks!

draft-ietf-tls-esni.md
chris-wood: Good point. I'll call out that assumption explicitly.

draft-ietf-tls-esni.md
chris-wood: Yep! I'll note that, too.

chris-wood:
Thanks for the feedback! I incorporated your suggestions and clarified a few more things. I'm merging as-is since they seemed mostly editorial.

PR #235, opened 2020-05-15
Suggest size of GREASE encrypted_ch extension value.

chris-wood:
This assumes the GREASE threat model is a misbehaving and lazy middlebox. That is, we aim to prevent ossification. It does nothing more to prevent an active attacker from learning whether or not a particular connection used GREASE. (That problem is probably better suited to something like MASQUE.)

cc @davidben @ekr @martinthomson


chris-wood:
Merging for now. We can come back later and address followup extensions if needed!

PR #236, opened 2020-05-21
s/ECHO/ECH

### LINKED ISSUE #232: Bikeshed: finalize on naming of "ECHO"
While cutsy, the name "ECHO" for "Encrypted Client Hello" seems like it could cause problems and significant confusion.  To those not actively involved in the working group, "ECHO" has very different connotations.  Some potential risks include:
* Other IETF standards have some form of "ECHO" functionality.  When bouncing this question off someone else, they responded with "TLS ECHO, like ICMP ECHO, except totally not"
* The immediate guess people may have on hearing about a TLS ECHO functionality would be to think "oh, like to echo back a message for heartbeat purposes?".  Given the potential risk for confusion with heartbeat (especially given "Heartbleed") this has some risk of giving TLS ECHO a bad rap from the start due to confusion.
* This functionality is going to have high profile with product managers and perhaps even government regulators and public policy.  Applying a name that has confusing lay-connotations is going to muddle conversations.
* This has a risk of confusion when used in other contexts.  For example, "echoconfig" as a SVCB parameter key has potential risk for confusing people reading the config and lacking context.
* ECHO is going to be harder to do searches for, both in Internet and intranet search engines.

Just calling it "ECH" or "EHELLO" may be safer?

(If this changes, SVCB will also need to change the "echoconfig".)

chris-wood:
Closes #232.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -1,6 +1,6 @@
 ---
-title: Encrypted Server Name Indication for TLS 1.3
-abbrev: TLS 1.3 SNI Encryption
+title: Encrypted ClientHello for TLS 1.3
+abbrev: TLS 1.3 Encrypted ClientHello
 docname: draft-ietf-tls-esni-latest
 category: exp
 
@@ -44,8 +44,7 @@ normative:
 
 --- abstract
 
-This document defines a simple mechanism for encrypting the
-Server Name Indication for TLS 1.3.
+This document defines a mechanism for encrypting a TLS 1.3 ClientHello.
 
 --- middle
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -1,6 +1,6 @@
 ---
-title: Encrypted Server Name Indication for TLS 1.3
-abbrev: TLS 1.3 SNI Encryption
+title: Encrypted ClientHello for TLS 1.3
+abbrev: TLS 1.3 Encrypted ClientHello
```
tfpauly, line 3: TLS Encrypted ClientHello —> TECH!

(Kidding, kidding)

chris-wood: Hah!

martinthomson: Seeing this, I think that you should remove 1.3 from the title.  Sure, this is 1.3 only, but I think that this design is fine for 1.4 also.  If it isn't, I'm sure we can deal with that eventually.  I don't want to find that we ship a 1.4 (or 4) that is compatible with this and have people tell me that it isn't compatible.

chris-wood: That's a fair point. I'll make that change in a separate PR!

PR #238, opened 2020-05-26
Remove TLS 1.3 from the title, and re-categorize as a ST document.

chris-wood:
Seeing no pushback on the list, this moves the draft from Experimental to ST.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -1,6 +1,6 @@
 ---
-title: Encrypted ClientHello for TLS
-abbrev: TLS Encrypted ClientHello
+title: TLS Encrypted Client Hello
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -1,8 +1,8 @@
 ---
-title: Encrypted ClientHello for TLS 1.3
-abbrev: TLS 1.3 Encrypted ClientHello
+title: Encrypted ClientHello for TLS
```
ekr, line 2: If we are going to change the title, let's make it match the abbrev

"TLS Encrypted Client Hello"

chris-wood: Works for me!

PR #242, opened 2020-05-31
Revert to experimental codepoints.

chris-wood:
We'll request early codepoints after we get some implementation experience.


chris-wood:
> I have no objection to echo_nonce -> echo_nonce_value but did you mean to do it in this PR?

Yeah, as I was specifying the extension codepoint I noticed the collision between this variable, so I figured changing it here was best.

PR #243, opened 2020-06-04
empty or random record digest for privacy

huitema:
As discussed on the list, add a small bit of text in the section 10.3 to outline the behavior of clients seeking to protect server's privacy.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -842,8 +842,12 @@ flushing any DNS or ECHConfig state upon changing networks.
 
 Optional record digests may be useful in scenarios where clients and client-facing
 servers do not want to reveal information about the client-facing server in the
-"encrypted_client_hello" extension. In such settings, servers must perform trial
-decrypt upon receipt of an empty record digest, which may exacerbate DoS attacks.
+"encrypted_client_hello" extension. In such settings, clients  will be sending
+either an empty record_digest field or a randomly generated record_digest field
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -842,8 +842,12 @@ flushing any DNS or ECHConfig state upon changing networks.
...
 Optional record digests may be useful in scenarios where clients and client-facing
 servers do not want to reveal information about the client-facing server in the
-"encrypted_client_hello" extension. In such settings, servers must perform trial
-decrypt upon receipt of an empty record digest, which may exacerbate DoS attacks.
+"encrypted_client_hello" extension. In such settings, clients  will be sending
+either an empty record_digest field or a randomly generated record_digest field
+in the ClientEncryptedCH. (The precise implementation choices for this privacy
+mechanism is out of scope for this document.) Servers
+in these settings cannot rely on the record_digest value to identify the key
+used by the client. They must perform trial decrypt, which may exacerbate DoS attacks.
```
chris-wood, line 850:
```suggestion
"encrypted_client_hello" extension. In such settings, clients send 
either an empty record_digest or a randomly generated record_digest
in the ClientEncryptedCH. (The precise implementation choice for this
mechanism is out of scope for this document.) Servers in these settings 
must perform trial decryption since they cannot identify the client's chosen 
ECH key using the record_digest value. As a result, support for optional 
record digests may exacerbate DoS attacks.
```

Issue #245, opened 2020-06-17
Describe risks and mitigations for privacy leaks from configuration digest


sharpie7:
The ID proposes an optional configuration digest parameter to be included in the clientHello. This parameter is advantageous because it improves the efficiency of the solution by avoiding the server having to do trial decryptions. However, in some scenarios the configuration digest could leak privacy information. In version 7 of the ID, section 10.2 on “Client Tracking” describes one scenario, but there are others. A more complete and general discussion of the risks and mitigations would be useful.

If a particular domain has a unique public key then the configuration digest becomes a fingerprint for that domain. An attacker that has a list of such domains that they want to monitor can build a dictionary of the corresponding digests use this to identify traffic to the domains on their list.

Some mitigations could be:
1)	Avoid the use of configuration digest
2)	Only deploy in a mode where many domains share the same public key to make finger-printing too general to be useful
3)	To change keys faster than attackers can update their list of fingerprints

chris-wood:
This should be addressed by #247. Please re-open (and suggest text edits!) if you think otherwise.

PR #247, opened 2020-06-17
Expand text around security and privacy goals.

chris-wood:
Addresses #245.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -815,10 +815,9 @@ of the anonymity set where possible. This means client-facing servers should use
 ECHConfig for as many hosts as possible. An attacker can distinguish two hosts that
 have different ECHConfig values based on the ClientEncryptedCH.record_digest value.
 This also means public information in a TLS handshake is also consistent across hosts.
-For example, if a client-facing server services k backend origin hosts, and exactly one of
-those backend origin hosts supports a different TLS configuration, such as a different
-set of cryptographic algorithms than the other (k - 1) hosts, it may be possible to
-identify this single host based on the contents of unencrypted handshake messages.
+For example, if a client-facing server services many backend origin hosts, only one of which
+supports some cipher suite, it may be possible to identify that host based on the contents
+of unencrypted handshake messages.
 
 ## Unauthenticated and Cleartext DNS {#cleartext-dns}
 
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -803,22 +803,21 @@ Informally, the primary security and privacy goals of ECH against an active atta
 as follows:
 
 1. Use of ECH does not weaken the security properties of TLS without ECH.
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -804,7 +797,30 @@ connection in this case.
...
+## Security and Privacy Goals
+
+Informally, the primary security and privacy goals of ECH against an active attacker are
+as follows:
+
+1. Use of ECH does not weaken the security properties of TLS without ECH.
+2. TLS connection establishment to a host with a specific ECHConfig is indistinguishable
+from a connection to any other host with the same ECHConfig. That is, given an anonymity
+set of servers of size k sharing the same ECHConfig, an attacker has negligible advantage
+in correctly identifying the host for a given connection with probability larger than 1/k.
```
davidben, line 809: Same ECHConfig is necessary, but not sufficient I think. If the hosts share an ECHConfig but have not other TLS config (the deleted paragraph above), the hosts are distinguishable. (I've lost track of whether the current level of ClientHello binding is sufficient to limit this to anonymity within the negotiated values or the full config...) The third paragraph mentions this, but it seems to be clarifying point (2), but point (2) doesn't actually imply this.

chris-wood: Fair :) I was trying to get away without too much text. I can clarify that this is necessary (and rework text accordingly).

chris-wood: @davidben, reworked. Please have another look!

draft-ietf-tls-esni.md
```diff
@@ -804,7 +797,30 @@ connection in this case.
...
+if a client-facing server uses distinct ECHConfig values for each host, then each anonymity set
+has size k = 1. Client-facing servers SHOULD deploy ECH in such a way so as to maximize the size
+of the anonymity set where possible. This means client-facing servers should use the same
+ECHConfig for as many hosts as possible. An attacker can distinguish two hosts that
+have different ECHConfig values based on the ClientEncryptedCH.record_digest value.
+This also means public information in a TLS handshake is also consistent across hosts.
+For example, if a client-facing server services k backend origin hosts, and exactly one of
+those backend origin hosts supports a different TLS configuration, such as a different
+set of cryptographic algorithms than the other (k - 1) hosts, it may be possible to
+identify this single host based on the contents of unencrypted handshake messages.
```
davidben, line 821: Super nitpicky nitpick: we can probably trim the "For example, [...]" sentence down to:

For example, if a client-facing server services many backend origin hosts, only one of which supports some cipher suite, it may be possible to identify that host based on the contents of unencrypted handshake messages.

(It's already giving an example, and being more specific uses fewer words.)

chris-wood: Works for me -- thanks!

sharpie7:
Thanks. The changes look to cover my point in #245.

PR #258, opened 2020-08-11
Fix superfluous padding edge cases.

chris-wood:
Thanks, @bemasc, for raising this!


bemasc:
Thanks for the improvement.  Note that this formula still has no length ambiguity when L == 31 and D == 32.  It's not clear to me whether that's consistent with the design goal.

chris-wood:
Can you please elaborate or suggest something different (in a PR)?

chris-wood:
Closing.

PR #268, opened 2020-08-15
Fix superfluous padding edge cases.

### LINKED ISSUE #252: Avoid padding cliff
Section 7.2 says
> 1. If L > D, add L - D bytes of padding. This rounds to the server's advertised hint, i.e., ECHConfig.maximum_name_length.
> 2. Otherwise, add 32 - (D % 32) bytes of padding. This rounds D up to the nearest multiple of 32 bytes.

This has some problems:

1. If L == D, or D > L and D % 32 == 0, this results in unnecessary padding.
2. If L == D == 31, there is no length protection.  The output is 32, and that is the only length that produces that output.

The easiest solution is to replace "32 - (D % 32)" with "(L - D) & 0x1f".  This pads up to the nearest L + 32*n.

If it's considered important to pad to a multiple of 32, I would suggest ensuring that the initial bucket is at least 32, e.g. 

> 1. If L >= D, add L - D bytes of padding. This rounds to the server's advertised hint, i.e., ECHConfig.maximum_name_length.
> 2. Otherwise, add 31 - ((D - 1) % 32) bytes of padding, plus an additional 32 bytes if L > 0 and D < L + 31 - ((L - 1) % 32)

chris-wood:
Consider the following values for L and D, where L is the server's hint
and D is the length of the domain name.

~~~
    D1     D2      D3
... -+--|---+---|---+---|----->
        L     32*n   32*(n+1)
~~~

This change pads D according to the following rules.

- D1: D <= L, so round up to L.
- D2: 32*(n-1) <= L < D <= 32*n, round up to 32*(n+1).
- D3: 32*(n-1) <= L <= 32*n < D, round up to 32*(n+1).

Case D2 exists because the anonymity set of of D, if rounded to 32*n
(instead of 32*(n+1), is less than 32. The new padding algorithm ensures
that the anonymity set for D whenever D > L is at least 32 elements.

Closes #252. Thanks, @bemasc!

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -488,10 +488,11 @@ server's anonymity set without server input. For the "server_name" extension
 with length D, clients SHOULD use the server's length hint L
 (ECHCOnfig.maximum_name_length) when computing the padding as follows:
 
-1. If L > D, add L - D bytes of padding. This rounds to the server's advertised
-   hint, i.e., ECHConfig.maximum_name_length.
-2. Otherwise, add 32 - (D % 32) bytes of padding. This rounds D up to the
-   nearest multiple of 32 bytes.
+1. If L >= D, add L - D bytes of padding. This rounds to the server's
+   advertised hint, i.e., ECHConfig.maximum_name_length.
+2. Otherwise, let P = 31 - ((D - 1) % 32), and add P bytes of padding, plus an
+   additional 32 bytes if D + P < L + 32. This rounds D up to the nearest
+   multiple of 32 bytes.
 
 In addition to padding ClientHelloInner, clients and servers will also need to
 pad all other handshake messages that have sensitive-length fields. For example,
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -488,10 +488,11 @@ server's anonymity set without server input. For the "server_name" extension
...
 
-1. If L > D, add L - D bytes of padding. This rounds to the server's advertised
-   hint, i.e., ECHConfig.maximum_name_length.
-2. Otherwise, add 32 - (D % 32) bytes of padding. This rounds D up to the
-   nearest multiple of 32 bytes.
+1. If L >= D, add L - D bytes of padding. This rounds to the server's
+   advertised hint, i.e., ECHConfig.maximum_name_length.
+2. Otherwise, let P = 31 - ((D - 1) % 32), and add P bytes of padding, plus an
+   additional 32 bytes if D + P < L + 32. This rounds D up to the nearest
+   multiple of 32 bytes.
```
bemasc, line 495: Technically this no longer "rounds D up to the nearest multiple of 32 bytes".  How about
```suggestion
   multiple of 32 bytes that provides at least 32 bytes of length ambiguity.
```

chris-wood: Better — thanks!

chris-wood:
cc @davidben, @kjacobs-moz, @dmcardle, @ekr, @martinthomson

PR #269, opened 2020-08-17
Proposal: Retry-only ECHConfig

bemasc:
A server operator always has the option to operate in a retry-only mode, which may be necessary if updating the published ECHConfig is difficult and long-lived keys are undesirable.  This change improves forward-secrecy in that configuration.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -479,6 +479,9 @@ particular,
 - encrypted_ch contains the HPKE encapsulated key (enc) and the ClientHelloInner
   ciphertext (encrypted_ch_inner).
 
+If `ECHConfig.public_key` is empty, the client SHOULD instead compute
+`config_digest`, `enc`, and `encrypted_ch` randomly, as in {{grease-extensions}}.
+
 The client MUST place the value of `ECHConfig.public_name` in the
 ClientHelloOuter "server_name" extension. The ClientHelloOuter MUST NOT contain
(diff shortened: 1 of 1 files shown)


davidben:
Are there folks looking to deploy such a thing? The current retry flow costs 2-RTT (TCP + TLS) on top of a normal handshake. I don't think that's something to encourage. It's currently meant as a recovery flow. I think something meant to be used more frequently would want to look different and, in particular, stay in one connection to keep the overhead at 1-RTT. Possibly even with some client state for subsequent connections, though squaring that with multi-CDN is tricky.

That sort of thing came up on the list, but IIRC folks weren't very excited about the complexity involved and we kept it as a simpler, recovery-only story.

bemasc:
I agree; it's too expensive for normal use, and not worth complicating clients to improve performance.

I'm thinking of use cases where DNS is not the distribution channel, and updating the ECHConfig is difficult.  It would be nice to be able to see that the ECHConfig is not "pinning" a key, and to have a shorter ECHConfig for very constrained distribution channels (e.g. a QR code).

A client like this might fetch and cache real ECHConfigs once it's connected, but I think we can leave that to the application layer.

cjpatton:
I like this idea!

chris-wood:
From the interim: we need more clarification of the use case. This can also be addressed in future work as an extension. @bemasc, can we close this and punt to a future change?

bemasc:
I don't think an extension would allow us to _omit_ information as suggested here, but it sounds like there's no consensus to support this so I don't want to create delay.

Issue #266, opened 2020-08-13
Security considerations: SNI leakage vectors


cjpatton:
Section 7.1 provides a bit of guidance on how ECH might interact with other extensions:

> The client MUST place the value of ECHConfig.public_name in the ClientHelloOuter
"server_name" extension. The ClientHelloOuter MUST NOT contain a "cached_info"
extension {{!RFC7924}} with a CachedObject entry whose CachedInformationType is
"cert", since this indication would divulge the true server name. The remaining
contents of the ClientHelloOuter MAY be identical to those in ClientHelloInner
but MAY also differ.

I think it's worth expanding on this in "Security Considerations". There may be other ways that private CH extensions get leaked, now or in the future.

davidben:
Now that we've switched to encrypting an entire ClientHello, I think the text wants to say something else anyway: if you want ECH + cached info, you simply include the cached_info extension in the ClientHelloInner, not the ClientHelloOuter. There wouldn't be much point in sending a true name cached_info in ClientHelloOuter since the true server would ignore it. Sending a public name cached_info in ClientHelloOuter is also plausible and also needn't be forbidden (it reveals information about the public name, which is public). It's also probably not that useful since it's just a recovery mechanism, but maybe to avoid sticking out, for clients that normally send it?

More generally, anything dependent on the inner name goes in ClientHelloInner and not ClientHelloOuter.

davidben:
I took a stab at replacement text in https://github.com/tlswg/draft-ietf-tls-esni/pull/270. Thoughts?

cjpatton:
I'm happy with it.

PR #270, opened 2020-08-17
Adjust and generalize cached_info guidance.

davidben:
Now that we've switched to encrypting the entire ClientHello, there's
nothing wrong with cached_info in ClientHelloOuter, provided it's based
on the public name rather than the inner name. And, more generally,
name-specific extensions take some care.

Replace the prohibition with more general guidance in Security
Considerations, with cached_info merely being an example.

This addresses issue #266.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -889,7 +889,7 @@ which would reveal a sensitive ClientHelloInner property.
 
 In particular, some extensions require the client send a server-name-specific
 value in the ClientHello. These values may reveal information about the
-encrypted server name. For example, the "cached_info" ClientHello extension
+true server name. For example, the "cached_info" ClientHello extension
 {{?RFC7924}} can contain the hash of a previously observed server certificate.
 The client SHOULD NOT send values associated with the true server name in the
 ClientHelloOuter. It MAY send such values in the ClientHelloInner. It also MAY
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -885,7 +885,9 @@ caused by such attacks.
 
 Any information that the client includes in the ClientHelloOuter is visible to
 passive observers. The client SHOULD NOT send values in the ClientHelloOuter
-which would reveal a sensitive ClientHelloInner property.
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -884,6 +881,24 @@ in order to force wasteful decryption. Servers that support this feature should,
...
 
+## Outer ClientHello
+
+Any information that the client includes in the ClientHelloOuter is visible to
+passive observers. The client SHOULD NOT send values in the ClientHelloOuter
+which would reveal a sensitive ClientHelloInner property.
+
+In particular, some extensions require the client send a server-name-specific
+value in the ClientHello. These values may reveal information about the
+encrypted server name. For example, the "cached_info" ClientHello extension
```
chris-wood, line 892:
```suggestion
true server name. For example, the "cached_info" ClientHello extension
```

draft-ietf-tls-esni.md
```diff
@@ -884,6 +881,24 @@ in order to force wasteful decryption. Servers that support this feature should,
...
+{{?RFC7924}} can contain the hash of a previously observed server certificate.
+The client SHOULD NOT send values associated with the true server name in the
+ClientHelloOuter. It MAY send such values in the ClientHelloInner. It also MAY
+send values associated with the public name in the ClientHelloOuter.
```
chris-wood, line 896:
```suggestion
ClientHelloOuter. It MAY send such values in the ClientHelloInner.
```

draft-ietf-tls-esni.md
```diff
@@ -884,6 +881,24 @@ in order to force wasteful decryption. Servers that support this feature should,
 for example, implement some form of rate limiting mechanism to limit the damage
 caused by such attacks.
 
+## Outer ClientHello
+
+Any information that the client includes in the ClientHelloOuter is visible to
+passive observers. The client SHOULD NOT send values in the ClientHelloOuter
+which would reveal a sensitive ClientHelloInner property.
```
chris-wood, line 888:
```suggestion
which would reveal a sensitive ClientHelloInner property, such as the true
server name. It MAY send values associated with the public name in the 
ClientHelloOuter.
```

PR #305, opened 2020-09-22
Rewrite security and privacy goals

### LINKED ISSUE #260: Need a high-level description of the security goal in the intro
In just a few sentences, let's define the anonymity set and say what it means for connections to end points in this set to be indistignuishable. We might also enumerate non-goals (though we still need consensus on what these are).

chris-wood:
Closes #260.

cc @cjpatton

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -49,7 +49,7 @@ encrypting a ClientHello message under a server public key.
 
 --- middle
 
-# Introduction
+# Introduction {#intro}
 
 DISCLAIMER: This is very early a work-in-progress design and has not yet seen
 significant (or really any) security analysis. It should not be used as a basis
@@ -85,10 +85,12 @@ both.
 The protocol specified by this document takes a different approach. It assumes
 that private origins will co-locate with or hide behind a provider (reverse
 proxy, application server, etc.) that protects SNIs for all of the domains it
-hosts. As a result, SNI protection does not indicate that the client is
-attempting to reach a private origin, but only that it is going to a particular
-service provider, which the observer could already tell from the visible IP
-address.
+hosts. These co-located servers form an anonymity set wherein all elements
+have a consistent configuration, e.g., set of supported application protocols,
+ciphersuites, TLS versions, and so on. SNI protection reveals that a client is
+connecting to a particular service provider without revealing the specific
+private server from the anonymity set. Thus, it leaks no more than what is
+already visible from the server IP address.
 
 This document specifies a new TLS extension, called Encrypted Client Hello
 (ECH), that allows clients to encrypt their ClientHello to a supporting server.
@@ -193,6 +195,12 @@ following actions:
 Upon receiving the server's response, the client determines whether ECH was
 accepted or rejected and proceeds with the handshake accordingly. (See
 {{client-behavior}} for details.)
+
+Informally, a primary goal of ECH is ensuring that connections to servers in the
+same anonymity set are indistinguishable from one another for passive and active
+attackers, modulo traffic analysis, without any existing security properties of
+TLS 1.3. See {{goals}} for more details about the ECH security and privacy
+goals.
 
 # Encrypted ClientHello Configuration {#ech-configuration}
 
@@ -819,13 +827,20 @@ Clients SHOULD NOT attempt to repair the connection in this case.
 
 # Security Considerations
 
-## Security and Privacy Goals
+## Security and Privacy Goals {#goals}
 
-Informally, the primary security and privacy goals of ECH against an active
-attacker are as follows:
+ECH considers two types of attackers: passive and active. Passive attackers can
+read packets from the network. They cannot perform any sort of active behavior
+such as probing servers or querying DNS. A middlebox that filters based on
+plaintext packet contents is one example of a passive attacker. In contrast,
+active attackers can write packets into the network for malicious purposes, such
+as interfering with existing connections, probing servers, and querying DNS. A
+censor is one example of an active attacker.
+
+Given these types of attackers, the primary goals of ECH are as follows.
 
 1. Use of ECH does not weaken the security properties of TLS without ECH.
-2. TLS connection establishment to a host with a specific ECHConfig and TLS
+1. TLS connection establishment to a host with a specific ECHConfig and TLS
    configuration is indistinguishable from a connection to any other host with
    the same ECHConfig and TLS configuration. (The set of hosts which share the
    same ECHConfig and TLS configuration is referred to as the anonymity set.)
@@ -841,6 +856,28 @@ public information in a TLS handshake is also consistent across hosts. For
 example, if a client-facing server services many backend origin hosts, only one
 of which supports some cipher suite, it may be possible to identify that host
 based on the contents of unencrypted handshake messages.
+
+Beyond these primary security and privacy goals, ECH also aims to hide, to some
+extent, (a) whether or not a specific server supports ECH and (b) whether or
+not a specific connection uses or negotiates ECH. In particular:
+
+1. ECH support is secret from passive attackers who do not know the ECH
+configuration, but not active attackers.
+1. ECH usage is secret from passive attackers who do not know the ECH
+configuration, but not active attackers. More specifically:
+
+- Passive attackers with a known ECH configuration can distinguish between a
+connection that negotiates ECH with that configuration and one which does not,
+i.e., because the latter used a GREASE "encrypted_client_hello" extension or a
+different ECH configuration.
+- Passive attackers without the ECH configuration cannot distinguish between a
+connection that negotiates ECH and one which uses a GREASE
+"encrypted_client_hello" extension.
+- Active attackers can distinguish between a connection that negotiates ECH and
+one which uses a GREASE "encrypted_client_hello" extension.
+
+As a result, ECH is not suitable for bypassing active adversaries such as
+censors.
 
 ## Unauthenticated and Plaintext DNS {#plaintext-dns}
 
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -87,8 +87,8 @@ that private origins will co-locate with or hide behind a provider (reverse
 proxy, application server, etc.) that protects sensitive ClientHello parameters,
 including the SNI, for all of the domains it hosts. These co-located servers
 form an anonymity set wherein all elements have a consistent configuration,
-e.g., the set of supported application protocols, ciphersuites, TLS versions, and
-so on. Usage of this mechanism reveals that a client is connecting to a
+e.g., the set of supported application protocols, ciphersuites, TLS versions,
+and so on. Usage of this mechanism reveals that a client is connecting to a
 particular service provider, but does not reveal which server from the anonymity
 set terminates the connection. Thus, it leaks no more than what is already
 visible from the server IP address.
@@ -198,8 +198,9 @@ was accepted and proceeds with the handshake accordingly. (See
 {{client-behavior}} for details.)
 
 Informally, a primary goal of ECH is ensuring that connections to servers in the
-same anonymity set are indistinguishable from one another without affecting any existing security properties of TLS 1.3. See
-{{goals}} for more details about the ECH security and privacy goals.
+same anonymity set are indistinguishable from one another without affecting any
+existing security properties of TLS 1.3. See {{goals}} for more details about
+the ECH security and privacy goals.
 
 # Encrypted ClientHello Configuration {#ech-configuration}
 
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -84,13 +84,14 @@ both.
 
 The protocol specified by this document takes a different approach. It assumes
 that private origins will co-locate with or hide behind a provider (reverse
-proxy, application server, etc.) that protects SNIs for all of the domains it
-hosts. These co-located servers form an anonymity set wherein all elements
-have a consistent configuration, e.g., set of supported application protocols,
-ciphersuites, TLS versions, and so on. SNI protection reveals that a client is
-connecting to a particular service provider without revealing the specific
-private server from the anonymity set. Thus, it leaks no more than what is
-already visible from the server IP address.
+proxy, application server, etc.) that protects sensitive ClientHello parameters,
+including the SNI, for all of the domains it hosts. These co-located servers
+form an anonymity set wherein all elements have a consistent configuration,
+e.g., set of supported application protocols, ciphersuites, TLS versions, and
+so on. Usage of this mechanism reveals that a client is connecting to a
+particular service provider, but does not reveal which server from the anonymity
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -85,10 +85,12 @@ both.
 The protocol specified by this document takes a different approach. It assumes
 that private origins will co-locate with or hide behind a provider (reverse
 proxy, application server, etc.) that protects SNIs for all of the domains it
-hosts. As a result, SNI protection does not indicate that the client is
-attempting to reach a private origin, but only that it is going to a particular
-service provider, which the observer could already tell from the visible IP
-address.
+hosts. These co-located servers form an anonymity set wherein all elements
```
cjpatton, line 88: "form" -> "comprise"

draft-ietf-tls-esni.md
```diff
@@ -85,10 +85,12 @@ both.
 The protocol specified by this document takes a different approach. It assumes
 that private origins will co-locate with or hide behind a provider (reverse
 proxy, application server, etc.) that protects SNIs for all of the domains it
```
cjpatton, line 87: "that protects SNIs" -> "sensitive ClientHello parameters, including the SNI,"

cjpatton, line 87: "for all of the domains" -> "on behalf of all domains"

draft-ietf-tls-esni.md
```diff
@@ -194,6 +196,12 @@ Upon receiving the server's response, the client determines whether ECH was
 accepted or rejected and proceeds with the handshake accordingly. (See
 {{client-behavior}} for details.)
 
+Informally, a primary goal of ECH is ensuring that connections to servers in the
```
cjpatton, line 199: "a" -> "the"
"is ensuring" -> "is to ensure"

draft-ietf-tls-esni.md
```diff
@@ -85,10 +85,12 @@ both.
 The protocol specified by this document takes a different approach. It assumes
 that private origins will co-locate with or hide behind a provider (reverse
 proxy, application server, etc.) that protects SNIs for all of the domains it
-hosts. As a result, SNI protection does not indicate that the client is
-attempting to reach a private origin, but only that it is going to a particular
-service provider, which the observer could already tell from the visible IP
-address.
+hosts. These co-located servers form an anonymity set wherein all elements
+have a consistent configuration, e.g., set of supported application protocols,
+ciphersuites, TLS versions, and so on. SNI protection reveals that a client is
```
cjpatton, line 90: "SNI protection" -> "Usage of this mechanism"

draft-ietf-tls-esni.md
```diff
@@ -85,10 +85,12 @@ both.
...
+connecting to a particular service provider without revealing the specific
```
cjpatton, line 91: "a particular service provider without revealing the specific private serer from the anonymity set" -> "a particular service provider, but does not reveal which server terminates the connection".

draft-ietf-tls-esni.md
```diff
@@ -194,6 +196,12 @@ Upon receiving the server's response, the client determines whether ECH was
 accepted or rejected and proceeds with the handshake accordingly. (See
 {{client-behavior}} for details.)
 
+Informally, a primary goal of ECH is ensuring that connections to servers in the
+same anonymity set are indistinguishable from one another for passive and active
+attackers, modulo traffic analysis, without any existing security properties of
```
cjpatton, line 201: "modulo traffic analysis": I would suggest cutting this and leaving it to Security Considerations. It's a matter that a sophisticated service provider might try to address, so we don't want to exclude this here.

draft-ietf-tls-esni.md
```diff
@@ -49,7 +49,7 @@ encrypting a ClientHello message under a server public key.
 
 --- middle
 
-# Introduction
+# Introduction {#intro}
```
cjpatton, line 52: I think the term "anonymity set" needs to be more clearly defined here. It warrants a short paragraph.

draft-ietf-tls-esni.md
```diff
@@ -819,13 +827,20 @@ Clients SHOULD NOT attempt to repair the connection in this case.
...
-## Security and Privacy Goals
+## Security and Privacy Goals {#goals}
+
+ECH considers two types of attackers: passive and active. Passive attackers can
+read packets from the network. They cannot perform any sort of active behavior
+such as probing servers or querying DNS. A middlebox that filters based on
+plaintext packet contents is one example of a passive attacker. In contrast,
+active attackers can write packets into the network for malicious purposes, such
+as interfering with existing connections, probing servers, and querying DNS. A
+censor is one example of an active attacker.
```
cjpatton, line 838: A more important example of an active attacker is the one that tries to break the primary security goal of ECH. A cesnor is an example of a "don't stick out" attacker. I suggest replacing the last line with something like this: 

> In short, an active attacker corresponds to the conventional threat model for TLS ({{RFC8446}}, Section ??).

draft-ietf-tls-esni.md
```diff
@@ -842,6 +857,28 @@ example, if a client-facing server services many backend origin hosts, only one
 of which supports some cipher suite, it may be possible to identify that host
 based on the contents of unencrypted handshake messages.
 
+Beyond these primary security and privacy goals, ECH also aims to hide, to some
+extent, (a) whether or not a specific server supports ECH and (b) whether or
+not a specific connection uses or negotiates ECH. In particular:
```
cjpatton, line 862: "a specific uses or negotiates ECH" -> "ECH was accepted in a particular connection".

draft-ietf-tls-esni.md
```diff
@@ -842,6 +857,28 @@ example, if a client-facing server services many backend origin hosts, only one
...
+
+1. ECH support is secret from passive attackers who do not know the ECH
```
cjpatton, line 864: I would replace this enum list with a one-liner at the end of the preceding paragraph. E.g.,:

> This protocol aims to achieve both properties, assuming the attacker is passive and does not know the set of ECH configurations offered by the server. It does not achieve these goals for active attackers.

draft-ietf-tls-esni.md
```diff
@@ -842,6 +857,28 @@ example, if a client-facing server services many backend origin hosts, only one
...
+configuration, but not active attackers.
+1. ECH usage is secret from passive attackers who do not know the ECH
+configuration, but not active attackers. More specifically:
+
+- Passive attackers with a known ECH configuration can distinguish between a
+connection that negotiates ECH with that configuration and one which does not,
+i.e., because the latter used a GREASE "encrypted_client_hello" extension or a
```
cjpatton, line 871: Add reference to {{grease-extensions}}

draft-ietf-tls-esni.md
```diff
@@ -842,6 +857,28 @@ example, if a client-facing server services many backend origin hosts, only one
 of which supports some cipher suite, it may be possible to identify that host
 based on the contents of unencrypted handshake messages.
 
+Beyond these primary security and privacy goals, ECH also aims to hide, to some
```
cjpatton, line 860: The text below should have a reference to the "don't stick out" doc.

draft-ietf-tls-esni.md
```diff
@@ -842,6 +857,28 @@ example, if a client-facing server services many backend origin hosts, only one
...
+connection that negotiates ECH with that configuration and one which does not,
+i.e., because the latter used a GREASE "encrypted_client_hello" extension or a
+different ECH configuration.
+- Passive attackers without the ECH configuration cannot distinguish between a
+connection that negotiates ECH and one which uses a GREASE
+"encrypted_client_hello" extension.
+- Active attackers can distinguish between a connection that negotiates ECH and
+one which uses a GREASE "encrypted_client_hello" extension.
+
+As a result, ECH is not suitable for bypassing active adversaries such as
```
cjpatton, line 879: I think we should elaborate on this a bit more. Readers might take issue with the claim that ECH is not suitable for censorship circumvention. In particular, there may be deployment scenarios in which keeping the ECH configuration a secret is feasible, and "don't stick out" security against active adversaries might be possible in that setting.

cjpatton (changes requested):

draft-ietf-tls-esni.md
chris-wood: We can, but I don't think we need to do that here.

cjpatton: At a minimum, I would suggest removing this line. We can have a new PR for language around censorship circumvention.

draft-ietf-tls-esni.md
chris-wood: That works for me!

draft-ietf-tls-esni.md
chris-wood: It's defined where it's used.

draft-ietf-tls-esni.md
chris-wood: I'll add that, instead of replace the censor example.

draft-ietf-tls-esni.md
chris-wood: Wouldn't hurt!

draft-ietf-tls-esni.md
chris-wood: Nice! I'll do that.

draft-ietf-tls-esni.md
chris-wood: We also have not done the analysis to know if it is possible in that setting. So, for now, I'd prefer we just be up front about this.

draft-ietf-tls-esni.md
cjpatton: Yeah, better to put it where it's first defined. I guess my issue is how it's defined. It's not that they have the same configuration ... what matters is simply that the ClientHello protection service is provided on their behalf. Maybe replace "wherein all elements have a consistent configuration..." with "of servers for which the service is provided".

draft-ietf-tls-esni.md
cjpatton: Ack.

draft-ietf-tls-esni.md
chris-wood: Chatted offline -- will improve the definition in a separate issue!

draft-ietf-tls-esni.md
```diff
@@ -84,11 +84,14 @@ both.
...
 that private origins will co-locate with or hide behind a provider (reverse
-proxy, application server, etc.) that protects SNIs for all of the domains it
-hosts. As a result, SNI protection does not indicate that the client is
-attempting to reach a private origin, but only that it is going to a particular
-service provider, which the observer could already tell from the visible IP
-address.
+proxy, application server, etc.) that protects sensitive ClientHello parameters,
+including the SNI, for all of the domains it hosts. These co-located servers
+form an anonymity set wherein all elements have a consistent configuration,
+e.g., set of supported application protocols, ciphersuites, TLS versions, and
```
ekr, line 90:
```suggestion
e.g., the set of supported application protocols, ciphersuites, TLS versions, and
```

draft-ietf-tls-esni.md
```diff
@@ -194,6 +197,11 @@ Upon receiving the server's response, the client determines whether ECH was
 accepted or rejected and proceeds with the handshake accordingly. (See
 {{client-behavior}} for details.)
 
+Informally, a primary goal of ECH is ensuring that connections to servers in the
+same anonymity set are indistinguishable from one another for passive and active
+attackers without affecting any existing security properties of TLS 1.3. See
+{{goals}} for more details about the ECH security and privacy goals.
```
ekr, line 203:
```suggestion
same anonymity set are indistinguishable from one another without affecting any existing security properties of TLS 1.3. See
{{goals}} for more details about the ECH security and privacy goals.
```

ekr: No need to say what kind of attacker if it's both.

draft-ietf-tls-esni.md
```diff
@@ -819,13 +827,21 @@ Clients SHOULD NOT attempt to repair the connection in this case.
...
+## Security and Privacy Goals {#goals}
+
+ECH considers two types of attackers: passive and active. Passive attackers can
+read packets from the network. They cannot perform any sort of active behavior
+such as probing servers or querying DNS. A middlebox that filters based on
+plaintext packet contents is one example of a passive attacker. In contrast,
+active attackers can write packets into the network for malicious purposes, such
+as interfering with existing connections, probing servers, and querying DNS. In
+short, an active attacker corresponds to the conventional threat model for
+TLS 1.3 {{RFC8446}}. A censor is one example of an active attacker.
```
ekr, line 839:
```suggestion
TLS 1.3 {{RFC8446}}.
```
I would just stay out of censorship. I think some censors are maybe passive?

draft-ietf-tls-esni.md
```diff
@@ -842,6 +858,29 @@ example, if a client-facing server services many backend origin hosts, only one
...
+Beyond these primary security and privacy goals, ECH also aims to hide, to some
+extent, (a) whether or not a specific server supports ECH and (b) whether or
+not ECH was accepted for a particular connection. ECH aims to achieve both
+properties, assuming the attacker is passive and does not know the set of ECH
+configurations offered by the client-facing server. It does not achieve these
+properties for active attackers. More specifically:
+
+- Passive attackers with a known ECH configuration can distinguish between a
+connection that negotiates ECH with that configuration and one which does not,
+i.e., because the latter used a GREASE "encrypted_client_hello" extension (as
```
ekr, line 870:
```suggestion
because the latter used a GREASE "encrypted_client_hello" extension (as
```

I don't think i.e., is adding much here.

draft-ietf-tls-esni.md
```diff
@@ -842,6 +858,29 @@ example, if a client-facing server services many backend origin hosts, only one
...
+specified in {{grease-extensions}}) or a different ECH configuration.
+- Passive attackers without the ECH configuration cannot distinguish between a
+connection that negotiates ECH and one which uses a GREASE
+"encrypted_client_hello" extension.
```
ekr, line 874: Well, sort of. If they have connection memory they can see that everyone is using different config hashes.

Actually, I wonder if we should salt the hash.

draft-ietf-tls-esni.md
```diff
@@ -842,6 +858,29 @@ example, if a client-facing server services many backend origin hosts, only one
...
+- Active attackers can distinguish between a connection that negotiates ECH and
+one which uses a GREASE "encrypted_client_hello" extension.
+
+As a result, ECH is not suitable for bypassing active adversaries such as
+censors.
```
ekr, line 879: I would also not say "censor" here.

ekr (changes requested):

draft-ietf-tls-esni.md
chris-wood: We can, yeah, but let's do that in a separate PR. (I'm assuming "without the ECH configuration" means it doesn't keep state.)

ekr: OK

PR #306, opened 2020-09-22
Add MTI cipher suite based on x25519, HKDF-SHA256, and AES-128-GCM.

### LINKED ISSUE #251: Common supported KEM
It might make sense to have a default KEM that all implementations would need to support for each ECH version.

This would prevent clients to have to support all KEMs at the risk of falling back to plaintext because they don't support the server-supported KEM.

chris-wood:
Closes #251.

cc @cjpatton @ekr @fredericjacobs


cjpatton:
Looks good, presuming the HPKE draft is stable and the section numbers won't change.

PR #312, opened 2020-09-28
Match legacy_session_id values.

### LINKED ISSUE #310: legacy_session_id should match between inner and outer
If the client uses TLS 1.3's [middlebox compatibility mode](https://tools.ietf.org/html/rfc8446#appendix-D.4), there are a couple subtle points:

* Even though the middlebox will never see the inner ClientHello, we still need a non-empty legacy_session_id, because some servers use that value to condition the other bits.

* More importantly, the inner legacy_session_id needs to match the outer legacy_session_id! Otherwise the server will echo back the wrong one and trip up all the buggy middleboxes again.

Alternatively, we could say the server uses the outer ClientHello to negotiate middlebox compatibility mode. That would reduce bytes on the wire, but it'd upset all the split mode arrangements.

davidben:
Fixes #310.


chris-wood (approved):

chris-wood:
Thanks!

Issue #263, opened 2020-08-13
"outer_extensions": What's the purpose of the hash?


cjpatton:
Field `OuterExtensions.hash` is computed as "digest of the complete ClientHelloInner" (including presumably, the "outer_extensions" extension). I don't understand the analytical value of this hash. Can anyone explain its reasoning?

ekr:
The hash is intended to prevent the attacker from influencing the reconstructed CHInner by manipulating CHOuter. Consider the contrived example of an extension which is separately protected but only works for SNI-A and not SNI-B. E.g., it's constructed by doing E(K_sni, SNI). If that extension were included in CHOuter and then imported into CHInner via this mechanism, then an attacker could substitute it and use a reaction attack to determine whether the CH referred to A or B.

However, with the hash (which may be not well-specified), this attack gets caught by the client-facing server and the handshake is rejected without leaking information.

cjpatton:
My initial understanding of this mechanism was incorrect. What happens is that the hash is computed over the ClientHelloInner that is consumed by the backend server. This is intended to ensure that if ECH accepts, then the backend server consumes the CH intended by the client. (PR https://github.com/tlswg/draft-ietf-tls-esni/pull/293 makes this more clear, I think.)

I wonder if this mechanism is redundant. Agreement on the CH between the client and backend server is provided by the binding of the transcript hash to the handshake traffic secret, so any attempt to manipulate the ClientHelloInner by fiddling with the ClientHelloOuter should be detected. On the other hand, the binding of the outer extensions to ClientHelloInner ensures agreement on the CH *prior to the backend server consuming it*. This is a strictly stronger property, which may or may not be useful.

> Consider the contrived example of an extension which is separately protected but only works for SNI-A and not SNI-B.

I agree that the attack works, but the example seems *really* contrived :) I'd be interested if there are more natural extensions for which this extra layer of security would be useful.

cjpatton:
From the interim meeting, the consensus seems to be that we should err on the conservative side.

cjpatton:
Let's close this issue and leave the mechanism as is.

cjpatton:
Hrm, before closing I think it might be worth noting, in the spec, the stronger security property this provides.

bemasc:
[...]

PR #317, opened 2020-09-30
Simplify session resumption logic.

### LINKED ISSUE #285: Possible to simplify interaction with session resumption?
The current draft contains various rules around ECH and resumption.

> Per {{client-behavior}} and {{server-behavior}}, implementations are required to track, alongside each PSK established by a previous connection, whether the connection negotiated this extension with the "ech_accept" response type. If so, this is referred to as an "ECH PSK". Otherwise, it is a "non-ECH PSK". This may be implemented by adding a new field to client and server session states.

> When offering an encrypted ClientHello, the client MUST NOT offer to resume any non-ECH PSKs.

> If the server resumed a session or negotiated a session that did not use a certificate for authentication, the client MUST abort the connection with an "illegal_parameter" alert. This case is invalid because {{send-ech}} requires the client to only offer ECH-established sessions, and {{server-behavior}} requires the server to decline ECH-established sessions if it did not accept ECH.

> The server MUST ignore all PSK identities in the ClientHello which correspond to ECH PSKs. ECH PSKs offered by the client are associated with the ECH name. The server was unable to decrypt then ECH name, so it should not resume them when using the plaintext SNI name. This restriction allows a client to reject resumptions in {{auth-public-name}}.

> If the server sends a NewSessionTicket message, the corresponding ECH PSK MUST be ignored by all other servers in the deployment when not negotiating ECH, including servers which do not implement this specification.

These rules date to before we started encrypting the entire ClientHello and were intended to avoid mixing up session resumptions between the public and private name, since the same pre_shared_key extension was used for both, and amount to saying implementations must add a boolean to their session state.

Now that the public and private names correspond to different ClientHellos altogether, the client can (and should) only send inner name tickets in the inner ClientHello and outer name tickets in the outer ClientHello. (#270) I suspect that means we no longer need further ECH special cases here.

chris-wood:
Closes #285.

We need to decide if we want to allow a GREASE+ticket distinguisher, and if not, how we GREASE the "pre_shared_key" extension in a meaningful way. The latter seems like somewhat of a cat and mouse game, so I'm not really enthusiastic about trying to specify that behavior, but I'd like to hear what others think.

cc @davidben, @cjpatton, @martinthomson

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -447,8 +447,7 @@ The client then constructs the ClientHelloInner message just as it does a
 standard ClientHello, with the exception of the following rules:
 
 1. It MUST NOT offer to negotiate TLS 1.2 or below.
-1. It MUST NOT offer to resume any non-ECH PSK or any session for TLS 1.2 and
-   below.
+1. It MUST NOT offer to resume any session for TLS 1.2 and below.
 1. It MAY offer any other extension in the ClientHelloOuter except those that
    have been incorporated into the ClientHelloInner as described in
    {{outer-extensions}}.
@@ -756,7 +755,7 @@ the following section, or forwards the TLS connection to the backend server (if
 in Split Mode). In the latter case, it does not make any changes to the TLS
 messages, but just blindly forwards them.
 
-If the server sends a NewSessionTicket message, the corresponding ECH PSK MUST
+If the server sends a NewSessionTicket message, the corresponding PSK MUST
 be ignored by all other servers in the deployment when not negotiating ECH,
 including servers which do not implement this specification.
 
@@ -1227,7 +1226,7 @@ ECH mitigates against this attack by (1) prohibiting the "pre_shared_key"
 extension in ClientHelloOuter and (2) requiring servers ignore this extension
 when processing ClientHelloOuter. Thus, if the server accepts the inner
 ClientHello, it only validates binders in the inner ClientHello. This means
-that ECH PSKs are used within the HPKE encryption envelope.
+that PSKs are used within the HPKE encryption envelope.
 
 # IANA Considerations
 
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -375,15 +375,6 @@ This document also defines the "ech_required" alert, which clients MUST send
 when it offered an "encrypted_client_hello" extension that was not accepted by
 the server. (See {{alerts}}.)
 
-[[OPEN ISSUE: The following text is a remnant of the section on "ech_nonce". It
-specifies client/server behavior that is unrelated to "ech_nonce".]]
-
-Per {{client-behavior}} and {{server-behavior}}, implementations are required to
-track, alongside each PSK established by a previous connection, whether ECH was
-accepted for this connection. If so, this is referred to as an "ECH PSK".
-Otherwise, it is a "non-ECH PSK". This may be implemented by adding a new field
-to client and server session states.
-
 ## Incorporating Outer Extensions {#outer-extensions}
 
 Some TLS 1.3 extensions can be quite large and having them both in the inner and
@@ -449,8 +440,8 @@ standard ClientHello, with the exception of the following rules:
    constructed as described below.
 1. The value of `ECHConfig.public_name` MUST be placed in the "server_name"
    extension.
-1. It MUST NOT include the "pre_shared_key" extension. [[OPEN ISSUE: The text
-   needs to be updated to reflect this rule.]]
+1. It MUST NOT include the "pre_shared_key" extension. (See 
+   {{flow-resumption-oracle}}.)
 
 The client then constructs the ClientHelloInner message just as it does a
 standard ClientHello, with the exception of the following rules:
@@ -669,7 +660,7 @@ If the client attempts to connect to a server and does not have an ECHConfig
 structure available for the server, it SHOULD send a GREASE {{?RFC8701}}
 "encrypted_client_hello" extension as follows:
 
-- Set the "suite" field  to a supported ECHCipherSuite. The selection SHOULD
+- Set the "suite" field to a supported ECHCipherSuite. The selection SHOULD
   vary to exercise all supported configurations, but MAY be held constant for
   successive connections to the same server in the same session.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -690,6 +681,10 @@ If the server sends an "encrypted_client_hello" extension, the client MUST check
 the extension syntactically and abort the connection with a "decode_error" alert
 if it is invalid.
 
+[[OPEN ISSUE: if the client sends a GREASE "encrypted_client_hello" extension, 
+should it also send a GREASE "pre_shared_key" extension? If not, GREASE+ticket
+is a trivial distinguisher.]]
```
cjpatton, line 686: I suggest leaving this OPEN ISSUE open and address it post -08. This is one way that real ECH sticks out from covertext, but there are likely others. What is needed is a big picture "don't stick out" review of ECH.

draft-ietf-tls-esni.md
```diff
@@ -1229,11 +1223,11 @@ can be used to test the encrypted SNI value of specific ClientHello messages.
 ~~~
 {: #tls-resumption-psk title="Message flow for resumption and PSK"}
 
-ECH mitigates against this attack by requiring servers not mix-and-match
-information from the inner and outer ClientHello. For example, if the server
-accepts the inner ClientHello, it does not validate binders in the outer
-ClientHello. This means that ECH PSKs are used within the HPKE encryption
-envelope.
+ECH mitigates against this attack by (1) prohibiting the "pre_shared_key" 
+extension in outer ClientHello and (2) requiring servers ignore this extension
```
cjpatton, line 1227: nit: "outer ClientHello" -> "the ClientHelloOuter". Here and below.

draft-ietf-tls-esni.md
```diff
@@ -1229,11 +1223,11 @@ can be used to test the encrypted SNI value of specific ClientHello messages.
...
-ECH mitigates against this attack by requiring servers not mix-and-match
-information from the inner and outer ClientHello. For example, if the server
-accepts the inner ClientHello, it does not validate binders in the outer
-ClientHello. This means that ECH PSKs are used within the HPKE encryption
-envelope.
+ECH mitigates against this attack by (1) prohibiting the "pre_shared_key" 
+extension in outer ClientHello and (2) requiring servers ignore this extension 
+when processing the outer ClientHello. Thus, if the server accepts the inner 
+ClientHello, it only validates binders in the inner ClientHello. This means 
+that ECH PSKs are used within the HPKE encryption envelope.
```
cjpatton, line 1230: Is "ECH PSK" an obsolete term?

chris-wood: We use it when noting tickets/PSKs associated with an ECH accepted connection.

draft-ietf-tls-esni.md
chris-wood: I made this change more broadly.

draft-ietf-tls-esni.md
chris-wood: I'm fine with that, I think, though we may want to reconsider it before the GREASE experiments commence.

draft-ietf-tls-esni.md
```diff
@@ -449,8 +440,8 @@ standard ClientHello, with the exception of the following rules:
    constructed as described below.
 1. The value of `ECHConfig.public_name` MUST be placed in the "server_name"
    extension.
-1. It MUST NOT include the "pre_shared_key" extension. [[OPEN ISSUE: The text
-   needs to be updated to reflect this rule.]]
+1. It MUST NOT include the "pre_shared_key" extension. (See
+   {{flow-resumption-oracle}}.)
 
 The client then constructs the ClientHelloInner message just as it does a
 standard ClientHello, with the exception of the following rules:
```
davidben, line 447: (Noticed this while looking at the text for something else: the list below also has a reference to "non-ECK PSK" that should be dropped here.)

chris-wood: I explicitly kept that, but I don't really feel strongly about keeping it. So I just removed it.

davidben: Hehe, I mean this PR removes the definition of "ECH PSK" and "non-ECH PSK" so we probably shouldn't reference them. :-) I was hoping we'd no longer have to associate this extra boolean with each session.

cjpatton:
This looks good. I still feel that there are a lot of ways that ECH usage might stick out, even to passive observers. "pre_shared_key" is one of many. If anything I would change the "OPEN ISSUE" to encompass top-to-bottom "don't stick out" review.

draft-ietf-tls-esni.md
```diff
@@ -1235,11 +1228,11 @@ can be used to test the encrypted SNI value of specific ClientHello messages.
...
-ECH mitigates against this attack by requiring servers not mix-and-match
-information from the inner and outer ClientHello. For example, if the server
-accepts the inner ClientHello, it does not validate binders in the outer
-ClientHello. This means that ECH PSKs are used within the HPKE encryption
-envelope.
+ECH mitigates against this attack by (1) prohibiting the "pre_shared_key"
+extension in ClientHelloOuter and (2) requiring servers ignore this extension
+when processing ClientHelloOuter. Thus, if the server accepts the inner
+ClientHello, it only validates binders in the inner ClientHello. This means
+that PSKs are used within the HPKE encryption envelope.
```
davidben, line 1235: I think this section might be anchoring too much on PSKs. I don't think it's actually a general attack mitigated by a more general fix.

Back when it was ESNI, the final ClientHello was malleable. An attacker could change some parameters while carrying over the secret server name. If the server interprets the influenced parameters differently based on the secret server name, the attacker can use that as an oracle. The most natural vector is the PSK, because resumptions may be SNI-partitioned. But if you have, say, different ALPN preferences by hostname, that would also equally be a vector.

We've fixed this by:
1. Encrypting the entire ClientHello gives us a convenient way to bundle all parameters together, so it's natural that servers don't look at the outer CH. (The section talks about a server that mixes the two CHs up, which is totally bizarre.)
2. Everything that goes into the final ClientHello is directly or indirectly authenticated by the AEAD. (OuterExtensions.inner_digest or #323.)

I'll take a stab at a replacement for this section...

davidben (approved):
I think this looks good modulo the PSK oracle section seeming kinda odd to begin with.

draft-ietf-tls-esni.md
chris-wood: > (The section talks about a server that mixes the two CHs up, which is totally bizarre.)

IMO, this is why we kept the section. I agree that it's bizarre, but not much of a stretch from the original attack vector. I'd like to keep some mention of this mix-and-match problem for historical reasons, though I'm definitely happy to look at simpler text!

davidben: Dunno if it's simpler, but it's hopefully more general. I think the description of the mitigation should touch on OuterExtensions.inner_digest's role in authenticating the inner ClientHello, which would have answered #263. That is, if we allowed you to compress pre_shared_keys for whatever reason, this attack would still be impossible because of that hash.

(Not that you'd want to compress pre_shared_key since a compressed is public and leaking the actual ticket gives you a different kind of oracle. That falls under "things that depend on secret things are also secret".)

davidben: https://github.com/tlswg/draft-ietf-tls-esni/pull/327

Issue #309, opened 2020-09-28
Ensuring 1.3 is used when ECH is accepted


cjpatton:
Usage of 1.3 is enforced by the following rules.

* Client (Section 6.1):
    > It MUST offer to negotiate TLS 1.3 or above [in ClientHelloOuter]. [...] It MUST NOT offer to negotiate TLS 1.2 or below [in ClientHelloInner].

* Client-facing server (7.1):
    > Upon receiving an "encrypted_client_hello" extension, the client-facing server MUST check that it is able to negotiate TLS 1.3 or greater. If not, it MUST abort the connection with a "handshake_failure" alert.

* The backend server does not enforce 1.3 usage. (At least, I couldn't find an explicit rule in the text.)

Requiring that 1.3 is the only version offered in the ClientHelloInner is sufficient to ensure that 1.3 is negotiated by the client and backend server in case of acceptance. However, my sense is that the client would be easier to implement if this rule were softened to:

> It MUST offer to negotiate TLS 1.3 in the ClientHelloInner.

We could then add the following behavior to the backend server (Section 7.2):

> If the ClientHelloInner indicates ECH was accepted [i.e., there is an empty "encrypted_client_hello" extension in the ClientHello], but the client does not offer support for TLS 1.3 or above, then the backend server MUST abort the handshake with a "[some error]" alert.

What do y'all think?

cjpatton:
I'm happy to put together a PR if folks are happy with this change.

davidben:
I don't feel strongly either way. I actually hadn't realized the client was already forbidden from offering TLS 1.2. The server needs to check anyway, and it's much more natural to check the negotiated version, so I don't think it helps the server in either direction. (@dmcardle since we were talking about this recently.)

There's no real reason for the client to offer TLS 1.2 in the inner ClientHello, and it could send a potentially smaller ClientHello (fewer cipher suites, omit renegotiation_info, extended_master_secret, shorter supported_versions). One footnote is a shorter supported_versions is still more bytes than a compressed supported_versions, but I think the savings elsewhere exceed that. (At least for now. Maybe 10 years from now, there will be very few TLS 1.2 ciphers and very many post-TLS-1.2 versions before TLS 1.2 is gone??)

I guess the other factor is it may also be more convenient for the client not to have juggle two different min_version values around in its ClientHello construction logic. But I feel like the client should decide based on how many bytes it takes on the wire.

cjpatton:
[...]

PR #320, opened 2020-10-01
Fix ECH and version negotiation order.

davidben:
The version check on the outer ClientHello doesn't make much sense
since, when ECH is accepted, the version is based on the inner
ClientHello anyway. See
https://github.com/tlswg/draft-ietf-tls-esni/issues/309#issuecomment-702233855.

I've left a TODO since it seems weird to have a client requirement like
that be so load-bearing without a corresponding server check. Not sure
where we should go there.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -699,8 +699,10 @@ MAY offer to resume sessions established without ECH.
 ## Client-Facing Server
 
 Upon receiving an "encrypted_client_hello" extension, the client-facing server
-MUST check that it is able to negotiate TLS 1.3 or greater. If not, it MUST
-abort the connection with a "handshake_failure" alert.
+determines if it can decrypt it, prior to negotiating any other TLS parameters.
+Note that successfully decrypting the extension will result in a new
+ClientHello to process, so even the client's TLS version preferences may have
+changed.
 
 The ClientECH value is said to match a known ECHConfig if there exists
 an ECHConfig that can be used to successfully decrypt
@@ -790,6 +792,13 @@ MUST confirm ECH acceptance by setting ServerHello.random[24:32] to
 where HKDF-Expand-Label and HKDF-Extract are as defined in {{RFC8446}}. The
 value of ServerHello.random[0:24] is generated as usual by invoking a secure
 random number generator (see {{RFC8446}}, Section 4.1.2).
+
+\[\[TODO: If the ClientHello offers TLS 1.2 or below, it's possible the backend
+server will negotiate it and leak the server name via an unencrypted server
+certificate. This shouldn't happen because clients are forbidden from offering
+earlier versions, but perhaps either the backend server should check (easy in
+Shared Mode, hard in Split Mode), or maybe the client-facing server should
+check for the presence of old TLS versions in supported_versions.\]\]
 
 # Compatibility Issues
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -791,6 +793,13 @@ where HKDF-Expand-Label and HKDF-Extract are as defined in {{RFC8446}}. The
 value of ServerHello.random[0:24] is generated as usual by invoking a secure
 random number generator (see {{RFC8446}}, Section 4.1.2).
 
+\[\[TODO: If the ClientHello offers TLS 1.2 or below, it's possible the backend
```
cjpatton, line 796: Which ClientHello?

draft-ietf-tls-esni.md
```diff
@@ -699,8 +699,10 @@ MAY offer to resume sessions established without ECH.
 ## Client-Facing Server
 
 Upon receiving an "encrypted_client_hello" extension, the client-facing server
-MUST check that it is able to negotiate TLS 1.3 or greater. If not, it MUST
-abort the connection with a "handshake_failure" alert.
+determines if it can decrypt it, prior to negotiating any other TLS parameters.
```
cjpatton, line 702: s/can decrypt it/can accept ECH/, since "accept ECH" is a well-defined term.

draft-ietf-tls-esni.md
```diff
@@ -791,6 +793,13 @@ where HKDF-Expand-Label and HKDF-Extract are as defined in {{RFC8446}}. The
 value of ServerHello.random[0:24] is generated as usual by invoking a secure
 random number generator (see {{RFC8446}}, Section 4.1.2).
 
+\[\[TODO: If the ClientHello offers TLS 1.2 or below, it's possible the backend
+server will negotiate it and leak the server name via an unencrypted server
+certificate. This shouldn't happen because clients are forbidden from offering
+earlier versions, but perhaps either the backend server should check (easy in
+Shared Mode, hard in Split Mode), or maybe the client-facing server should
```
cjpatton, line 800: Why is this hard in Split Mode? We could mandate that if the backend server observes the empty ECH extension (sent in the ClientHelloInner in order to signal the backend server to confirm acceptance), it aborts unless it can negotiate 1.3 or higher.

cjpatton:
I like this change, but it might be worth resolving the TODO.

draft-ietf-tls-esni.md
chris-wood, line 702:
```suggestion
determines if it will accept ECH, prior to negotiating any other TLS parameters.
```

draft-ietf-tls-esni.md
chris-wood: Presumably the inner, given below.

chris-wood (commented):
> I've left a TODO since it seems weird to have a client requirement like that be so load-bearing without a corresponding server check. Not sure where we should go there.

I'm fine with omitting a check on the server. I don't see much difference between a client implementation choosing to ignore the MUST in generating its ClientHelloInner supported_versions list, and a client implementation not supporting ECH at all. Plus, I could see future extensions or variants of ECH where we allow TLS 1.2 for the backend, and the server's handshake traffic is tunneled between the client and client-facing-server, and a check might make that difficult to do.

draft-ietf-tls-esni.md
davidben: Yeah, the backend server doesn't see the outer one.

draft-ietf-tls-esni.md
davidben: (Resolved by committing @chris-wood's suggestion.)

davidben:
> I'm fine with omitting a check on the server. 

Removed in favor of some clarifying text on the client side.

chris-wood (approved):

Issue #295, opened 2020-09-02
Greasing to supporting servers


wbl:
@cjpatton and I were sitting down and thinking and we realized the following scenario can happen:

1. example.com is on hosting.com, which supports ECH
1. example.com has an HTTPS record, an A record, AAAA record pointing at the IP for hosting.com
1. A user of a browsers latest release that supports ECH and greases the extension is using a resolver that doesn't resolve HTTPS records

This results in an ECH request being made with SNI=example.com, and a rejection of the ECH. So now the question is what the client should do and what should servers expect?
* Try again, with the outer SNI the one matching the one in the new config
* Use the connection authoritatively for example.com since the connection has already revealed the issue.

davidben:
The intention was the latter, but the text could probably be clearer. I don't think there's much point in trying again since the name has already been revealed. See https://github.com/tlswg/draft-ietf-tls-esni/issues/274#issuecomment-680182501

chris-wood:
@wbl is this still an issue now that #287 is merged?

wbl:
I think the client behavior remains equally unspecified. In particular the lengthy argument @cjpatton and I had about server enforcement of the client behavior I think isn't changed.

cjpatton:
Care to file a PR, presuming the latter behavior is intended?

cjpatton:
A proposed resolution is here: #319.

cjpatton:
@wbl, a proposed fix has been merged. See #321.

wbl:
I think this text works.

PR #321, opened 2020-10-01
Clarify client handling of retry keys with GREASE.

davidben:
The intent was that the server always sends the retry keys (it cannot
distinguish stale keys from GREASE when the client is actually trying to
connect to the public name), while the client just ignores them, but I
did a poor job of describing this, so add a key sentence.

This addresses #295.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -688,7 +688,8 @@ structure available for the server, it SHOULD send a GREASE {{?RFC8701}}
 
 If the server sends an "encrypted_client_hello" extension, the client MUST check
 the extension syntactically and abort the connection with a "decode_error" alert
-if it is invalid.
+if it is invalid. It otherwise ignores the extension and MUST NOT use the retry
+keys.
 
 Offering a GREASE extension is not considered offering an encrypted ClientHello
 for purposes of requirements in {{client-behavior}}. In particular, the client


cjpatton:
To recap the discussion on #319, your point is that the client-facing server needn't attempt to distinguish GREASE from non-GREASE based on outer SNI, because in either case the client's response is to ignore the retry configuration and continue as usual.

I think this is weird because if rejection was the intent, then the client supposed to abort with "ech_required". Does the client know, unambiguously, whether the ECH was rejected by the server? It's not clear to me that this is the case.

davidben:
I mean, yeah, that's how this was designed and is what we merged to the spec. :-)

This formulation, in particular, does not break if the client tries to connect to the public name (it's a perfectly valid URL) without knowing the ECH config (maybe it the HTTPS record didn't get through Do53) but still sends GREASE (because the intent is that every client can GREASE). It seems to me #319 breaks this. Do you have an alternate formulation which keeps this working?

Connecting to the public name is especially important if the public name is a non-throwaway domain. The only requirement on the public name is that it's some name which knows about and serves the client-facing server's config. (I.e. this particular backend server is the same entity as the client-facing server.) #319 doesn't relax this requirement, since the public name continues to be the entity trusted to replace, and possibly disable, ECH on config mismatch.

chris-wood:
> I think this is weird because if rejection was the intent, then the client is supposed to abort with "ech_required". So the server only knows that non-GREASE was intended once it receives this signal.

I don't see this as an issue. If the client sends GREASE, it doesn't expect to use ECH, so it won't abort. The text here only requires that the client check the extension if present, and ignore it if absent.

cjpatton:
I don't have an alternative formulation. I was evnisioning the public name only being verified on the ECH rejection path, or when neither real nor dummy ECH was offered to the client-facing server.

I'm fine with rejection not being known until "ech_required". I support this PR as the resolution to #295.

Issue #323, opened 2020-10-05
Re-re-visit outer ClientHello binding (saves 65 bytes)


davidben:
We previously decided not to bind the outer ClientHello to the ECH decryption, because doing so is rather a hassle. We'd need something similar to the TLS 1.3 PSK binder mess, with the added complexity of ordering relative to the PSK binders themselves. Additionally, while the binding helps with some of the don't stick out properties, there were other attacks on distinguishability anyway. So, instead, we mitigated the inner ClientHello malleability with OuterExtensions.inner_digest and left the outer ClientHello unbound.

However, we've since decided not to allow PSKs in the outer ClientHello. Perhaps we should revisit this. Suppose the AD parameter to the ECH decryption were the entire outer ClientHello, minus the ECH extension. I believe this means we no longer need OuterExtensions.inner_digest. That would give an immediate 33-byte savings. It would then unlock a further 32 byte savings by amending #312 to say:

* Client always encodes an empty legacy_session_id in ECH plaintext.
* Client-facing server copies outer legacy_session_id to inner ClientHello as part of decompression.

(The reason we can't do this easily right now is that, if the inner ClientHello lacks outer_extension, nothing's authenticating legacy_session_id. Since the field in TLS 1.3 is all but ignored, it's "obvious" that the malleability doesn't matter, but it's nicer to not need to think about such things.)

From looking at Chrome's ClientHellos and working through what we'd send (drop 1.2-only extensions, compress most), it looks like the inner ClientHello will be maybe 130-140 bytes, plus a padded server_name. The 65 bytes above would almost halve the overhead.

In some sense what's going on here is our 16-byte tag on the AEAD in the ECH ciphertext is going to waste. We didn't authenticate everything that goes into reconstructing the inner ClientHello (outer_extensions), so instead we authenticate another hash (OuterExtensions.inner_digest) and check that afterwards. (To that end, if we do this, we probably need to say the server MUST reject encrypted_client_hello showing up in outer_extensions, that way all inputs to decompression are authenticated. Equivalently, decompression is done relative to the ECH-extension-less AD input rather than the original outer ClientHello.)

cjpatton:
I like this idea. How do you think the AD ought to be structured? The ClientHelloOuter, but with an an empty "encrypted_client_hello" extension?

davidben:
[...]

PR #330, opened 2020-10-09
Consistently construct ClientHelloInner before ClientHelloOuter.

### LINKED ISSUE #329: ClientHello construction order is inconsistent
I noticed this while drafting a PR for #323. We're a little inconsistent about whether the ClientHelloInner or ClientHelloOuter is constructed first.

Section 4 says inner before outer:
> When a client wants to establish a TLS session with the backend server, it constructs its ClientHello as usual (we will refer to this as the ClientHelloInner message) and then encrypts this message using the public key of the ECH configuration. It then constructs a new ClientHello (ClientHelloOuter)

Section 5.1 implies inner before outer:
> When sending ClientHello, the client first computes ClientHelloInner, including any PSK binders, and then MAY substitute extensions which it knows will be duplicated in ClientHelloOuter.

But section 6.1 says outer before inner:
> Next, the client constructs the ClientHelloOuter message just as it does a standard ClientHello, with the exception of the following rules:
> ...
> The client then constructs the ClientHelloInner message just as it does a standard ClientHello, with the exception of the following rules:

I think inner before outer makes more sense. You certainly can't finish the outer one before you've computed the inner one. (Of course, an implementation can pick whatever order is most convenient, but I think the spec should be consistent with itself.)

davidben:
Closes #329.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -432,36 +432,35 @@ identified by `ECHConfig.cipher_suites`. Once a suitable configuration is found,
 the client selects the cipher suite it will use for encryption. It MUST NOT
 choose a cipher suite not advertised by the configuration.
 
-Next, the client constructs the ClientHelloOuter message just as it does a
-standard ClientHello, with the exception of the following rules:
-
-1. It MUST offer to negotiate TLS 1.3 or above.
-1. It MUST include an "encrypted_client_hello" extension with a payload
-   constructed as described below.
-1. The value of `ECHConfig.public_name` MUST be placed in the "server_name"
-   extension.
-1. It MUST NOT include the "pre_shared_key" extension. (See
-   {{flow-clienthello-malleability}}.)
-
-The client then constructs the ClientHelloInner message just as it does a
+Next, the client constructs the ClientHelloInner message just as it does a
 standard ClientHello, with the exception of the following rules:
 
 1. It MUST NOT offer to negotiate TLS 1.2 or below. Note this is necessary to
    ensure the backend server does not negotiate a TLS version that is
    incompatible with ECH.
 1. It MUST NOT offer to resume any session for TLS 1.2 and below.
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -432,36 +432,35 @@ identified by `ECHConfig.cipher_suites`. Once a suitable configuration is found,
...
-   ClientHelloOuter.random. Instead, It MUST generate a fresh
-   ClientHelloInner.random using a secure random number generator. (See
-   {{flow-client-reaction}}.)
 1. It SHOULD contain TLS padding {{!RFC7685}} as described in {{padding}}.
+
+The client then constructs the ClientHelloOuter message just as it does a
+standard ClientHello, with the exception of the following rules:
+
+1. It MUST offer to negotiate TLS 1.3 or above.
+1. Any extensions compressed as described in {{outer-extensions}} must match
```
cjpatton, line 448: This rule could be made more clear. Suggestion:
> Any extension that appears in the payload of the "outer_extensions" extension of the ClientHelloInner MUST appear in the ClientHelloOuter.

davidben: I didn't want to talk about outer_extensions specifically appearing in ClientHelloInner because of #331. The ClientHelloInner doesn't really have outer_extensions, only the payload of the ECH extension. We haven't settled yet on text there, or a processing order of actually encrypting things (I'll need to fiddle with that as part of #323), so for now I figure citing the actual process was the least ambiguous.

The distinction is important because of which is actually used in the handshake. (Transcript and whatnot.)

cjpatton: Then I would just leave an "OPEN ISSUE" so we know to fix this later.

davidben: Done.

cjpatton (approved):

chris-wood (approved):

Issue #262, opened 2020-08-12
"outer_extensions" mechanism is lossy compression


cjpatton:
Say we want the outer CH to have extensions a and b. Let's write this as
```
  ClientHelloOuter = [a, b]
```
Say the inner CH should have extensions a, b, c, and d. But b is really big, so we want a shallow copy it. Then we write the compressed inner CH as
```
   ClientHelloInner = [a, c, d, outer_extensions = [ shallow_copy_of_b ]]
```
The decompressed ClientHelloInner should be `[a, b, c, d]`, but using `outer_extensions` destroys the intended order.

davidben:
Extension order in TLS isn't meaningful, so I think that's fine, provided the code producing the inner ClientHello and the code encoding it get to coordinate. (I.e. a client that says "I don't care about secrecy of `key_share` and will send the same one on both sides" can probably live with `key_share` being at the end.)

There have been cases where extension order has interop impacts due to server bugs. (E.g. the last extension needs to be non-empty to avoid a bug in some old servers.) But those old servers wouldn't see inner ClientHellos anyway.

chris-wood:
I think maybe a concern is that the client and server wind up using extension lists with different orders, for whatever reasons, thereby yielding different transcripts. Wouldn't that be an issue?

davidben:
Right, hence "provided the code producing the inner ClientHello and the code encoding it get to coordinate". If compression is done with inner + outer ClientHello being produced by the same thing, the client can simply take this into account and always put referenced extensions at the end. If it's a proxying picture, where the client does not know it is doing ECH (does that work? I suspect the ech_nonce would be an issue), that may not work.

chris-wood:
Right. I don't expect the latter (proxy case) to be common, though perhaps modifying this to prevent misuse might be useful.

cjpatton:
This make me worry about interoperability. For a client and server to speak to eachother they need to adopt the same convention that is not specified anywhere.

davidben:
It is certainly necessary that the decompression function produces a deterministic ClientHello, down to what goes into the transcript. I think section 6.1 does that. That's a different criteria from requiring that every possible input ClientHello be efficiently compressible.

cjpatton:
I don't compare about it being "efficiently" compressible, but I see your point about the transcript being the same. Just to clarify, my concern is in Section 6.1: 

> Then,
   the client removes and and replaces extensions from ClientHelloInner
[...]

PR #335, opened 2020-10-11
Disambiguate the encoded vs actual ClientHelloInner.

### LINKED ISSUE #331: Encoded vs actual ClientHelloInner is a little ambiguous
There are two ClientHelloInners, the actual one used in the handshake, and the encoded one with outer_extensions substituted in. Both are described as ClientHelloInner, but this makes it slightly ambiguous which goes into, say, the handshake transcript.

(It's late now, so I'll leave this as an issue without a PR. If we don't end up resolving this separately, I'll probably fix this as part of #323. If we decide to implicitly copy over legacy_session_id, the encryption payload becomes less and less defensibly an actual ClientHelloInner.)

davidben:
Closes #331. Along the way, I've made two additional tweaks:

- Renamed outer_extension to outer_extensions, now that it contains multiple extensions. One of the diagrams even already referred to "outer_extensions" because I didn't realize it was called "outer_extension".

- Clarified in compression process and ClientHelloInner construction rules that the to-be-compressed extensions must be consecutive. Hopefully that helps with #262.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -1274,7 +1274,7 @@ for ExtensionType (defined in {{!RFC8446}}):
 
 1. encrypted_client_hello(0xff08), with "TLS 1.3" column values being set to
    "CH, EE", and "Recommended" column being set to "Yes".
-3. outer_extensions(0xff04), with the "TLS 1.3" column values being set to "CH",
+3. outer_extensions(0xff04), with the "TLS 1.3" column values being set to "",
    and "Recommended" column being set to "Yes".
 
 ## Update of the TLS Alert Registry {#alerts}
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -417,6 +417,10 @@ in ClientHelloInner. The list of outer extensions,
 OuterExtensions.outer_extensions, includes those which were removed from
 EncodedClientHelloInner, in the order in which they were removed.
 
+Finally, EncodedClientHelloInner is serialized as a ClientHello structure,
+defined in Section 4.1.2 of {{RFC8446}}. Note this does not include the
+four-byte header included in the Handshake structure.
+
 The client-facing server computes ClientHelloInner by reversing this process.
 It scans EncodedClientHelloInner for an "outer_extensions" extension. If there
 is none, the ClientHelloInner is the EncodedClientHelloInner. Otherwise, it
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -381,11 +381,11 @@ Some TLS 1.3 extensions can be quite large and having them both in
 ClientHelloInner and ClientHelloOuter will lead to a very large overall size.
 One particularly pathological example is "key_share" with post-quantum
 algorithms. In order to reduce the impact of duplicated extensions, the client
-may use the "outer_extension" extension.
+may use the "outer_extensions" extension.
 
 ~~~
     enum {
-       outer_extension(0xff04), (65535)
+       outer_extensions(0xff04), (65535)
     } ExtensionType;
 
     struct {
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -1265,7 +1270,7 @@ for ExtensionType (defined in {{!RFC8446}}):
 
 1. encrypted_client_hello(0xff08), with "TLS 1.3" column values being set to
    "CH, EE", and "Recommended" column being set to "Yes".
-3. outer_extension(0xff04), with the "TLS 1.3" column values being set to "CH",
+3. outer_extensions(0xff04), with the "TLS 1.3" column values being set to "CH",
```
davidben, line 1273: Should the TLS 1.3 column be "" instead of "CH"? It's not *really* a ClientHello extension.

davidben:
> Clarified in compression process and ClientHelloInner construction rules that the to-be-compressed extensions must be consecutive. Hopefully that helps with #262.

@cjpatton FYI. I'm hoping this will help clarify why the ordering thing isn't actually a big deal.

davidben:
I've updated the PR to include text to disambiguate the serialization.

It's ambiguous whether the encoding includes the 4-byte (12-byte in DTLS) header from the Handshake structure or is just the body, from the ClientHello structure. I've picked the smaller one. This also saves having to answer, in DTLS, what happens when the inner sequence number is wrong.

draft-ietf-tls-esni.md
cjpatton: I think so.

draft-ietf-tls-esni.md
```diff
@@ -442,14 +454,16 @@ standard ClientHello, with the exception of the following rules:
    incompatible with ECH.
 1. It MUST NOT offer to resume any session for TLS 1.2 and below.
 1. It SHOULD contain TLS padding {{!RFC7685}} as described in {{padding}}.
+1. If it intends to compress any extensions (see {{outer-extensions}}), it
+   MUST order those extensions consecutively.
```
cjpatton, line 458: It should be possible to relax this for the existing "inner_digest" mechanism. However, I suspect this is will be essential once #336 lands.

cjpatton (approved):

draft-ietf-tls-esni.md
davidben: Hrm, I'm not sure I follow. I was thinking less a security requirement and more a correctness/syntax one. The outer_extensions mechanism, inner digest or AAD, can't express more than one consecutive chunk of slides extensions. That means that when you assemble the ClientHello, you need to adjust the order to fit your intended compression.

cjpatton: My point is that the map from ClientHelloInner to EncodedClientHelloInner need not be invertible. To compute "inner_digest", the client just needs to be able to compute the same map from EncodedClientHelloInner to ClientHelloInner that the server does. It's not necessary that the extensions be consecutive, although it would be much easier to implement this correctly if they were.

In any case, I'm happy with this rule.

davidben: I'm still a little confused. Just double-checking that the text is clear enough and we're on the same page... (If we're both happy with the text, but because we interpreted it differently, that's not very helpful!)

I am saying it *is* important that ClientHelloInner to EncodedClientHelloInner is invertible. This is true in the status quo, and it remains true after #336. Handshake correctness requires that the server reconstruct the *exact* same value, byte-for-byte. Even though the extension order isn't very meaningful semantically, it changes what goes in the transcript and binders, so both sides need to agree. So if the client first assembles:

`ClientHelloInner = (key_share, supported_versions, supported_groups, pre_shared_keys)`

then it compresses key_share + supported_groups as:

`EncodedClientHelloInner = (outer_extensions(key_share, supported_groups), supported_versions, pre_shared_keys)`

The server will reconstruct:

`ClientHelloInner' = (key_share, supported_groups, supported_versions, pre_shared_keys)`

These don't match, so the handshake will fail, both at the transcript and at checking the PSK binders. This isn't a security question because we haven't gotten there yet. It's a basic correctness question.

Now, the thing that went wrong here is that the entire process of swapping extensions out for outer_extensions *only works if the removed extensions are consecutive*. Thus, this PR adds "Removed extensions MUST be ordered consecutively in ClientHelloInner" to EncodedClientHelloInner construction. For that precondition to hold, the client must have ordered the extensions it wanted to compress, thus we add a matching rule, this sentence, to ClientHelloInner construction. Now everything works out and the compression process is correct.

davidben: Oh, I think I might see the misunderstanding.

> [...] the client just needs to be able to compute the same map from EncodedClientHelloInner to ClientHelloInner that the server does.

This does not work. The client does not compute EncodedClientHelloInner to ClientHelloInner at all. It cannot. By the time we've gotten this far, we've already generated all the PSK binders.

cjpatton: Oof, I forgot about the PSK binders :D Thanks for taking the time to clarify my misunderstanding.

chris-wood (approved):

PR #336, opened 2020-10-11
Replace inner_digest with ClientHelloOuterAAD

### LINKED ISSUE #323: Re-re-visit outer ClientHello binding (saves 65 bytes)

davidben:
Closes #323. The first commit shaves 33 bytes (or 49 with SHA-384) from the ciphertext by replacing inner_digest with ClientHelloOuterAAD. This also allows for more compact ClientHelloInner encodings when non-extension fields are expected to match. The second commit shaves a further 32 bytes.

We could probably further optimize the encoding but I've left it alone here. In particular, if we merge #316, all non-extension fields may as well be implicit. We could then replace the encryption payload with a plain extensions block.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -397,7 +397,7 @@ reference an extension in ClientHelloOuter.
 When sending ClientHello, the client first computes ClientHelloInner, including
 any PSK binders. It then computes a new value, the EncodedClientHelloInner, by
 first making a copy of ClientHelloInner. It then replaces the legacy\_session\_id
-field with an empty value.
+field with an empty string.
 
 The client then MAY substitute extensions which it knows will be duplicated in
 ClientHelloOuter. To do so, the client removes and replaces extensions from
@@ -426,10 +426,10 @@ ClientHelloInner.
 
 To prevent a network attacker from modifying the reconstructed ClientHelloInner
 (see {{flow-clienthello-malleability}}), ECH authenticates ClientHelloOuter by
-deriving a ClientHelloOuterAAD value. This is computed by serializing ClientHelloOuter
-with the "encrypted_client_hello" extension removed.
-ClientHelloOuterAAD is then passed as the associated data parameter to the
-HPKE encryption.
+deriving a ClientHelloOuterAAD value. This is computed by serializing
+ClientHelloOuter with the "encrypted_client_hello" extension removed.
+ClientHelloOuterAAD is then passed as the associated data parameter to the HPKE
+encryption.
 
 Note the decompression process in {{encoding-inner}} forbids
 "encrypted_client_hello" in OuterExtensions. This ensures the unauthenticated
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -426,8 +426,8 @@ ClientHelloInner.
 
 To prevent a network attacker from modifying the reconstructed ClientHelloInner
 (see {{flow-clienthello-malleability}}), ECH authenticates ClientHelloOuter by
-deriving a ClientHelloOuterAAD value. This is computed by removing the
-"encrypted_client_hello" extension from ClientHelloOuter and serializing it.
+deriving a ClientHelloOuterAAD value. This is computed by serializing ClientHelloOuter
+with the "encrypted_client_hello" extension removed.
 ClientHelloOuterAAD is then passed as the associated data parameter to the
 HPKE encryption.
 
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -388,34 +388,21 @@ may use the "outer_extensions" extension.
        outer_extensions(0xff04), (65535)
     } ExtensionType;
 
-    struct {
-       ExtensionType outer_extensions<2..254>;
-       uint8 inner_digest<32..255>;
-    } OuterExtensions;
+    ExtensionType OuterExtensions<2..254>;
 ~~~
 
 OuterExtensions consists of one or more ExtensionType values, each of which
-reference an extension in ClientHelloOuter, and a digest of the complete
-ClientHelloInner.
+reference an extension in ClientHelloOuter.
 
 When sending ClientHello, the client first computes ClientHelloInner, including
 any PSK binders. It then computes a new value, the EncodedClientHelloInner, by
 first making a copy of ClientHelloInner. It then MAY substitute extensions
 which it knows will be duplicated in ClientHelloOuter. To do so, the client
-computes the digest of the entire ClientHelloInner message as:
-
-~~~
-  inner_digest = Expand(Extract("", inner),
-                        "tls13 ech inner digest", Nh)
-~~~
-
-where `inner` is the ClientHelloInner structure and `Extract`, `Expand`, and
-`Nh` are as defined by the KDF API in {{!I-D.irtf-cfrg-hpke}}. Then, the client
 removes and replaces extensions from EncodedClientHelloInner with a single
 "outer_extensions" extension. Removed extensions MUST be ordered consecutively
-in ClientHelloInner. The list of outer extensions,
-OuterExtensions.outer_extensions, includes those which were removed from
-EncodedClientHelloInner, in the order in which they were removed.
+in ClientHelloInner. The list of outer extensions, OuterExtensions, includes
+those which were removed from EncodedClientHelloInner, in the order in which
+they were removed.
 
 Finally, EncodedClientHelloInner is serialized as a ClientHello structure,
 defined in Section 4.1.2 of {{RFC8446}}. Note this does not include the
@@ -425,15 +412,26 @@ The client-facing server computes ClientHelloInner by reversing this process.
 It scans EncodedClientHelloInner for an "outer_extensions" extension. If there
 is none, the ClientHelloInner is the EncodedClientHelloInner. Otherwise, it
 replaces the extension with the corresponding sequence of extensions in
-the ClientHelloOuter. If any referenced extensions are missing, the server
-MUST abort the connection with an "illegal_parameter" alert.
-The server then computes the digest of the reconstructed
-ClientHelloInner. If the digest does not equal OuterExtensions.inner_digest,
-then the server MUST abort the connection with a "decrypt_error" alert.
+the ClientHelloOuter. If any referenced extensions are missing or if
+"encrypted_client_hello" appears in the list, the server MUST abort the
+connection with an "illegal_parameter" alert.
 
 The "outer_extensions" extension is only used for compressing the
 ClientHelloInner. It MUST NOT be sent in either ClientHelloOuter or
 ClientHelloInner.
+
+## Authenticating the Outer ClientHello {#authenticating-outer-clienthello}
+
+To prevent a network attacker from modifying the reconstructed ClientHelloInner
+(see {{flow-clienthello-malleability}}), ECH authenticates ClientHelloOuter by
+deriving a ClientHelloOuterAAD value. This is computed by removing the
+"encrypted_client_hello" extension from ClientHelloOuter and serializing it.
+ClientHelloOuterAAD is then passed as the associated data parameter to the
+encryption.
+
+Note the decompression process in {{outer-extensions}} forbids
+"encrypted_client_hello" in OuterExtensions. This ensures the unauthenticated
+portion of ClientHelloOuter is not incorporated into ClientHelloInner.
 
 # Client Behavior {#client-behavior}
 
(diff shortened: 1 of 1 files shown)


davidben:
> We could then replace the encryption payload with a plain extensions block.

I guess that doesn't quite work, but we could replace it with the new random value, and an extensions block. Something like:

```
struct {
  Random random;
  Extension extensions<8..2^16-1>;
} EncodedClientHelloInner;
```

Anyway, a topic for a later PR, depending on how strongly we want to match cipher suites for #316. The size savings are a bit more marginal beyond that.

draft-ietf-tls-esni.md
```diff
@@ -425,16 +412,27 @@ The client-facing server computes ClientHelloInner by reversing this process.
...
 ClientHelloInner.
 
+## Authenticating the Outer ClientHello {#authenticating-outer-clienthello}
+
+To prevent a network attacker from modifying the reconstructed ClientHelloInner
+(see {{flow-clienthello-malleability}}), ECH authenticates ClientHelloOuter by
+deriving a ClientHelloOuterAAD value. This is computed by removing the
+"encrypted_client_hello" extension from ClientHelloOuter and serializing it.
+ClientHelloOuterAAD is then passed as the associated data parameter to the
+encryption.
```
cjpatton, line 432: s/to the encryption/for AEAD encryption.

cjpatton, line 432: `ClientHelloOuterAAD` is a bit different than we discussed in #323. In particular, I thought it was decided that we should predict the length of the "encrypted_client_hello" extension and authenticate as many zeros. What changed your mind? For the record, the approach is here seems a lot simpler to me :)

draft-ietf-tls-esni.md
```diff
@@ -425,16 +412,27 @@ The client-facing server computes ClientHelloInner by reversing this process.
...
-then the server MUST abort the connection with a "decrypt_error" alert.
+the ClientHelloOuter. If any referenced extensions are missing or if
+"encrypted_client_hello" appears in the list, the server MUST abort the
+connection with an "illegal_parameter" alert.
 
 The "outer_extensions" extension is only used for compressing the
 ClientHelloInner. It MUST NOT be sent in either ClientHelloOuter or
 ClientHelloInner.
 
+## Authenticating the Outer ClientHello {#authenticating-outer-clienthello}
```
cjpatton, line 425: Suggestion: s/Outer ClientHello/ClientHelloOuter.

draft-ietf-tls-esni.md
```diff
@@ -375,7 +375,7 @@ This document also defines the "ech_required" alert, which clients MUST send
 when it offered an "encrypted_client_hello" extension that was not accepted by
 the server. (See {{alerts}}.)
 
-## Incorporating Outer Extensions {#outer-extensions}
+## Encoding the Inner ClientHello {#encoding-inner-clienthello}
```
cjpatton, line 378: Suggestion: s/Inner ClientHello/ClientHelloInner.

draft-ietf-tls-esni.md
```diff
@@ -396,25 +396,27 @@ reference an extension in ClientHelloOuter.
...
 any PSK binders. It then computes a new value, the EncodedClientHelloInner, by
-first making a copy of ClientHelloInner. It then MAY substitute extensions
-which it knows will be duplicated in ClientHelloOuter. To do so, the client
-removes and replaces extensions from EncodedClientHelloInner with a single
-"outer_extensions" extension. Removed extensions MUST be ordered consecutively
-in ClientHelloInner. The list of outer extensions, OuterExtensions, includes
-those which were removed from EncodedClientHelloInner, in the order in which
-they were removed.
+first making a copy of ClientHelloInner. It then clears the legacy_session_id
+field.
```
cjpatton, line 400: Does this mean to send a 0-length `legacy_session_id`, or does this mean to send a `legacy_session_id` with the same length, but with all zeros? Presumably you mean the former?

cjpatton:
I like this change a lot. At some point while implementing ECH, I noted that it would be nice to have an explicit binding of CHOuter to the handshake in case of ECH acceptance. This provides that.

draft-ietf-tls-esni.md
davidben: Oh I probably was just unclear in that discussion. I never preferred the zeros formulation. :-) That was more me musing on how annoying it would be if there were another reason to prefer it. I like this better because, as you say, it's simpler.

draft-ietf-tls-esni.md
davidben: (Will deal with uploading a new revision after base PRs go through, since it's all a mess.)

Works for me, though it's unclear to me how much we like sticking identifiers in section titles, which is why I used the words instead. E.g. RFC8446's section titles say things like "Supported Versions" and this document had "Incorporating Outer Extensions" rather than "outer_extensions". Although, admittedly, "ClientHello" is itself an identifier, but I figured it was a more established one (elsewhere we have "Sending an encrypted ClientHello").

¯\\\_(ツ)_/¯

davidben: Oh, and we have "Outer ClientHello" as a section title down in Security Considerations, though I suppose that too is my fault. :-)

draft-ietf-tls-esni.md
```diff
@@ -375,53 +375,65 @@ This document also defines the "ech_required" alert, which clients MUST send
 when it offered an "encrypted_client_hello" extension that was not accepted by
 the server. (See {{alerts}}.)
 
-## Incorporating Outer Extensions {#outer-extensions}
+## Encoding the Inner ClientHello {#encoding-inner-clienthello}
```
chris-wood, line 378:
```suggestion
## Encoding ClientHelloInner {#encoding-inner-clienthello}
```

draft-ietf-tls-esni.md
```diff
@@ -375,53 +375,65 @@ This document also defines the "ech_required" alert, which clients MUST send
...
+
+The "outer_extensions" extension is only used for compressing the
+ClientHelloInner. It MUST NOT be sent in either ClientHelloOuter or
+ClientHelloInner.
 
-~~~
-  inner_digest = Expand(Extract("", inner),
-                        "tls13 ech inner digest", Nh)
-~~~
+## Authenticating the Outer ClientHello {#authenticating-outer-clienthello}
```
chris-wood, line 425:
```suggestion
## Authenticating ClientHelloOuter {#authenticating-outer-clienthello}
```

draft-ietf-tls-esni.md
davidben: Done. Though I used HPKE encryption since it looks like the rest of the document mostly refers to HPKE rather than the internal AEAD.

draft-ietf-tls-esni.md

draft-ietf-tls-esni.md

draft-ietf-tls-esni.md
davidben: Reworded to be clearer.

draft-ietf-tls-esni.md
davidben: Applied manually since I needed to rebase.

draft-ietf-tls-esni.md
davidben: Applied manually since I needed to rebase.

davidben:
Rebased and review comments addressed. The PR should also no longer be in a confusing state if you look at the diff.

cjpatton (approved):

draft-ietf-tls-esni.md
```diff
@@ -484,15 +483,21 @@ implementations need to take care to ensure that sensitive extensions are not
...
 guidance.
 
-To encrypt EncodedClientHelloInner, the client first needs to generate the HPKE
-encryption context. It computes the encapsulated key, context, HRR key (see
-{{client-hrr}}), and payload as:
+To encrypt EncodedClientHelloInner, the client first computes
+ClientHelloOuterAAD as described in {{authenticating-outer}}. Note this
+requires the "encrypted_client_hello" be computed after all other extensions.
+In particular, this is possible because the "pre_shared_key" extension is
+forbidden.
```
chris-wood, line 490:
```suggestion
forbidden in ClientHelloOuter.
```

chris-wood (approved):

draft-ietf-tls-esni.md
```diff
@@ -388,53 +388,53 @@ may use the "outer_extensions" extension.
...
-~~~
-
-where `inner` is the ClientHelloInner structure and `Extract`, `Expand`, and
-`Nh` are as defined by the KDF API in {{!I-D.irtf-cfrg-hpke}}. Then, the client
-removes and replaces extensions from EncodedClientHelloInner with a single
-"outer_extensions" extension. Removed extensions MUST be ordered consecutively
-in ClientHelloInner. The list of outer extensions,
-OuterExtensions.outer_extensions, includes those which were removed from
+first making a copy of ClientHelloInner. It then replaces the legacy_session_id
+field with an empty value.
```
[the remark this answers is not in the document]
davidben: The empty string. I previously wrote "clears the value", but @cjpatton objected to that one. :-P How about s/empty value/empty string/? We haven't introduced a new encoded structure yet (https://github.com/tlswg/draft-ietf-tls-esni/pull/336#issuecomment-707341408), so we need to put *something* in there.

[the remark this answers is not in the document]
chris-wood: So a zero-length string, as opposed to a 32 byte string of zeroes? (Whatever makes that more clear is fine by me.)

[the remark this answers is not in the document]
davidben: I mean the point is to encode fewer bytes, so yeah. I feel like, given the field is defined by `opaque legacy_session_id<0..32>`, it would be unreasonable to interpret "empty value" or "empty string" as a 32-byte string of zeros. :-P

I dunno, I think all the options so far have been pretty clear, so I'm not the best judge here. The intent is to stick the zero-length string in there. Which one do you prefer?
* clears the legacy_session_id field
* replaces the legacy_session_id field with an empty value
* replaces the legacy_session_id field with a zero-length value
* replaces the legacy_session_id field with an empty string
* replaces the legacy_session_id field with a zero-length string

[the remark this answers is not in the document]
cjpatton: +1 "replaces the legacy_session_id field with an empty string"

draft-ietf-tls-esni.md
```diff
@@ -426,8 +426,8 @@ ClientHelloInner.
 
 To prevent a network attacker from modifying the reconstructed ClientHelloInner
 (see {{flow-clienthello-malleability}}), ECH authenticates ClientHelloOuter by
-deriving a ClientHelloOuterAAD value. This is computed by removing the
-"encrypted_client_hello" extension from ClientHelloOuter and serializing it.
+deriving a ClientHelloOuterAAD value. This is computed by serializing ClientHelloOuter
+with the "encrypted_client_hello" extension removed.
 ClientHelloOuterAAD is then passed as the associated data parameter to the
 HPKE encryption.
```
cjpatton, line 432: nit: Adjust margins in this paragraph so each line breaks at 80 characters.

draft-ietf-tls-esni.md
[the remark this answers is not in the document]
davidben: Okay, empty string it is. I don't care, just pick one. :-)

draft-ietf-tls-esni.md

draft-ietf-tls-esni.md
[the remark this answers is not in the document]

cjpatton (approved):

PR #337, opened 2020-10-13
Tweak section titles and comparison against criteria.

davidben:
We use a mix of sentence case and title case in the document. The RFC style guide says to use title case. This PR aligns with that.
https://tools.ietf.org/html/rfc7322#section-3.4

Additionally, the ESNI criteria document has changed slightly since the comparison against criteria section was written. I've aligned the titles with the final wording in RFC8744. RFC8744 has also merged "Proper security context" and "Split server spoofing", so I've done the same here. "Support Multiple Protocols" also now talks about transport protocols and ALPN, so I've added a few words about that.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -306,7 +306,7 @@ Clients MUST parse the extension list and check for unsupported mandatory
 extensions. If an unsupported mandatory extension is present, clients MUST
 ignore the `ECHConfig`.
 
-# The "encrypted_client_hello" extension {#encrypted-client-hello}
+# The "encrypted_client_hello" Extension {#encrypted-client-hello}
 
 The encrypted ClientHelloInner is carried in an "encrypted_client_hello"
 extension, defined as follows:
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -1059,34 +1059,35 @@ less unusual and part of typical client behavior. In other words, if all Web
...
 
-### Forward secrecy
+### Maintain Forward Secrecy
 
 This design is not forward secret because the server's ECH key is static.
 However, the window of exposure is bound by the key lifetime. It is RECOMMENDED
 that servers rotate keys frequently.
 
-### Proper security context
+### Enable Multi-party Security Contexts
```
cjpatton, line 1068: s/party/Party/ ?

davidben: Yeah, I thought that too, but RFC8744 does it this way.
https://www.rfc-editor.org/rfc/rfc8744.html#name-enable-multi-party-security

¯\\\_(ツ)_/¯

chris-wood (approved):

PR #341, opened 2020-10-15
Fix up description of server message forwarding.

davidben:
This fixes a few issues with the  "Upon determining the true SNI"
paragraph:

- It talkes about determining the "true SNI", but we should probably
  talk about the ClientHelloInner more generally.

- It refers to behavior in "the following section" which no longer
  matches. This is a remnant of before Shared Mode was described as a
  special case of Split Mode, so remove it.

- It says the server forwards messages unmodified, which isn't true. It
  needs to forward the decrypted ClientHello. It also needs to process
  HRR.

I've cited the HRR section for the modified decryption procedure, which
doesn't really exist right now, but #339 is handling that.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -486,7 +486,7 @@ guidance.
 
 To encrypt EncodedClientHelloInner, the client first needs to generate the HPKE
 encryption context. It computes the encapsulated key, context, HRR key (see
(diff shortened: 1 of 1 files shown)


cjpatton (approved):
Yup, this aligns with my understanding!

chris-wood (approved):

PR #343, opened 2020-10-16
Pick from an unassigned range.

### LINKED ISSUE #322: Tidy up all codepoints before -08
Note to self. Let's pick a set and range that works.

chris-wood:
Closes #322.

cc @davidben, @cjpatton

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -233,7 +233,7 @@ The ECH configuration is defined by the following `ECHConfigs` structure.
         uint16 version;
         uint16 length;
         select (ECHConfig.version) {
-          case 0xff08: ECHConfigContents contents;
+          case 0xfe08: ECHConfigContents contents;
         }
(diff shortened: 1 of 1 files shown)


cjpatton (commented):
What will be our policy if we need to add additional extension points, e.g., to address #264? Do we start incrementing from `0xfe00`?

chris-wood:
> What will be our policy if we need to add additional extension points, e.g., to address #264? Do we start incrementing from 0xfe00?

Yeah, that's what I figured. I doubt we'll need more than 8, but we can pick a different range if that works?

cjpatton:
Yeah, this works for me. I don't see any reason why we should keep the code points dense.

cjpatton (approved):

davidben (commented):
Sure, those work. I usually `random.randint(0, 0xffff)` to avoid thinking about it, but whatever. :-)

davidben (approved):

Issue #342, opened 2020-10-16
Consider shortening config_id to 8 bytes


chris-wood:
This matches the acceptance signal length, and we've already accepted that probability of collision.

PR #344, opened 2020-10-16
Trial decryption and ECHConfig binding fixes

davidben:
The spec describes an optional trial decryption mode, but most of the text contradicts it. This PR fixes the text up and tries to unify their processing models. In doing so, it flips the error handling from decrypt_error to fallback to be consistent. This, combined with the ClientHelloOuterAAD change, plugs an active sticking out attack for free. It also makes #342 easier because GREASE config_id collisions with real config_ids are now harmless.

Along the way, @chris-wood and @cjpatton observed that we're not authenticating the ECHConfig at all. (It used to be authenticated by way of the config_id, which was a digest.) Instead, just pass the ECHConfig into the info field of the HPKE context. They also noticed the server rule about PSKs on ClientHelloOuter didn't work with GREASE. Instead, we can simply remove it.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -782,15 +782,11 @@ unmodified.
 
 Otherwise, if all candidate ECHConfigs fail to decrypt the extension, the
 client-facing server MUST ignore the extension and proceed with the connection
-using ClientHelloOuter, with the following added behavior:
-
-- It MUST include the "encrypted_client_hello" extension in its
-  EncryptedExtensions with the "retry_configs" field set to one or more
-  ECHConfig structures with up-to-date keys. Servers MAY supply multiple
-  ECHConfig values of different versions. This allows a server to support
-  multiple versions at once.
-- If offered, the server MUST ignore the "pre_shared_key" extension in the
-  ClientHello.
+using ClientHelloOuter. This connection proceeds as usual, except the server
+MUST include the "encrypted_client_hello" extension in its EncryptedExtensions
+with the "retry_configs" field set to one or more ECHConfig structures with
+up-to-date keys. Servers MAY supply multiple ECHConfig values of different
+versions. This allows a server to support multiple versions at once.
 
 Note that decryption failure could indicate a GREASE ECH extension (see
 {{grease-extensions}}), so it is necessary for servers to proceed with the
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -722,7 +722,7 @@ MAY offer to resume sessions established without ECH.
 
 # Server Behavior {#server-behavior}
 
-## Client-Facing Server
+## Client-Facing Server {#client-facing-server}
 
 Upon receiving an "encrypted_client_hello" extension, the client-facing server
 determines if it will accept ECH, prior to negotiating any other TLS parameters.
@@ -806,9 +806,12 @@ transmitted on the wire by the client:
 1. If CH1 contains the "encrypted_client_hello" extension but CH2 does not, or
    if CH2 contains the "encrypted_client_hello" extension but CH1 does not, then
    the server MUST abort the handshake with an "illegal_parameter" alert.
-1. Suppose the "encrypted_client_hello" extension is sent in both CH1 and CH2,
-   If the configuration identifier (see {{ech-configuration}}) differs between
-   CH1 and CH2, then the server MUST abort with an "illegal_parameter" alert.
+1. If the "encrypted_client_hello" extension is sent in both CH1 and CH2,
+   the server follows the procedure in {{client-facing-server}} to decrypt the
+   second ClientHelloOuter, but it uses the previously-selected ECHConfig as
+   the set of candidate ECHConfigs. If decryption fails, the server aborts the
+   connection with a "decrypt_error" alert rather than continuing the handshake
+   with the second ClientHelloOuter.
 
 [[OPEN ISSUE: If the client-facing server implements stateless HRR, it has no
 way to send a cookie, short of as-yet-unspecified integration with the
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -494,15 +492,17 @@ encapsulated key, context, HRR key (see {{client-hrr}}), and payload as:
 
 ~~~
     pkR = Deserialize(ECHConfig.public_key)
-    enc, context = SetupBaseS(pkR, "tls13 ech")
+    enc, context = SetupBaseS(pkR, "tls13 ech\0" + ECHConfig)
```
cjpatton, line 495: I wonder if `\0` is the right way to right the null byte. (RFC8446 doesn't encode 0x00 in a string at all: see [RFC8446, 4.4.3](https://tools.ietf.org/html/rfc8446#section-4.4.3).) I'm fine with it as-is, since I think it's pretty clear what you mean ... I just wanted flag it in case there's a "correct" way to write this.

draft-ietf-tls-esni.md
```diff
@@ -512,6 +512,11 @@ a `ClientECH` with the following values:
 - `enc`, as computed above; and
 - `payload`, as computed above.
 
+If optional configuration identifiers (see {{optional-configs}})) are used, the
+`config_id` field MAY be empty or randomly generated. Unless specified by the
```
cjpatton, line 516: Let's drop "randomly generated"? I think this would be OK, given how client-facing server behavior is specified. But I don' think we want to put this idea in people's head as a viable option. (Yet.)

draft-ietf-tls-esni.md
```diff
@@ -512,6 +512,11 @@ a `ClientECH` with the following values:
...
+application using (D)TLS or externally configured on both sides,
+implementations MUST NOT use this mode.
```
cjpatton, line 518: Suggest changing "MUST NOT use this mode" to "MUST compute config_id as specified in {{whatever-section-describes-ClientECH}}".

draft-ietf-tls-esni.md
```diff
@@ -714,68 +722,54 @@ MAY offer to resume sessions established without ECH.
...
-  ClientHello.
-
-Note that an unrecognized ClientECH.config_id value may be a GREASE ECH
-extension (see {{grease-extensions}}), so it is necessary for servers to proceed
-with the connection and rely on the client to abort if ECH was required. In
-particular, the unrecognized value alone does not indicate a misconfigured ECH
-advertisement ({{misconfiguration}}). Instead, servers can measure occurrences
-of the "ech_required" alert to detect this case.
+First, the server collects a set of candidate ECHConfigs. This set is
+determined in one of the two following methods:
```
cjpatton, line 734: s/determined in/determined by/

draft-ietf-tls-esni.md
```diff
@@ -797,9 +806,12 @@ transmitted on the wire by the client:
...
    the server MUST abort the handshake with an "illegal_parameter" alert.
-1. Suppose the "encrypted_client_hello" extension is sent in both CH1 and CH2,
-   If the configuration identifier (see {{ech-configuration}}) differs between
-   CH1 and CH2, then the server MUST abort with an "illegal_parameter" alert.
+1. If the "encrypted_client_hello" extension is sent in both CH1 and CH2,
+   the server follows the procedure in {{client-facing-server}} to decrypt the
+   second ClientHelloOuter, but it uses the previously-selected ECHConfig as
+   the set of candidate ECHConfigs. If decryption fails, the server aborts the
+   connection with a "decrypt_error" alert rather than continuing the handshake
+   with the second ClientHelloOuter.
```
cjpatton, line 814: This works for me. However, to align with the new text above, if decryption fails here then the server should ignore ECH and continue with ClientHelloOuter.

cjpatton (changes requested):
This is pretty close!

draft-ietf-tls-esni.md
davidben: Oh? {{optional-configs}} has explicitly said it can be random for a while now. (GREASE prescribes a random one too.)

draft-ietf-tls-esni.md

draft-ietf-tls-esni.md

draft-ietf-tls-esni.md
```diff
@@ -797,9 +807,12 @@ transmitted on the wire by the client:
 1. If CH1 contains the "encrypted_client_hello" extension but CH2 does not, or
    if CH2 contains the "encrypted_client_hello" extension but CH1 does not, then
    the server MUST abort the handshake with an "illegal_parameter" alert.
-1. Suppose the "encrypted_client_hello" extension is sent in both CH1 and CH2,
-   If the configuration identifier (see {{ech-configuration}}) differs between
-   CH1 and CH2, then the server MUST abort with an "illegal_parameter" alert.
+1. If the "encrypted_client_hello" extension is sent in both CH1 and CH2,
+   the server follows the procedure in {{client-facing-server}} to decrypt the
+   second ClientHelloOuter, but it uses the previously-selected ECHConfig as
```
cjpatton, line 812: s/ClientHelloOuter/ClientHelloInner/?

draft-ietf-tls-esni.md
cjpatton: Right again!

cjpatton (changes requested):
One last thing to check.

draft-ietf-tls-esni.md
davidben: Hah! I meant to write "decrypt the second ClientHelloOuter [into the second ClientHelloInner]", but I guess you more naturally read it as "decrypt the second ClientHelloInner [from of the second ClientHelloOuter]". Which means I should probably find a different wording all together. Will wordsmith.

davidben: How's this new version?

chris-wood: LGTM!

cjpatton (approved):

chris-wood (approved):

PR #345, opened 2020-10-16
Only extract 8 bytes for the config_id

### LINKED ISSUE #342: Consider shortening config_id to 8 bytes

davidben:
This doesn't bind anything, so no sense wasting bytes. Closes #342.

(Note: this is orthogonal to #344, but it makes more sense with that PR.)

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -334,12 +334,11 @@ provided in the corresponding `ECHConfig.cipher_suites` list.
 
 config_id
 : The configuration identifier, equal to
-`Expand(Extract("", config), "tls ech config id", Nh)`, where `config` is the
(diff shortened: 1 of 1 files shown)


davidben:
I guess this makes the text in #344 a little odd because "Compare ClientECH.config_id against identifiers of known ECHConfigs and select *the* one that matches, if any, as a candidate.", is a little less clear. If we want to do this, it maybe should say "and select any that match", even though there'll really just be one match.

davidben:
(Rebased and made the text talk about multiple potential matches.)

ekr (approved):
I agree this is fine.

chris-wood (approved):

PR #347, opened 2020-10-19
Fix ServerHello.random behavior in older TLS versions.

davidben:
As written, the backend server performs this ServerHello.random modification unconditionally when it sees an empty ECH extension. Although the client is required to only offer TLS 1.3, the backend server doesn't check this.

However, as noted elsewhere in the document, these same 8 bytes are used for the TLS 1.3 downgrade protection mechanism. That mechanism only works if the server *never* produces a TLS 1.2 ServerKeyExchange signature without the signal, so it's important that the server *not* signal ECH acceptance. Write this down.


cjpatton (approved):
Yup, this is a good idea.

chris-wood (approved):
Oops. Good catch!

PR #351, opened 2020-10-30
Rearrange the server HRR text.

davidben:
Now that we have separate client HRR and server HRR sections, move some
server text to the server section. Also clarify that the second
ClientHelloOuterAAD is computed from the second ClientHelloOuter, not
reusing the first value.

(Some cleanup in advance of writing up the HPKE context reuse version.)


ekr (approved):
Can we merge this? The number of PRs is getting confusing.

@chris-wood

chris-wood (approved):

Issue #325, opened 2020-10-06
Specify cross-HRR key rotation issue(s)


chris-wood:
If a server implements ECH decryption via an RPC call, it's possible for the client's ECH keys to be valid for CH1 but invalid for CH2, perhaps due to a race condition that updates the server's keys across these messages. This should be exceedingly rare in practice, so failing the connection with an alert is probably fine, but we should probably be clear about this.

cc @cjpatton

cjpatton:
There are two ways a accept-then-HRR-then-reject can happen:
1. The client sends a `config_id` after HRR that differs from the `config_id` it sent before HRR.
1. The `config_id` is the same, but the server no longer recognizes it.

The latter is the server's fault, and I think the only thing to do is abort with "internal_error". However, I think requiring that the `config_id` doesn't change is a good idea. I wonder if this is implied by the client behavior?

> If the server sends a HelloRetryRequest in response to the ClientHello, the client sends a second updated ClientHello per the rules in [RFC8446]. However, at this point, the client does not know whether the server processed ClientHelloOuter or ClientHelloInner, and MUST regenerate both values to be acceptable. Note: if the inner and outer ClientHellos use different groups for their key shares or differ in some other way, then the HelloRetryRequest may actually be invalid for one or the other ClientHello, in which case a fresh ClientHello MUST be generated, ignoring the instructions in HelloRetryRequest. Otherwise, the usual rules for HelloRetryRequest processing apply.

In any case, I think we should have the server abort with "illegal_parameter" if the `config_id` changes over HRR.

davidben:
Is this any different from, say, the server's cipher suite capabilities changing in between CH1 and CH2 and thus the server not being able to meet what they said with HRR?

I guess I've always assumed that, within the context of a single "handshake" from the perspective of the client, we expect the server to behave self-consistently and, if it can't, I guess the server can internal_error and live with failing one or two connections across server restarts.

cjpatton:
> Is this any different from, say, the server's cipher suite capabilities changing in between CH1 and CH2 and thus the server not being able to meet what they said with HRR?

I think this is a little different, since it can be triggered by changing the set of ECHConfigs the server knows, and not necessarily the set of cipher suites.

> I guess I've always assumed that, within the context of a single "handshake" from the perspective of the client, we expect the server to behave self-consistently and, if it can't, I guess the server can internal_error and live with failing one or two connections across server restarts.

[...]

Issue #333, opened 2020-10-11
Stateless HRR and split mode


davidben:
I was thinking about #316 and the discussion on server HRR checks started by https://github.com/tlswg/draft-ietf-tls-esni/issues/325#issuecomment-705796082, and I'm not sure HRR cookies even work in split mode right now. First off, suppose we don't merge something like #316. The current text says:

> Note: if the inner and outer ClientHellos use different groups for
> their key shares or differ in some other way, then the HelloRetryRequest may	
> actually be invalid for one or the other ClientHello, in which case a fresh	
> ClientHello MUST be generated, ignoring the instructions in HelloRetryRequest.	
> Otherwise, the usual rules for HelloRetryRequest processing apply

This means if you get a cookied HRR only good for ClientHelloInner, there's no requirement to stick the cookie in ClientHelloOuter! But now the client-facing server can't see the cookie, so it can't be stateless. It would seem like #316 fixes this, but now we have a more fundamental problem: whose cookie?

Suppose the client-facing server wishes to handle HRR statelessly. It decrypts the first ClientHelloInner and forwards it to the backend server. If the backend server responds with HRR, the client-facing server must forward the HRR as-is, or the transcript breaks. But the client-facing server has no opportunity to bind its state (ech_hrr_key) into the HRR cookie. (Perhaps there isn't even one. If there is, it's encrypted with the backend server's key anyway. Either way, it's fixed now.)

The simplest solution to all this is to only support stateful client-facing servers in split mode, unless the client-facing and backend servers can coordinate enough to share one cookie. I don't know if that would satisfy everyone. I could imagine doing something horrific to HRR.cookie, akin to what we did to ServerHello.random, but that sounds really messy.

cjpatton:
Let's mark this as an "OPEN ISSUE" in the draft.

davidben:
> Let's mark this as an "OPEN ISSUE" in the draft.

Done. https://github.com/tlswg/draft-ietf-tls-esni/pull/338

Issue #326, opened 2020-10-06
Include ECH context in ClientHelloOuterAAD


chris-wood:
@davidben mentions possibly binding the outer CH to the ECH encryption in #323. If we don't do that, we should at least bind the encryption to the non-encrypted parts of the ECH extension (`config_id`, `suite`, and so on).

cc @cjpatton

chris-wood:
Closing in favor of #323. We can revisit the ECH extension AD information later if desired.

chris-wood:
Re-opening to consider adding *all* data to AAD. @davidben wrote out the flow if we did this:

- Make your CHInner. This should largely be running through your usual CH assembly logic.
- Encode it with the extensions you expect to compress.
- Make your CHOuter, minus the ECH extension. This should largely run through your existing CH assembly logic.
- Run Deserialize and SetupBaseS.
- Compute CHOuterAAD with half an ECH extension.
- Take EncodedCHInner and CHOuterAAD and run through the context.Seal logic to assemble the final CHOuter.

So, not great, but also not too terrible. I'd prefer we avoid falling into the "gee, I sure wish we bound more to the encryption" state, given the history of ECH, though I agree that this isn't cheap. 

@ekr, what do you think?

cjpatton:
To be clear, the suggestion is to change AEAD encryption so that it authenticates `ClientECH.cipher_suite`, `.config_id`, and `.enc`, correct? This change seems sensible to me. One might argue that this is redundant because these parameters are incorporated into the transcript. On the other hand, ECH usage is resolved before the transcript is known! It seems like a good idea to authenticate these parameters before consuming them.

davidben:
They aren't incorporated on ECH accept, which is mildly interesting. ECH accept also doesn't negotiate with them, so it's probably okay, though there is a "don't stick out" failure (depending on threat model) for the optional config_id flow if we don't authenticate them.

PR #352, opened 2020-10-30
Use the same HPKE context between the two ClientHellos

### LINKED ISSUE #349: Avoid computing a new HPKE contexts on HRR?
ECH currently adds an ECDH operation on *both* the first and second ClientHello. That's extra asymmetric operations, and if folks end up offloading those keys, it's two RPCs. Given that we need to bind the two decryptions together anyway, and HPKE supports multiple encrypted messages (it has an internal sequence number), perhaps it's better to just use that.

Here's a possibility: the first ClientHello makes a new HPKE context and calls context.Seal on it. The second ClientHello simply calls context.Seal on the same context, which internally increments the sequence number. Probably we copy over the first ClientHello's cipher_suite, config_id, and enc, even though they don't really do anything.

The server, if stateful, just calls context.Open to mirror the client behavior. If stateless, probably the simplest thing is to pickle the AEAD key inside the cookie. [Edit: I accidentally wrote "ticket" earlier.] It probably could also recompute the HPKE context and bump the sequence number up by one, maybe checking the selected ECHConfig and enc values didn't change. (The text already requires checking the former.)

### LINKED ISSUE #356: Clients should send consistent GREASE ECH extensions between CH1 and CH2
I'll try to put together a PR for this later, but filing this now so I don't forget. (@dmcardle FYI) The draft currently says you randomize this both times, which won't work.

Also "MUST check the extension syntactically" appears to have been a little ambiguous, since how deeply you parse ECHConfigs gets a little complicated. I'll adjust that too.

davidben:
(Note: this PR sits on top of #351, though hopefully the combined diff is also readable enough.)

Closes #349. This achieves the same security properties of ech_hrr_key, but gives us a few improvements:

- Client-facing servers that offload their HPKE operations to RPCs do not need to worry about race conditions with rotation between the two ClientHellos. (See #325.)

- We only perform one addition asymmetric operation with ECH, rather than sometimes two. For servers with offloaded ECH keys, this also avoids an extra RPC call.

- HPKE implementations for ECH only need SetupBase* and not SetupPSK*.

- The description of HRR handling in general can be greatly simplified. Rather than running the server's CH1 algorithm with various modifications, we can just write a straightforward, simpler algorithm.

This PR also adds some missing text for GREASE on HRR, which closes #356. (Although it's worth noting that HRR connections can be easily distinguished by an active attacker anyway. This PR doesn't change this, but makes it much clearer.)

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -504,13 +504,12 @@ In particular, this is possible because the "pre_shared_key" extension is
 forbidden in ClientHelloOuter.
 
 The client then generates the HPKE encryption context. Finally, it computes the
-encapsulated key, context, HRR key (see {{client-hrr}}), and payload as:
+encapsulated key, context, and payload as:
 
 ~~~
     pkR = Deserialize(ECHConfig.public_key)
     enc, context = SetupBaseS(pkR,
                               "tls ech" || 0x00 || ECHConfig)
-    ech_hrr_key = context.Export("tls ech hrr key", 32)
     payload = context.Seal(ClientHelloOuterAAD,
                            EncodedClientHelloInner)
 ~~~
@@ -663,31 +662,26 @@ may actually be invalid for one or the other ClientHello, in which case a fresh
 ClientHello MUST be generated, ignoring the instructions in HelloRetryRequest.
 Otherwise, the usual rules for HelloRetryRequest processing apply.
 
-Clients bind encryption of the second ClientHelloInner to encryption of the
-first ClientHelloInner via the derived ech_hrr_key by running a modified
-encryption process:
+The client encodes the second ClientHelloInner as in {{encoding-inner}}, using
+the second ClientHelloOuter for any referenced extensions. It then encrypts
+the new EncodedClientHelloInner value as a second message with the previous
+HPKE context:
 
 ~~~
-    pkR = Deserialize(ECHConfig.public_key)
-    enc, context = SetupPSKS(pkR, "tls ech" || 0x00 || ECHConfig,
-                             ech_hrr_key, "hrr key")
     payload = context.Seal(ClientHelloOuterAAD,
                            EncodedClientHelloInner)
 ~~~
 
-The `info` parameter to SetupPSKS is the concatenation of "tls ech", a
-zero byte, and the serialized ECHConfig. ClientHelloOuterAAD is computed from
-the second ClientHelloOuter as described in {{authenticating-outer}}. Note the
-encrypted value is also authenticated by ech_hrr_key. The rationale for this
-is described in {{flow-hrr-hijack}}.
+ClientHelloOuterAAD is computed as described in {{authenticating-outer}}, but
+again using the second ClientHelloOuter. The `cipher_suite`, `config_id`, and
+`enc` fields in the new ClientECH structure are copied from the first
+ClientHelloOuter. Note an HPKE context maintains a sequence number, so this
+operation internally uses a fresh nonce for the AEAD operation. Reusing the
+HPKE context avoids an attack described in {{flow-hrr-hijack}}.
 
 If the client offered ECH in the first ClientHello, then it MUST offer ECH in
 the second. Likewise, if the client did not offer ECH in the first ClientHello,
 then it MUST NOT not offer ECH in the second.
-
-[[OPEN ISSUE: Should we be using the PSK input or the info input?  On the one
-hand, the requirements on info seem weaker, but maybe actually this needs to be
-secret? Analysis needed.]]
 
 ## GREASE ECH {#grease-ech}
 
@@ -708,6 +702,11 @@ structure available for the server, it SHOULD send a GREASE {{?RFC8701}}
   is the ciphertext expansion of the selected AEAD scheme and L is the size of
   the EncodedClientHelloInner the client would compute when offering ECH, padded
   according to {{padding}}.
+
+When sending a second ClientHello in response to a HelloRetryRequest, the
+client copies the "cipher_suite", "config_id", and "enc" fields. It generates a
+new "payload" field, using the length of a padded second EncodedClientHelloInner
+for L.
 
 If the server sends an "encrypted_client_hello" extension, the client MUST check
 the extension syntactically and abort the connection with a "decode_error" alert
@@ -733,11 +732,11 @@ described in {{backend-server}}.
 
 ## Client-Facing Server {#client-facing-server}
 
-Upon receiving a non-empty "encrypted_client_hello" extension, the client-facing
-server determines if it will accept ECH, prior to negotiating any other TLS
-parameters. Note that successfully decrypting the extension will result in a new
-ClientHello to process, so even the client's TLS version preferences may have
-changed.
+Upon receiving a non-empty "encrypted_client_hello" extension in an initial
+ClientHello, the client-facing server determines if it will accept ECH, prior
+to negotiating any other TLS parameters. Note that successfully decrypting the
+extension will result in a new ClientHello to process, so even the client's TLS
+version preferences may have changed.
 
 First, the server collects a set of candidate ECHConfigs. This set is
 determined by one of the two following methods:
@@ -769,7 +768,6 @@ corresponding to ECHConfig, as follows:
                          "tls ech" || 0x00 || ECHConfig)
     EncodedClientHelloInner = context.Open(ClientHelloOuterAAD,
                                            ClientECH.payload)
-    ech_hrr_key = context.Export("tls ech hrr key", 32)
 ~~~
 
 ClientHelloOuterAAD is computed from ClientHelloOuter as described in
@@ -782,11 +780,11 @@ considering candidate ECHConfigs.
 
 Upon determining the ClientHelloInner, the client-facing server then forwards
 the ClientHelloInner to the appropriate backend server, which proceeds as in
-{{backend-server}}. If the backend server responds with a HelloRetryRequest, the
-client-facing server forwards it, decrypts the client's second ClientHelloOuter
-using the modified procedure in {{server-hrr}}, and forwards the resulting
-second ClientHelloInner. The client-facing server forwards all other TLS
-messages between the client and backend server unmodified.
+{{backend-server}}. If the backend server responds with a HelloRetryRequest,
+the client-facing server forwards it, decrypts the client's second
+ClientHelloOuter using the procedure in {{server-hrr}}, and forwards the
+resulting second ClientHelloInner. The client-facing server forwards all other
+TLS messages between the client and backend server unmodified.
 
 Otherwise, if all candidate ECHConfigs fail to decrypt the extension, the
 client-facing server MUST ignore the extension and proceed with the connection
@@ -805,34 +803,35 @@ unrecognized value alone does not indicate a misconfigured ECH advertisement
 
 ### Handling HelloRetryRequest {#server-hrr}
 
-In case a HelloRetryRequest (HRR) is sent, the client-facing server MUST
-consistently accept or decline ECH between the two ClientHellos, using the same
-ECHConfig, and abort the handshake if this is not possible. This is achieved as
-follows. Let CH1 and CH2 denote, respectively, the first and second ClientHello
-transmitted on the wire by the client:
+After sending a HelloRetryRequest, the client-facing server does not repeat the
+steps in {{client-facing-server}} with the second ClientHelloOuter. Instead it
+continues the selections from the first ClientHelloOuter as follows:
 
-1. If CH1 contains the "encrypted_client_hello" extension but CH2 does not, or
-   if CH2 contains the "encrypted_client_hello" extension but CH1 does not, then
-   the server MUST abort the handshake with an "illegal_parameter" alert.
-1. If the "encrypted_client_hello" extension is sent in CH2, the server follows
-   the procedure in {{client-facing-server}} to decrypt the extension, but it
-   uses the previously-selected ECHConfig as the set of candidate ECHConfigs.
-   If decryption fails, the server aborts the connection with a "decrypt_error"
-   alert rather than continuing the handshake with the second ClientHelloOuter.
-
-When decrypting the second ClientECH.payload, the client-facing server
-performs a corresponding process to {{client-hrr}}:
+If the client-facing server accepted ECH, it checks the second ClientHelloOuter
+also contains the "encrypted_client_hello" extension. If not, it MUST abort the
+handshake with a "missing_extension" alert. Otherwise, it decrypts the new
+ClientECH.payload as a second message with the previous HPKE context:
 
 ~~~
-    context = SetupPSKR(ClientECH.enc, skR,
-        "tls ech" || 0x00 || ECHConfig, ech_hrr_key, "hrr key")
     EncodedClientHelloInner = context.Open(ClientHelloOuterAAD,
                                            ClientECH.payload)
 ~~~
 
-ClientHelloOuterAAD is computed from the second ClientHelloOuter as described
-in {{authenticating-outer}}. The `info` parameter to SetupPSKR is the
-concatenation of "tls ech", a zero byte, and the serialized ECHConfig.
+ClientHelloOuterAAD is computed as described in {{authenticating-outer}}, but
+using the second ClientHelloOuter. If decryption fails, the client-facing
+server MUST abort the handshake with a "decrypt_error" alert. Otherwise, it
+reconstructs the second ClientHelloInner from the new EncodedClientHelloInner
+as described in {{encoding-inner}}, using the second ClientHelloOuter for
+any referenced extensions.
+
+The client-facing server then forwards the resulting ClientHelloInner to the
+backend server. It forwards all subsequent TLS messages between the client and
+backend server unmodified.
+
+If the client-facing server rejected ECH, or if the first ClientHello did not
+include an "encrypted_client_hello" extension, the client-facing server
+proceeds with the connection as usual, without decrypting any new
+ClientECH.payload value.
 
 [[OPEN ISSUE: If the client-facing server implements stateless HRR, it has no
 way to send a cookie, short of as-yet-unspecified integration with the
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -648,46 +647,27 @@ may actually be invalid for one or the other ClientHello, in which case a fresh
...
+    payload = context.Seal(ClientHelloOuterAAD,
+                           EncodedClientHelloInner)
 ~~~
 
-ClientHelloOuterAAD is computed from the second ClientHelloOuter as described
-in {{authenticating-outer}}. The `info` parameter to SetupPSKR is computed as
-above.
+ClientHelloOuterAAD is computed as described in {{authenticating-outer}}, but
+again using the second ClientHelloOuter. The `cipher_suite`, `config_id`, and
+`enc` fields in the new ClientECH structure are copied from the first
```
davidben, line 662: This fields also don't matter. I wonder if we should say you empty them instead...

(Insert usual remark that I think we got HelloRetryRequest wrong. :-) )

draft-ietf-tls-esni.md
```diff
@@ -801,20 +785,35 @@ of the "ech_required" alert to detect this case.
...
-
-1. If CH1 contains the "encrypted_client_hello" extension but CH2 does not, or
-   if CH2 contains the "encrypted_client_hello" extension but CH1 does not, then
-   the server MUST abort the handshake with an "illegal_parameter" alert.
-1. If the "encrypted_client_hello" extension is sent in CH2, the server follows
-   the procedure in {{client-facing-server}} to decrypt the extension, but it
-   uses the previously-selected ECHConfig as the set of candidate ECHConfigs.
-   If decryption fails, the server aborts the connection with a "decrypt_error"
-   alert rather than continuing the handshake with the second ClientHelloOuter.
+After sending a HelloRetryRequest, the client-facing server does not repeat the
```
davidben, line 788: Self-review: "After sending a HelloRetryRequest" is slightly odd because, on ECH accept, the client-facing server forwards the HRR from the backend server. At the same time, this section also talks about ECH reject, since we want to clarify that you don't process ECH on the second ClientHello as usual.

Perhaps: "After sending or forwarding a HelloRetryRequest"?

draft-ietf-tls-esni.md
cjpatton: They need to be present in order for stateless-HRR servers to correctly re-compute the HPKE context, right? Incidentally, how do stateless-HRR servers correctly compute the nonce?

davidben: Stateless-HRR servers don't work at all right now anyway (#333), so we can't really talk about what goes in the cookie because there's no way for the client-facing server to contribute to the cookie right now. When we do, yes, we should say something about this. (See also https://github.com/tlswg/draft-ietf-tls-esni/issues/349#issuecomment-713724970)

As to what the stateless-HRR server needs, it needs to store *something* that allows it to recreate the HPKE context on the second ClientHello. Things that would work:

* Store the HPKE encryption key, nonce, and sequence number. (Alternatively, know the sequence number is always 1...)
* Store the client `enc` value and sequence number (or assume 1), and some handle to the ECHConfig private key, so it can recreate the HPKE context and fast-forward the context to the right sequence number.
* Store a hash of the client `enc` value, check it matches, and then do as above.

Only the last of these actually needs the client to send `enc` again. Note what would not work great is blindly recomputing the `enc` value, given {{flow-hrr-hijack}}. Given #325, my inclination is the first option is most straightforward anyway.

cjpatton: > Stateless-HRR servers don't work at all right now anyway (#333), so we can't really talk about what goes in the cookie because there's no way for the client-facing server to contribute to the cookie right now.

Isn't this only true for Split Mode / stateless-HRR?

draft-ietf-tls-esni.md
cjpatton: That sounds good to me.

draft-ietf-tls-esni.md
```diff
@@ -787,34 +785,35 @@ of the "ech_required" alert to detect this case.
...
-   uses the previously-selected ECHConfig as the set of candidate ECHConfigs.
-   If decryption fails, the server aborts the connection with a "decrypt_error"
-   alert rather than continuing the handshake with the second ClientHelloOuter.
-
-When decrypting the second ClientECH.payload, the client-facing server
-performs a corresponding process to {{client-hrr}}:
+If the client-facing server accepted ECH, it checks the second ClientHelloOuter
+also contains the "encrypted_client_hello" extension. If not, it MUST abort the
+handshake with a "missing_extension" alert. Otherwise, it decrypts the new
+ClientECH.payload as a second message with the previous HPKE context:
```
cjpatton, line 795: As long as "cipher_suite", "config_id", "enc" are retransmitted in CH2, should the server verify that they don't change?

draft-ietf-tls-esni.md
davidben: It's true for both. There's no text supporting Shared Mode / stateless-HRR right now because Shared Mode is treated as a special-case of Split Mode. It's "obvious" what they're supposed to do, but without any text to support it, we can't talk about what goes in the cookie.

As part of #333, we'll need to address all of that. In the meantime, this PR doesn't change things because we already have no text talking about how ech_hrr_key goes in the cookie. All of that is #333.

draft-ietf-tls-esni.md
davidben: Checking CH1 and CH2 consistency is a huge mess to do. We completely got the design of HRR wrong in RFC8446, and it's too late to do anything about that now. Rather than expend a whole lot of energy on everyone's part, from spec text to implementations to constantly having this discussion, I'd rather we follow RFC8446's example here and leave it alone.

davidben: That is, RFC8446 is already quite restrictive about CH1/CH2 consistency, yet servers are not required to (and in practice do not) check consistency for much, if any, of it.
https://tools.ietf.org/html/rfc8446#section-4.1.2

Indeed if you think about what a server would need to do here, it's clear we just got the design wrong. A server that is capable of checking some individual field most likely stored the value of the field, in which case there was no reason for the client to re-send it in the first place! (This grew organically from an earlier design where the server actually reconstructed CH1 from CH2, but that had all kinds of attacks and was generally a mess. When we dropped that, we probably should have stopped to reevaluate it. Oh well.)

draft-ietf-tls-esni.md
cjpatton: Ack. You can resolve this comment.

draft-ietf-tls-esni.md
ekr: This seems like it's orthogonal, right?

ekr (approved):
This seems generally good. I'm not sure why we didn't do this in the first place. Are reusable contexts new?

draft-ietf-tls-esni.md
davidben: Well, the question is whether we mandate the server check the redundant fields. I think my default is to prefer matching rfc8446 in not saying much. (With a side of grumpiness about how weird HRR is. :-) ) We probably need to do a better job describing HRR in rfc8446bis. I figure the less other specs say about it, the more room we have to figure out what we actually want to require.

chris-wood:
This is much simpler, but, if I understand it correctly, I think it *might* be a little bit of a regression. The previous export-then-inject design forced the server to process both ECH configs and encapsulated keys in both CH1 and CH2, whereas here the server just ignores those in CH2 and uses the values from CH1. That lets clients change these fields in CH2 without any effect. Now, I don't really know of anything that *could* go wrong here, but it smells like the sort of "insufficient binding" problem we've continually run into before. (For example, I think an on-path attacker could modify the `enc` or `config_id` fields in CH2, without touching anything else. Decryption would succeed, as the server used CH1's `enc` value, and the transcript is unaffected since the server would choose the CH inner value. Is that a way for an on-path attacker to quickly check if the connection is real or fake ECH?)

Chris asked if we should require servers to verify these match, and as you pointed out, that's a mess. (We also acknowledged that servers missing these checks is something we should expect to happen in the wild, which is somewhat why we are concerned about this re-binding attack in the first place.)

I wonder if we can solve this in a different way: [put these ECH extension fields in the AAD](https://github.com/tlswg/draft-ietf-tls-esni/issues/326). That would prevent the modification above, I think, and shouldn't be *too* bad in terms of additional complexity.

davidben:
> This is much simpler, but, if I understand it correctly, I think it might be a little bit of a regression. The previous export-then-inject design forced the server to process both ECH configs and encapsulated keys in both CH1 and CH2, whereas here the server just ignores those in CH2 and uses the values from CH1. That lets clients change these fields in CH2 without any effect. Now, I don't really know of anything that could go wrong here, but it smells like the sort of "insufficient binding" problem we've continually run into before.

I think allowing the actual client to change ignored fields is a general property of HelloRetryRequest. Because we don't expect servers to check consistency between CH1 and CH2, a client that goes out of their way to send something inconsistent may find that their (CH1, CH2) tuple is interpreted ambiguously. But the transcript reflects the whole tuple, so, as far as the handshake itself, the client is just causing its own preferences to be misrepresented here.

Ideally ambiguous things would be unrepresentable and consistently treated as errors but HRR is what it is. :-/

> (For example, I think an on-path attacker could modify the enc or config_id fields in CH2, without touching anything else. Decryption would succeed, as the server used CH1's enc value, and the transcript is unaffected since the server would choose the CH inner value. Is that a way for an on-path attacker to quickly check if the connection is real or fake ECH?)

I think real vs. fake ECH on HRR connections is entirely hopeless, short of completely redoing HRR. If CH1 uses ECH, we need to require CH2 also use ECH. That means ECH decryption failures in CH2 are fatal, and you can distinguish the two by simply replaying CH1 and perturbing the ciphertext in CH2. However, note that CHOuterAAD means whether a ClientHello triggers HRR is not under attacker control.

That said, I think there's a variant of this on CH1 in the optional config_id mode. (See below.)

> Chris asked if we should require servers to verify these match, and as you pointed out, that's a mess. (We also acknowledged that servers missing these checks is something we should expect to happen in the wild, which is somewhat why we are concerned about this re-binding attack in the first place.)

Yeah, this is the "I think we got HRR wrong and have to live with it" problem. :-)

> I wonder if we can solve this in a different way: put these ECH extension fields in the AAD. That would prevent the modification above, I think, and shouldn't be too bad in terms of additional complexity.

I don't think I believe this matters for HRR (only attack I can see is real vs fake ECH, which is hopeless with HRR anyway), but now that you mention it, I think there *is* a variant of this for the optional config_id mode in the non-HRR case.

In CH1, we don't bind config_id, enc, and cipher_suite to the encryption, so a priori an attacker might be able to perturb them without changing ECH. Perturbing enc and cipher_suite seems likely to break ECH for other reasons, but config_id is interesting. In the normal mode, it too will break ECH. In the optional config_id mode, config_id is just a placeholder. That means an attacker can distinguish optional config_id ECH from {GREASE or config_id ECH} by flipping the config_id, sending it to the server, and seeing if the client can decrypt the result.

(Of course, all this is relative to an unclear and undefined "don't stick out" model.)

Complexity-wise, I think it's fine, but note it changes the processing order somewhat interestingly, both for the spec and implementations. Right now the spec says:

1. Make your CHInner. This should largely be running through your usual CH assembly logic.
2. Encode it with the extensions you expect to compress.
3. Make your CHOuter, minus the ECH extension. This should largely run through your existing CH assembly logic and computes CHOuterAAD.
4. Take EncodedCHInner and CHOuterAAD and run through the Deserialize, SetupBaseS, context.Seal logic to assemble the final CHOuter.

In order to put these fields in AAD, we'd need to say:

1. Make your CHInner. This should largely be running through your usual CH assembly logic.
2. Encode it with the extensions you expect to compress.
3. Make your CHOuter, minus the ECH extension. This should largely run through your existing CH assembly logic.
4. Run Deserialize and SetupBaseS.
5. Compute CHOuterAAD with half an ECH extension.
6. Take EncodedCHInner and CHOuterAAD and run through the context.Seal logic to assemble the final CHOuter.

draft-ietf-tls-esni.md
```diff
@@ -801,20 +785,35 @@ of the "ech_required" alert to detect this case.
...
+as described in {{encoding-inner}}, using the second ClientHelloOuter for
+any referenced extensions.
+
+The client-facing server then forwards the resulting ClientHelloInner to the
+backend server. It forwards all subsequent TLS messages between the client and
+backend server unmodified.
+
+If the client-facing server rejected ECH, or if the first ClientHello did not
+include an "encrypted_client_hello" extension, the client-facing server
+proceeds with the connection as usual, without decrypting any new
```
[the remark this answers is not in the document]
davidben: Hrm. I wrote "any" so that this text also works for the no ECH extension case and the "if any" version had lots of clauses and commas. :-)

> If the client-facing server rejected ECH, or if the first ClientHello did not
> include an "encrypted_client_hello" extension, the client-facing server
> proceeds with the connection as usual, without decrypting the new
> ClientECH.payload value, if any.

Given that, are you happy with "any new ClientECH.payload value"? Or do you prefer ", if any" or maybe some other wording altogether?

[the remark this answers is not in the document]
chris-wood: "Any" to me implied "possibly one of the many new values," but the CH only has one value.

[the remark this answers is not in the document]
davidben: Hehe. I read "any ClientECH.payload value [singular]" to mean zero or one, but maybe that's not the right way to use that word. :-)

How about:

> If the client-facing server rejected ECH, or if the first ClientHello did not
> include an "encrypted_client_hello" extension, the client-facing server
> proceeds with the connection as usual. The server does not decrypt the
> second ClientHello's ClientECH.payload value, if one was included.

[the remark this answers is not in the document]
chris-wood: Perfect!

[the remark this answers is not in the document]
davidben: Reworded.

cjpatton:
> In CH1, we don't bind config_id, enc, and cipher_suite to the encryption, so a priori an attacker might be able to perturb them without breaking the decryption. Perturbing enc and cipher_suite seems likely to break decryption anyway, but config_id is interesting. In the normal mode, it too will break decryption. In the optional config_id mode, config_id is just a placeholder. That means an attacker can distinguish optional config_id ECH from {GREASE or config_id ECH} by flipping the config_id, sending it to the server, and seeing if the client can decrypt the result.

A similar "don't stick out" distinguisher: Flip a bit of ClientECH.payload in CH2. The client-facing server will abort with "decrypt_error" if real-ECH is in use but not if grease-ECH is in use. This attack is unavoidable if we want to ensure that the server doesn't change from CHInner to CHOuter across the HRR. (Enforcing this invariant seems like a good idea to me.)

davidben:
Right, I mentioned that HRR attack above. ( "I think real vs. fake ECH on HRR connections is entirely hopeless [...]".)

Since avoiding this is hopeless in HRR, I don't think we should use it as a justification for binding bits of CH2's ClientECH or checking they match (which don't actually do the same thing). Whereas attacks on CH1 are interesting because that situation is less obviously hopeless. So maybe we want to add more things to CHOuterAAD for that, though that is orthogonal to this PR, so I think we should deal with it separately.

cjpatton:
Agreed. (I was in the middle of editing my comment above to say the same thing when I saw your message!)

cjpatton (approved):

chris-wood:
>This attack is unavoidable if we want to ensure that the server doesn't change from CHInner to CHOuter across the HRR. 

Yeah, this morning I came to the conclusion that mitigating this sort of distinguisher for HRR was not really possible unless we relax the inner/outer matching invariant. I'm fine pursuing more AAD as a separate change on top of this, since it's strictly an improvement (and HRR is hopeless).

chris-wood:
@davidben can you please resolve conflicts?

davidben:
Rebased.

draft-ietf-tls-esni.md
davidben: Unresolving this since we only resolved @cjpatton's clarification question, not the actual issue.

Thinking about this some more, I think we actually should require `enc` be empty on CH2. The reason is, of the three stateless implementation options listed above, only the first one is robust. The second and third rely on the server performing a check that, if they forget, still interops. But if they forget the check, {{flow-hrr-hijack}} applies again. Given @cjpatton's question, I also suspect this mistake is reasonably likely.

Clearing `enc` forces servers go pick the first option, which is the more robust one. We're also guaranteed that a server, stateful or stateless, does not need `enc` (or `cipher_suite`) because you need to process ECH on CH1 regardless. Thus, I propose:

* CH2's `enc` MUST be empty. Servers check this when parsing (since it's easy).
* CH2's  `cipher_suite` MUST be {copied over, zero; not sure which}. Servers check this one too, just because it's really easy.

Even more minimal would be if CH2's extension contained `payload` and nothing else, but we don't have any precedent for extension syntaxes varying between CH1 and CH2, so probably we shouldn't go down that route.

@ekr, I'm curious what your thoughts on this are.

davidben: Oh same probably goes for `config_id` in CH2. Force the server to use the same ECHConfig rather than accidentally swapping it.

draft-ietf-tls-esni.md
```diff
@@ -663,52 +662,61 @@ may actually be invalid for one or the other ClientHello, in which case a fresh
...
-
 ## GREASE ECH {#grease-ech}
 
 If the client attempts to connect to a server and does not have an ECHConfig
 structure available for the server, it SHOULD send a GREASE {{?RFC8701}}
-"encrypted_client_hello" extension as follows:
+"encrypted_client_hello" extension in the first ClientHello as follows:
 
-- Set the "cipher_suite" field to a supported ECHCipherSuite. The selection
+- Set the `cipher_suite` field to a supported ECHCipherSuite. The selection
```
davidben, line 700: I have no idea when we use quotation marks or backticks, but we seem to use backticks when referring to ClientECH fields?

draft-ietf-tls-esni.md
davidben: (Updated the PR with this version.)

draft-ietf-tls-esni.md
davidben: (Supposing we go with https://github.com/tlswg/draft-ietf-tls-esni/pull/352#discussion_r518868173, the redundant fields are gone and this is moot.)

draft-ietf-tls-esni.md
cjpatton: I've been going with backticks for structures and their fields and quotes for extension names.

cjpatton (approved):
LGTM.

chris-wood (approved):

PR #357, opened 2020-11-06
Correcting incorrect integer equivalents for ECH and outer_extensions

sudheesh001:
Signed-off-by: Sudheesh Singanamalla <sudheesh@cloudflare.com>

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -317,7 +317,7 @@ extension, defined as follows:
 
 ~~~
     enum {
-       encrypted_client_hello(0xfe08), (65535)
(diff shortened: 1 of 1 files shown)


sudheesh001:
@chris-wood This is ready for review.

chris-wood (approved):

davidben:
I don't think this is right. It's not an integer equivalent but a max value for sizing the enum.
https://tools.ietf.org/html/rfc8446#section-3.5

(The TLS presentation syntax is sometimes confusing.)

chris-wood:
Oh, well that's annoying. And confusing. I'll revert this.

sudheesh001:
Ah that was confusing. Apologies.

Issue #358, opened 2020-11-06
ECH contradicts second ClientHello consistency requirements in RFC8446


davidben:
I just noticed this as I was pondering our various HRR issues. RFC8446, section 4.1.2 says:

> [...] The client will also send a ClientHello when the server has responded to its ClientHello with a HelloRetryRequest.  In that case, the client MUST send the same ClientHello without modification, except as follows:
>
> * [...various cases regarding existing extensions...]
>
> *  Other modifications that may be allowed by an extension defined in the future *and present in the HelloRetryRequest*.

https://tools.ietf.org/html/rfc8446#section-4.1.2

The "and present in the HelloRetryRequest" portion is fun. In RFC8446 as written, we're not allowed to change the `encrypted_client_hello` extension on the second ClientHello unless the server sent `encrypted_client_hello` in HelloRetryRequest, which it doesn't. With the protocol as-is, it seems we'll at least need some text to deal with the contradiction.

The impact is practice is thankfully limited. If the servers does not enforce consistency between the two ClientHellos, it doesn't matter. (They're not required to, but they're not explicitly forbidden from it either, and a stateless server may need to enforce some minimum consistency to avoid getting into a broken state.) If a server does, ECH's behavior on HRR will break it. Such a server is supposed to handshake with ClientHelloOuter, so it just affects the retry flow. That's maybe okay, but mildly annoying. Administrators using server software that checks need to know to first upgrade to a version that doesn't check before trying to deploy ECH.

We can also add an ECH extension to HRR. Then, on ECH-reject + HRR, the server perhaps wouldn't send the extension and the client would be required to replay the extension. This is a waste in many ways, but would work. This would break don't stick out, but some threat models, don't stick out with HRR is hopeless anyway. (See https://github.com/tlswg/draft-ietf-tls-esni/pull/352#issuecomment-721824226.) It would also be weirdly asymmetric between ECH and SH.

That said, this constraint is also pretty obnoxious. For instance, #313 currently adds a new padding extension codepoint, because the old one isn't defined for other messages. However, the RFC8446 allowance only works for the existing padding code point, so you don't want to send any other code points in ClientHello, lest the server force you to keep it unchanged in ClientHello2. So maybe we should confirm no servers did the fully strict check and undo that constraint? (Although, for #313, I personally favor the padding message strategy anyway.)

davidben:
[...]

PR #361, opened 2020-11-18
Copy the entire extension on HRR + GREASE for now.

davidben:
This is a temporary mitigation for the compatibility risks from #358, as
any clients experimenting with GREASE will impact servers that otherwise
don't expect ECH. To resolve this issue overall, we may need a more
involved solution.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -712,10 +712,11 @@ structure available for the server, it SHOULD send a GREASE {{?RFC8701}}
   according to {{padding}}.
 
 When sending a second ClientHello in response to a HelloRetryRequest, the
-client copies the `cipher_suite` field from the first ClientHello. It sets
-`config_id`, and `enc` to the empty string. Finally, it generates a new
-`payload` field, using the length of a padded second EncodedClientHelloInner
-for L.
+client copies the entire "encrypted_client_hello" extension from the first
+ClientHello.
+
+[[OPEN ISSUE: The above doesn't match HRR handling for either ECH acceptance or
+rejection. See issue #358.]]
 
 If the server sends an "encrypted_client_hello" extension, the client MUST check
 the extension syntactically and abort the connection with a "decode_error" alert


draft-ietf-tls-esni.md
```diff
@@ -712,10 +712,11 @@ structure available for the server, it SHOULD send a GREASE {{?RFC8701}}
...
 When sending a second ClientHello in response to a HelloRetryRequest, the
-client copies the `cipher_suite` field from the first ClientHello. It sets
-`config_id`, and `enc` to the empty string. Finally, it generates a new
-`payload` field, using the length of a padded second EncodedClientHelloInner
-for L.
+client copies the entire "encrypted_client_hello" extension from the first
+ClientHello.
+
+[[OPEN ISSUE: The above doesn't match HRR handling for either ECH acceptance or
+rejection. See issue #358.]]
```
cjpatton, line 719: I'm confused by the word "match" here. Do you mean to say that greased ECH looks different from real ECH after HRR?

davidben: Yes. This is #358. There is a potential compatibility issue with changing CH2's ECH extension when talking to a server that doesn't expect ECH. We also need to change CH2's ECH extension when actually speaking ECH.

This PR resolves the compatibility issue, just so we can get this out of the way of draft-09. It does not resolve the bulk of #358 which interacts with every other HRR problem.

cjpatton: I think this is a good idea for ECH-09, I'm merely suggesting that the wording of the OPEN ISSUE could be clearer.

davidben: Ah. What wording would you prefer? Perhaps: "OPEN ISSUE: The extension value is not copied when sending a real ECH extension. See issue #358."

cjpatton: That's better, but I think we should make this an explicit don't-stick-out issue. Slight revision: "OPEN ISSUE: The extension value is not copied when sending a real ECH extension, causing this code path to stick out. See issue #358."

davidben: It's not just about sticking out. ECH acceptance and ECH rejection currently also violate the RFC8446 requirement. This PR doesn't change it because it would break them, and it's mostly okay compatibility-wise because it would only affect servers that have opted in via DNS. (But only mostly. The recovery flow on a strictly enforcing server is broken.)

davidben: That's why I specifically wanted this to just be a quick summary and then a pointer to the issue where we'll capture the actual discussion in a more fluid form.

chris-wood: Yeah, I think the comment is fine as-is.

chris-wood (approved):
Given the compatibility concerns @davidben found, and based on discussion in IETF 109, I think this is prudent to merge for -09. Doing so now.

PR #364, opened 2020-11-20
Remove backticks from empty string literal.

dmcardle:
The backtick-quoted empty string translates to """" in the text output.


dmcardle:
Not sure if this is the best solution, but the quadruple quote in the txt is goofy.

chris-wood (approved):

chris-wood:
This seems fine, yeah. Thanks for catching it!

PR #365, opened 2020-11-21
Quick editorial pass.

dmcardle:
Tried to clarify some text, PTAL :)

For the backend server, I think the `ClientHelloInner..ServerHelloECHConf` notation is a little confusing (pedantically, shouldn't it be `...`?). Can't we just concatenate the two messages and call it a day?

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -500,8 +500,8 @@ it does a standard ClientHello, with the exception of the following rules:
    {{flow-client-reaction}}.)
 1. It MUST include an "encrypted_client_hello" extension with a payload
    constructed as described below.
-1. The value of `ECHConfig.contents.public_name` MUST be placed in the "server_name"
-   extension.
+1. The value of `ECHConfig.contents.public_name` MUST be placed in the
+   "server_name" extension.
 1. It MUST NOT include the "pre_shared_key" extension. (See
    {{flow-clienthello-malleability}}.)
 
@@ -537,10 +537,10 @@ encapsulated key, context, and payload as:
 ~~~
 
 Note that the HPKE functions Deserialize and SetupBaseS are those which match
-`ECHConfig.contents.kem_id` and the AEAD/KDF used with `context` are those which match
-the client's chosen preference from `ECHConfig.contents.cipher_suites`. The `info`
-parameter to SetupBaseS is the concatenation of "tls ech", a zero byte, and
-the serialized ECHConfig.
+`ECHConfig.contents.kem_id` and the AEAD/KDF used with `context` are those which
+match the client's chosen preference from `ECHConfig.contents.cipher_suites`.
+The `info` parameter to SetupBaseS is the concatenation of "tls ech", a zero
+byte, and the serialized ECHConfig.
 
 The value of the "encrypted_client_hello" extension in the ClientHelloOuter is
 a `ClientECH` with the following values:
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -942,7 +942,7 @@ It then computes a string
     accept_confirmation =
         Derive-Secret(Handshake Secret,
                       "ech accept confirmation",
-                      ClientHelloInner || ServerHelloECHConf)
+                      ClientHelloInner...ServerHelloECHConf)
 ~~~
 
 where Derive-Secret and Handshake Secret are as specified in {{RFC8446}},
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -280,7 +280,7 @@ extensions
 generating a ClientHello message. These are described below
 ({{config-extensions}}).
 
-The client-facing server advertises a sequence ECH configurations to clients,
+The client-facing server advertises a sequence of ECH configurations to clients,
 serialized as follows.
 
 ~~~~
@@ -366,16 +366,16 @@ retry_configs
 decreasing order of preference, to be used by the client in subsequent
 connection attempts.
 
-This document also defines the "ech_required" alert, which clients MUST send
+This document also defines the "ech_required" alert, which the client MUST send
 when it offered an "encrypted_client_hello" extension that was not accepted by
 the server. (See {{alerts}}.)
 
 ## Encoding the ClientHelloInner {#encoding-inner}
 
-Some TLS 1.3 extensions can be quite large and having them both in
-ClientHelloInner and ClientHelloOuter will lead to a very large overall size.
-One particularly pathological example is "key_share" with post-quantum
-algorithms. In order to reduce the impact of duplicated extensions, the client
+Some TLS 1.3 extensions can be quite large, thus repeating them in the
+ClientHelloInner and ClientHelloOuter can lead to an excessive overall size.
+One pathological example is "key_share" with post-quantum
+algorithms. To reduce the impact of duplicated extensions, the client
 may use the "ech_outer_extensions" extension.
 
 ~~~
@@ -591,7 +591,7 @@ ClientHello extensions can be computed in this way.
 In contrast, clients do not know the longest SNI value in the client-facing
 server's anonymity set without server input. For the "server_name" extension
 with length D, clients SHOULD use the server's length hint L
-(ECHCOnfig.maximum_name_length) when computing the padding as follows:
+(ECHConfig.maximum_name_length) when computing the padding as follows:
 
 1. If L >= D, add L - D bytes of padding. This rounds to the server's
    advertised hint, i.e., ECHConfig.maximum_name_length.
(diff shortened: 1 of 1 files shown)


chris-wood (approved):
The simplified notation LGTM. @cjpatton, please have look!

draft-ietf-tls-esni.md
```diff
@@ -944,13 +944,11 @@ It then computes a string
     accept_confirmation =
         Derive-Secret(Handshake Secret,
                       "ech accept confirmation",
-                      ClientHelloInner..ServerHelloECHConf)
+                      ClientHelloInner || ServerHelloECHConf)
```
cjpatton, line 947: Actually, the `..` notation is meant to denote the sequence of messages between `ClientHelloInner` and `ServerHelloECHConf`. This sequence is empty here, but I think we should use the notation nonetheless, in order to be consistent with RFC8446.

draft-ietf-tls-esni.md
```diff
@@ -944,13 +944,11 @@ It then computes a string
...
 ~~~
 
 where Derive-Secret and Handshake Secret are as specified in {{RFC8446}},
-Section 7.1 and ClientHelloInner..ServerHelloECHConf refers to the sequence of
-handshake messages beginning with the first ClientHello and ending with
-ServerHelloECHConf. Finally, the backend server constructs its ServerHello
+Section 7.1. Finally, the backend server constructs its ServerHello
```
cjpatton, line 951: If we revert the above change, don't forget to revert this change as well.

cjpatton (changes requested):
- I like `..` because it's consistent with RFC 8446 (See the key schedule section.)
- `make` fails because of lines that exceed 80 characters.

draft-ietf-tls-esni.md
chris-wood: I don’t think this is necessary. || is more clear in this context given that the sequence is always empty.

cjpatton: What about incorporating the CH1 hash? Isn't that also something that's implied by the `..` notation?

draft-ietf-tls-esni.md
```diff
@@ -460,8 +460,8 @@ otherwise.
 
 To offer ECH, the client first chooses a suitable ECH configuration. To
 determine if a given `ECHConfig` is suitable, it checks that it supports the KEM
-algorithm identified by `ECHConfig.kem_id`, at least one KDF/AEAD algorithm
-identified by `ECHConfig.cipher_suites`, and the version of ECH indicated by
+algorithm identified by `ECHConfigContents.kem_id`, at least one KDF/AEAD algorithm
```
cjpatton, line 463: In this section, it might be clearer to refer to `ECHConfig.contents.kem_id`, since `ECHConfig` has been defined, but `ECHConfigContents` only implicitly refers to `ECHConfig.contents`.

draft-ietf-tls-esni.md
cjpatton: In other words, it denotes the whole transcript from ClientHelloInner to ServerHelloECHConf, including the messages on the HRR path.

cjpatton: (@chris-wood and I spoke offline and we agreed that the `..` notation should stay as-is.)

dmcardle: I find the range syntax a bit confusing here. If there were messages in between, it's not clear to me which variant of those messages would be selected, or whether some transformation would be applied to them.

I don't feel really strongly about this. Happy to revert it, but to be fully pedantic, I want to point out that RFC 8446 uses three dots between messages. I'll go ahead and use three dots here, unless two dots has a different meaning that I don't understand.

draft-ietf-tls-esni.md
dmcardle: Done!

draft-ietf-tls-esni.md
dmcardle: Agreed, made this change for the whole "Offering ECH" section.

draft-ietf-tls-esni.md
cjpatton: Oops, three dots is right! https://tools.ietf.org/html/rfc8446#section-7.1

I think my dyslexic brain is removing  a dot.

draft-ietf-tls-esni.md
```diff
@@ -591,10 +591,10 @@ ClientHello extensions can be computed in this way.
 In contrast, clients do not know the longest SNI value in the client-facing
 server's anonymity set without server input. For the "server_name" extension
 with length D, clients SHOULD use the server's length hint L
-(ECHCOnfig.maximum_name_length) when computing the padding as follows:
+(ECHConfigContents.maximum_name_length) when computing the padding as follows:
```
cjpatton, line 594: Is `ECHConfig.config.maximum_name_length` more appropriate here?

draft-ietf-tls-esni.md
```diff
@@ -591,10 +591,10 @@ ClientHello extensions can be computed in this way.
...
 
 1. If L >= D, add L - D bytes of padding. This rounds to the server's
-   advertised hint, i.e., ECHConfig.maximum_name_length.
+   advertised hint, i.e., ECHConfigContents.maximum_name_length.
```
cjpatton, line 597: `ECHConfig.contents.maximum_name_length`?

draft-ietf-tls-esni.md
```diff
@@ -622,7 +622,7 @@ usual, authenticating the connection for the true server name.
 #### Rejected ECH
 
 If the server used ClientHelloOuter, the client proceeds with the handshake,
-authenticating for ECHConfig.public_name as described in {{auth-public-name}}.
+authenticating for ECHConfigContents.public_name as described in {{auth-public-name}}.
```
cjpatton, line 625: `ECHConfig.contents.public_name`?

draft-ietf-tls-esni.md
```diff
@@ -942,11 +942,11 @@ It then computes a string
     accept_confirmation =
         Derive-Secret(Handshake Secret,
                       "ech accept confirmation",
-                      ClientHelloInner..ServerHelloECHConf)
+                      ClientHelloInner...ServerHelloECHConf)
```
cjpatton, line 945: Nice catch!

draft-ietf-tls-esni.md
```diff
@@ -458,9 +458,9 @@ otherwise.
 
 To offer ECH, the client first chooses a suitable ECH configuration. To
 determine if a given `ECHConfig` is suitable, it checks that it supports the KEM
-algorithm identified by `ECHConfig.kem_id`, at least one KDF/AEAD algorithm
-identified by `ECHConfig.cipher_suites`, and the version of ECH indicated by
-`ECHConfig.version`. Once a suitable configuration is found, the client selects
+algorithm identified by `ECHConfig.contents.kem_id`, at least one KDF/AEAD algorithm
+identified by `ECHConfig.contents.cipher_suites`, and the version of ECH indicated by
+`ECHConfig.contents.version`. Once a suitable configuration is found, the client selects
```
cjpatton, line 463: Do these lines break 80 characters?

cjpatton (changes requested):
Looks good, modulo minor nits!

draft-ietf-tls-esni.md
dmcardle: Yep, good catch! Reflowing the paragraph in a new commit.

draft-ietf-tls-esni.md
dmcardle: Done

draft-ietf-tls-esni.md
dmcardle: Done

draft-ietf-tls-esni.md
dmcardle: Seems reasonable, done.

draft-ietf-tls-esni.md
```diff
@@ -528,16 +529,16 @@ The client then generates the HPKE encryption context and computes the
...
     enc, context = SetupBaseS(pkR,
                               "tls ech" || 0x00 || ECHConfig)
     payload = context.Seal(ClientHelloOuterAAD,
                            EncodedClientHelloInner)
 ~~~
 
 Note that the HPKE functions Deserialize and SetupBaseS are those which match
-`ECHConfig.kem_id` and the AEAD/KDF used with `context` are those which match
-the client's chosen preference from `ECHConfig.cipher_suites`. The `info`
+`ECHConfig.contents.kem_id` and the AEAD/KDF used with `context` are those which match
```
cjpatton, line 540: This line and the one below it are longer than 80 characters. (Sorry if this is annoying ... lines that are longer than 80 characters cause `make` to fail. In fact, you should be getting emails from CI telling you that the build is failing.)

cjpatton (changes requested):

draft-ietf-tls-esni.md
dmcardle: Aha, I hadn't run make in a few commits, my bad. Fixed now.

cjpatton (approved):
Thanks!

Issue #369, opened 2020-11-24
Potential SNI leak via cross-ECH resumption


kjacobs-moz:
Martin pointed out a potential SNI leak via resumption under the following conditions:
1. A client connects to server "A" without ECH, and receives a session ticket.
2. The client later resumes the connection to server "A" using ECH. An observer can see that a PSK was accepted (via SH containing pre_shared_key). 
4. If any of the hosts previously visited by the client are known, an attacker can do a lookup of each for a config_id matching the resumed session. 

This effectively reduces the size of the anonymity set for a particular client/victim.

While it's true that the hostname is already leaked by the first connection without ECH, this seems to violate the security goals as they are written. We used to require separation between non-ECH and ECH PSKs, but it was removed in #285.

davidben:
Ah fun. I agree that observing a resumption does increase the probability the connection was one of the previous non-ECH connections the client made, when server just deployed ECH. (Though the client could also be visiting someone else in the anonymity set for a second time.) I'm not sure I follow what's going on in (3) though. Is that just checking if one of the previously-visited hosts shares a client-facing server with the connection under attack? That seems like something the attacker can regardless. (With or without resumption, one would expect a client's visited sites pre-ECH-deployment to somewhat predict with the ones post-ECH-deployment.)

As an aside, even if SH.pre_shared_key was hidden, the observer can infer resumption based on whether the server responded with a lot of data (it sent a certificate in there), or very little data (it sent no certificate in there). Unless the server adds some truly wasteful padding on resumption connections.

kjacobs-moz:
(3) just provides additional evidence that the SNI guess is correct (by further constraining set of possible hostnames).

cjpatton:
To mitigate this, it ought to be sufficient to have the client track whether a given PSK was established ruing a ECH handshake or a non-ECH handshake. It seems like this would be a lot more involved for a server to manage.

davidben:
Is it actually additional evidence? It doesn't seem to narrow the usual ECH anonymity set, just identifies it. If I understand you correctly, (3) is:

* The attacker observes a connection from the client with some config_id and some secret SNI A.
* The attacker queries DNS or the recovery flow[*] for the cleartext SNIs that the client previously visited and sees if one of them, B, has an ECHConfig with matching config_id.
* If there is a match, the attacker learns that B was in the same anonymity set as A.

[...]

PR #371, opened 2020-12-11
Clarify privacy goal scope

kjacobs-moz:
Addresses #369

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -1177,10 +1177,11 @@ in-band when possible, such as through the use of OCSP stapling, and clients
 SHOULD take steps to minimize or protect such requests during certificate
 validation.
 
-Attacks that rely on analysis of non-ECH traffic to infer server identity in an
-ECH connection are considered to be out of scope for this document. For example,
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -1177,6 +1177,11 @@ in-band when possible, such as through the use of OCSP stapling, and clients
 SHOULD take steps to minimize or protect such requests during certificate
 validation.
 
+Attacks that rely on analysis of non-ECH traffic to infer server identity in an
+ECH connection are considered to be out of scope for this document. For example,
+a client that connects to a particular host prior to ECH deployment may resume a
+previous session or exhibit similar connection patterns under ECH.
```
chris-wood, line 1183:
```suggestion
Attacks that rely on non-ECH traffic to infer server identity in an ECH connection are 
out of scope for this document. For example, a client that connects to a particular host 
prior to ECH deployment may later resume a connection to that same host after ECH 
deployment, thereby linking the resulting ECH connection to the original non-ECH 
connection.
```

kjacobs-moz:
Done. Thanks!

Issue #177, opened 2019-07-24
GREASE ESNI extensions are easily distinguished from real ones


davidben:
The draft has provisions for GREASE ESNI extensions, per "Do not stick out". However, this is not very effective if an attacker can distinguish a GREASE ESNI extension from a real one.

#154 tried to address this, but it doesn't work if the (public) ESNI record is known. There's also a simpler attack: replay the ClientHello with one byte of ciphertext corrupted. ESNI currently distinguishes key mismatch from decrypt failure, with the latter resulting in an alert. We can fix that by saying decrypt failure should behave like key mismatch. This still leaves record_digest.

The natural fix is to drop record_digest in favor of trial decryption. The cost is the server must perform a DH operation per known key. The number of keys the server needs depends on its DNS TTLs and key rotation. (I think it's 1 + ceil(dns_ttl / time_between_rotation), plus some leeway[*].)

Instead, we can keep a key name, but make the space dense. Suppose we replace it with a B-bit "key phase". The server picks a random starting key phase. Then, each time it rotates the keys, it increments the phase, with wraparound. The server then gets 2<sup>B</sup> keys for free. If it needs more, it still needs trial decryption but gains a 2<sup>B</sup> multiplicative factor in the cost. On the flip side, an individual GREASE extension with a random key phase has a 1/2<sup>B</sup> chance of colliding with a particular ESNI key.

I'm not sure we actually need that many live keys, so B = 2 is probably plenty. Or even B = 1 if we say you should just rotate slower than your DNS TTL.

[*] Note the robustness mechanism means that server doesn't need to cover _all_ clients with stale keys. The retry is expensive, so the server still needs to cover _almost all_ of them, but it can cut off the long tail.

davidben:
Mostly a note to myself for the actual PR: the draft should also discuss padding of the EncryptedExtensions message.

ekr:
I think we really need to go back to first principles here and ask what it is we are trying to accomplish.  First, it's public information whether a given server supports ECHO, so if an attacker wants to know that, it can mostly just ask [there is some complexity around getting candidate names, but note that you can always get the public name from the SNI value and any real attacker will see a lot of non-ECHO traffic and so will know a bunch of SNI values].

What's not public information is whether a given *connection* uses ECHO. The attacker might not know this either because:

1. It knows that the server is ECHO-supporting.
1. It has a limited view of just the connection and isn't willing to do do the work of trying to determine if the server is ECHO-supporting.

[...]

PR #376, opened 2021-01-26
Move to a key identity in lieu of the config identifier hash.

### LINKED ISSUE #375: Replace config_id with a server-chosen key_id
After very useful discussion with many folks, it seems now is the time to remove config_id. There are a couple reasons to do this:

- Servers need to know the correct hash function to use when computing the ID, and possibly need to compute multiple IDs for a given public key (since a config can have multiple hash-based KDFs), which is somewhat annoying and redundant.
- The entire ECHConfig is already authenticated, so using a hash to prove that the client made a determination based on the correct config contents is unnecessary.
- Servers will likely only ever have a small handful of keys in motion at any point in time, so anything more than a couple of bits to identify one of said keys is excessive.
- ... there are probably more.

The original design [had a server-chosen label, but we removed it](https://github.com/tlswg/draft-ietf-tls-esni/pull/43). (Oops!) It seems we're now in a better position to bring this back, but with some simplifications.

chris-wood:
We may want to change how we deal with GREASE (does a randomly selected byte stick out?), but this should address the issue for now. I think we can (should) treat these issues separately, and focus just on the `key_id` simplification here.

Closes #375.

cc @davidben, @cjpatton, @martinthomson, @ekr


ekr:
My intuition is that one byte is too short; I think 32 bits would be wiser. However, this is a slightly less serious problem than usual because we can rev the ECHConfig version number and the ECH extension if we find something longer is needed.

chris-wood:
> My intuition is that one byte is too short; I think 32 bits would be wiser. However, this is a slightly less serious problem than usual because we can rev the ECHConfig version number and the ECH extension if we find something longer is needed.

Yep, that sounds reasonable to me.

draft-ietf-tls-esni.md
```diff
@@ -215,14 +215,19 @@ The ECH configuration is defined by the following `ECHConfig` structure.
     struct {
         HpkeKdfId kdf_id;
         HpkeAeadId aead_id;
-    } ECHCipherSuite;
+    } HpkeSymmetricCipherSuite;
```
cjpatton, line 218: Why this name change?

draft-ietf-tls-esni.md
```diff
@@ -215,14 +215,19 @@ The ECH configuration is defined by the following `ECHConfig` structure.
...
 
     struct {
-        opaque public_name<1..2^16-1>;
-        HpkePublicKey public_key;
+        uint8 key_id;
         HpkeKemId kem_id;
-        ECHCipherSuite cipher_suites<4..2^16-4>;
+        HpkePublicKey public_key;
+        HpkeSymmetricCipherSuite cipher_suites<4..2^16-4>;
```
cjpatton, line 224: IMO it's a bit tidier to have `cipher_suites` immediately precede the `kem_id`.

draft-ietf-tls-esni.md
```diff
@@ -756,12 +766,12 @@ If the client attempts to connect to a server and does not have an ECHConfig
 structure available for the server, it SHOULD send a GREASE {{?RFC8701}}
 "encrypted_client_hello" extension in the first ClientHello as follows:
 
+- Set the `key_id` field to a random byte.
```
cjpatton, line 769: Whether this sticks out depends on how the real `key_id` is chosen. If chosen randomly then we're golden. However the chance of two randomly chosen key ids colliding is high.

draft-ietf-tls-esni.md
```diff
@@ -252,34 +257,44 @@ specification, the contents are an `ECHConfigContents` structure.
...
-public_key
-: The HPKE public key used by the client to encrypt ClientHelloInner.
+extensions
+: A list of extensions that the client must take into consideration when
+generating a ClientHello message. These are described below
+({{config-extensions}}).
+
+The `HpkeKeyConfig` structure contains the following fields:
+
+key_id
```
cjpatton, line 283: Why `key_id` instead of `config_id`? The new name sounds as if it should be possible to have a single key used by multiple configs, but what we actually want is a handle with which the server can resolve which config to use.

cjpatton (commented):
If we're going to use a single byte to identify an ECH config, then we should choose the byte deterministically. But because this may complicate greasing (the grease client needs to know the procedure by which identities are chosen), my preference would be to randomly generate it. But in order to minimize the collision probability, we ought to stick with the current length of 8 bytes.

draft-ietf-tls-esni.md
chris-wood: I don't have a strong feeling here. I think the only important bit is that the kem_id precedes the public key.

draft-ietf-tls-esni.md
chris-wood: It seems to capture the intent of the struct better.

draft-ietf-tls-esni.md
chris-wood: Since the identifier corresponds primarily to the HPKE public key, and not the public name et al. (If we lifted the key_id *out* of the HpkeKeyConfig, I think config_id would be better.)

chris-wood:
> But in order to minimize the collision probability, we ought to stick with the current length of 8 bytes.

This is somewhat of a bike shed, but 8 bytes seems quite large for something that will only have a handful of possible values at any given point in time. Is there a reason to make this longer beyond the performance cost of decryption failing as a result of collisions?

cjpatton:
> This is somewhat of a bike shed, but 8 bytes seems quite large for something that will only have a handful of possible values at any given point in time. Is there a reason to make this longer beyond the performance cost of decryption failing as a result of collisions?

I guess the only reason I can see is that we've settled on q^2 / 2^64 as our threshold for operational risks. In particular, the "acceptance" signal in the SH.random is 8 bytes (https://tools.ietf.org/html/draft-ietf-tls-esni-09#section-10.7). I suppose we could get away with a smaller identifier, since the q is much smaller in this case.

chris-wood:
> I guess the only reason I can see is that we've settled on q^2 / 2^64 as our threshold for operational risks. In particular, the "acceptance" signal in the SH.random is 8 bytes (https://tools.ietf.org/html/draft-ietf-tls-esni-09#section-10.7). I suppose we could get away with a smaller identifier, since the q is much smaller in this case.

This seems different from acceptance signal collision, which actually causes breakage. In contrast, a key_id collision here just means that the server wastes cycles attempting decryption, right?

cjpatton:
> This seems different from acceptance signal collision, which actually causes breakage. In contrast, a key_id collision here just means that the server wastes cycles attempting decryption, right?

A config collision is bad because the server has to figure out which config the client used. It can only do so by trial decryption.

chris-wood:
> A config collision is bad because the server has to figure out which config the client used. It can only do so by trial decryption.

Eh? If there's a collision, the server *knows* which key to try (the one it collided with!), and it will do so, fail to decrypt, and move on as per usual. (There's no trial decryption here, but maybe I'm missing something.)

cjpatton:
Just to put a fine point on this: the spec SHOULD NOT require that implementations support trial decryption. In particular, we ought to ensure that, with high probability, configuration identifiers don't collide. A one-byte, randomly-generated identifier would make collisions inevitable, meaning that implementations will have to support trial decryption.

chris-wood:
> Just to put a fine point on this: the spec SHOULD NOT require that implementations support trial decryption. In particular, we ought to ensure that, with high probability, configuration identifiers don't collide. A one-byte, randomly-generated identifier would make collisions inevitable, meaning that implementations will have to support trial decryption.

Yeah, that's reasonable. If we're going to abide by this, then the ID needs to be at least 8 bytes. I'd be fine with that assuming others are fine it. (@ekr suggested it be longer at the top anyway, so many we should just do it.)

chris-wood:
@cjpatton we're now at 8 bytes :-)

davidben:
> Just to put a fine point on this: the spec SHOULD NOT require that implementations support trial decryption. In particular, we ought to ensure that, with high probability, configuration identifiers don't collide. A one-byte, randomly-generated identifier would make collisions inevitable, meaning that implementations will have to support trial decryption.

I don't think one-byte necessary collides. First, note that the number of configs a server needs to have at a time is relatively small. Past the config's DNS TTL, you can retire it. That means you only need O(dns_ttl / time_between_rotation) configs at once. (From #177, I think the exact value is 1 + ceil(dns_ttl / time_between_rotation). Maybe add a little bit of leeway for people using stale DNS caches, though note we have a recovery flow so you only need to cover most, not all, of stale caches.) This means we don't need to worry about that many values.

Additionally, everything here is server-allocated. Whether you're worried about a birthday bound depends on how you allocate IDs. If it's random, you have a birthday bound, yes. If you cycle through a counter, you get the full range. (Compare how QUIC and DTLS 1.3 only send the low-order bits of the epoch.) If it's random, but you check for duplicates at key issuance, you also get the full range.

The counters or duplicate checks are trickier to issue, of course, though you can implement the trial decryption as a safety net. As of #344, the trial decryption and config_id modes are already unified. Specifically, decryption failure falls back to the outer ClientHello.

draft-ietf-tls-esni.md
cjpatton: Perhaps another way to resolve this would be to change `key_id` to simply `id`. The `key_` prefix seems a bit redundant.

draft-ietf-tls-esni.md
```diff
@@ -740,9 +750,9 @@ operation. Reusing the HPKE context avoids an attack described in
 The client then modifies the "encrypted_client_hello" extension in
 ClientHelloOuter as follows:
 
+- `key_id` is unchanged and contains the client's chosen HPKE key identifier.
```
cjpatton, line 753: Like we did for config_id, we could avoid re-transmitting the key_id by changing its type to a U8-prefixed string. IIRC there were upsides to not re-transmitting config_id.

chris-wood:
I agree with everything you said, but want to drill in on this:

> The counters or duplicate checks are trickier to issue, of course, though you can implement the trial decryption as a safety net. As of #344, the trial decryption and config_id modes are already unified. Specifically, decryption failure falls back to the outer ClientHello.

This seems to be key. I agree that the fallback case is clear and well understood, though one consideration is the case wherein a client attempts GREASE to a server that actually supports ECH. The probability of collision in that scenario is high, and if that scenario happens often, then decryption failure and fallback to the outer CH happens quite regularly. Bumping the length back to eight bytes removes the collision probability for that scenario, I think. 

I'd like to propose we treat the two problems here separately. That is, let's first see if we can land the key ID change (from a client-computed hash to a server-selected ID). And then let's see if we can safely make that ID smaller in size. I don't think we need to address both here. Would that work?

cjpatton (commented):
I believe that `key_id` should be a U8-prefixed string, just like `config_id` was. The spec currently allows for implementations to send an empty `ClientECH.config_id`, e.g., in "local discovery mode" (https://tools.ietf.org/html/draft-ietf-tls-esni-09#section-7.1).

draft-ietf-tls-esni.md
martinthomson: Any change is a change and they are all equally not-free, so minimizing changes is not a goal I would worry about.  In terms of ordering, I prefer what @chris-wood proposed.

martinthomson:
I think that the only thing that pushes toward a larger identifier space for me is the risk of collision with an active key.  I agree with David, that your likely number of keys is ~2 (I might go as high as 4 if the server is in an extended period of offering two types of KEM).  With a one-byte identifier, that's only 1/64 greased sessions hitting the expensive decryption code.

Flexibility here seems appealing, but I'm not sure that this is a place that needs it.  Operationally, this seems very much dependent on centralization, so I'm not buying any story based on loose coordination.  I also note that very small spaces reduce the options servers have for targeted tracking through the use of keys (though it is impossible to eliminate tracking, small spaces make the cost higher to operate tracking at any reasonable scale).

That is why I was motivated to go as low as 1 byte.  If you prefer a much lower probability of having to deal with failed decryption in response to greasing, then 2 bytes gets that down to lower than your likely failed connection rate at 1/16k.

chris-wood:
> With a one-byte identifier, that's only 1/64 greased sessions hitting the expensive decryption code.

Indeed, this is why I propose we separate these issues. Specifically, let's use this PR to land the eight byte key ID, and then make it smaller (as needed) in a followup change. (Personally, I'm fine with one, two, or four bytes here, as I think the decryption failure case is unavoidable.)

@davidben, @martinthomson, @ekr, @cjpatton: what say you? :-)

draft-ietf-tls-esni.md
chris-wood: We could do that, but then ClientECH would either have an inconsistently named field (key_id) or an ambiguously named field (id). I'd prefer to keep key_id.

draft-ietf-tls-esni.md
chris-wood: The upside is that the server doesn't have to check them for equality against HRR, but there's already a check for the cipher_suite field, so this seems like not a stretch.

draft-ietf-tls-esni.md
```diff
@@ -562,8 +562,8 @@ a `ClientECH` with the following values:
 - `payload`, as computed above.
 
 If optional configuration identifiers (see {{optional-configs}})) are used,
-`key_id` MAY be set to a randomly generated byte. Unless specified by the
-application using (D)TLS or externally configured on both sides,
+`key_id` MAY be set to a randomly generated eight bytes. Unless specified by
```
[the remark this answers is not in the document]
chris-wood: That would align this with GREASE, which seems like a fine thing to do. (That is, in both GREASE and optional configs, the ID would be random.)

draft-ietf-tls-esni.md
chris-wood: I reverted this change, too. It's now length prefixed so the HRR case is empty. We can discuss the size of the ID in a separate issue.

draft-ietf-tls-esni.md
```diff
@@ -546,14 +556,14 @@ byte, and the serialized ECHConfig.
...
-- `config_id`, the identifier of the chosen ECHConfig structure;
 - `enc`, as computed above; and
 - `payload`, as computed above.
 
-If optional configuration identifiers (see {{optional-configs}})) are used, the
-`config_id` field MAY be empty or randomly generated. Unless specified by the
-application using (D)TLS or externally configured on both sides,
+If optional configuration identifiers (see {{optional-configs}})) are used,
+`config_id` SHOULD be set to a randomly generated eight bytes. Unless specified
+by the application using (D)TLS or externally configured on both sides,
```
ekr, line 566: So when is empty valid?

draft-ietf-tls-esni.md
```diff
@@ -324,25 +339,20 @@ The payload MUST have the following structure:
...
+       opaque config_id<0..255>
        ECHCipherSuite cipher_suite;
-       opaque config_id<0..255>;
        opaque enc<1..2^16-1>;
        opaque payload<1..2^16-1>;
     } ClientECH;
 ~~~~
 
+config_id
+: The opaque byte identifier matching `ECHConfigContents.key_config.config_id`.
```
ekr, line 350:
```suggestion
: The ECHConfigContents.key_config.config_id for the chosen ECHConfig.
```

ekr (approved):
LGTM but I would split out the technical change from the refactor of the structures.

draft-ietf-tls-esni.md
```diff
@@ -324,25 +339,20 @@ The payload MUST have the following structure:
 
 ~~~~
     struct {
+       opaque config_id<0..255>
        ECHCipherSuite cipher_suite;
-       opaque config_id<0..255>;
```
cjpatton, line 328: Why change the wire format of the ECH extension? We might end up changing the type of `config_id` anyway, in which case the point is moot. However, for now, I think we should keep the extension parameters in the same order.

chris-wood: Reverted!

draft-ietf-tls-esni.md
chris-wood: It's always valid, I think. The server will just have to do trial decryption.

martinthomson: An empty identifier is just one possible value in the space.  It doesn't mean that other configurations aren't identified by non-empty identifiers.

martinthomson (approved):
Happy to have the identifier discussion separately.  It is unfortunately not important or clear enough to lead to easy agreement.  I can live with any outcome.

draft-ietf-tls-esni.md
```diff
@@ -896,8 +906,8 @@ first ClientHelloOuter as follows:
 If the client-facing server accepted ECH, it checks the second ClientHelloOuter
 also contains the "encrypted_client_hello" extension. If not, it MUST abort the
 handshake with a "missing_extension" alert. Otherwise, it checks that
-ClientECH.cipher_suite is unchanged, and that ClientECH.config_id and
-ClientECH.enc are empty. If not, it MUST abort the handshake with an
+ClientECH.cipher_suite and ClientECH.config_id are unchanged, and that
+ClientECH.enc is empty. If not, it MUST abort the handshake with an
```
cjpatton, line 910: I believe this change needs to be reverted as well. From {#client-hrr}:

> The client then modifies the "encrypted_client_hello" extension in ClientHelloOuter as follows:
> 
>   - **config_id is replaced with the empty string.**
>   - cipher_suite is unchanged and contains the client's chosen HPKE cipher suite.
>   - enc is replaced with the empty string.
>   - payload is replaced with the value computed above.

chris-wood: Fixed.

draft-ietf-tls-esni.md
```diff
@@ -324,25 +339,20 @@ The payload MUST have the following structure:
 
 ~~~~
     struct {
-       ECHCipherSuite cipher_suite;
-       opaque config_id<0..255>;
+       HpkeSymmetricCipherSuite cipher_suite;
+       opaque config_id<0..255>
```
cjpatton, line 343: nit: missing semicolon `;`

cjpatton (approved):
One little nit then -> LGTM.

draft-ietf-tls-esni.md
chris-wood:
```suggestion
       opaque config_id<0..255>;
```

davidben:
> Happy to have the identifier discussion separately. It is unfortunately not important or clear enough to lead to easy agreement. I can live with any outcome.

Ditto.

draft-ietf-tls-esni.md
```diff
@@ -215,14 +215,19 @@ The ECH configuration is defined by the following `ECHConfig` structure.
...
-        HpkePublicKey public_key;
+        opaque config_id<0..255>;
         HpkeKemId kem_id;
-        ECHCipherSuite cipher_suites<4..2^16-4>;
+        HpkePublicKey public_key;
+        HpkeSymmetricCipherSuite cipher_suites<4..2^16-4>;
+    } HpkeKeyConfig;
+
+    struct {
+        HpkeKeyConfig key_config;
```
ekr, line 228: It's a little goofy to have the config_id apply to the whole thing here, but I suppose it's inlined.

draft-ietf-tls-esni.md
ekr: What I'm saying here is that we have a way to indicate the identifier is optional, and that's to have it be 8 random bytes. So when would you use empty?

draft-ietf-tls-esni.md
ekr: I think we at least need a comment here to explain that.

chris-wood: Can you suggest text?

draft-ietf-tls-esni.md
chris-wood: Whenever the application wants it, I suppose? We're just recommending that these be random when optional. As Martin says, they don't need to be. Can you propose text if you think something else needs to be said here?

ekr: I think we should forbid the empty identifier.

draft-ietf-tls-esni.md
```diff
@@ -252,34 +257,44 @@ specification, the contents are an `ECHConfigContents` structure.
 
 The `ECHConfigContents` structure contains the following fields:
 
+key_config
+: A `HpkeKeyConfig` structure carrying the configuration information associated
+with the HPKE public key.
```
ekr, line 262:
```suggestion
with the HPKE public key. Note that this structure contains the `config_id` field, which applies to the
entire ECHConfigContents. Sites MUST NOT publish two differet`ECHConfigContents` values with the same
`HpkeKeyConfig` value.
```

draft-ietf-tls-esni.md
chris-wood: I'd like to not do anything here, especially given #379, wherein we move from a vector to a `uint`. Would that be OK?

PR #377, opened 2021-02-02
Remove some more remnants of ESNIKeys

davidben:
The reference to "retry keys" comes from when the field was called
retry_keys and the type ESNIKeys. We've since renamed them to
retry_config and ECHConfig, so the prose should match.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -364,7 +364,7 @@ with the following payload:
 retry_configs
 : An ECHConfigs structure containing one or more ECHConfig structures, in
 decreasing order of preference, to be used by the client in subsequent
-connection attempts.
+connection attempts. These are known as the server's "retry configurations".
 
 This document also defines the "ech_required" alert, which the client MUST send
 when it offered an "encrypted_client_hello" extension that was not accepted by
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -626,21 +626,23 @@ usual, authenticating the connection for the true server name.
...
 {{auth-public-name}}. If authentication or the handshake fails, the client MUST
-return a failure to the calling application. It MUST NOT use the retry keys.
+return a failure to the calling application. It MUST NOT use the retry
+configurations.
 
-Otherwise, when the handshake completes successfully with the public name
-authenticated, the client MUST abort the connection with an "ech_required"
-alert. It then processes the "retry_configs" field from the server's
-"encrypted_client_hello" extension.
+Otherwise, if the both authentication and the handshake complete successfully,
```
cbartle891, line 632: Nit: s/the both/both/

davidben: Huh, that's new. I fixed it in https://github.com/davidben/draft-ietf-tls-esni/tree/retry-config-naming, but GitHub doesn't seem to have noticed. Hopefully it's just some weird slowness on their end and the PR will notice sooner or later?

davidben: Okay, now it's noticed. Done.

chris-wood (approved):

PR #381, opened 2021-02-15
Truncate the config_id to a single byte.

### LINKED ISSUE #379: Consider a shorter config ID length
Assuming #376 lands, we need to decide how long we want this server-chosen identifier to be. There are various operational and privacy considerations at play. For example, shorter identifiers mean that GREASE clients may produce colliding values with higher probability, and the cost of trial decryption on the server may be an overly expensive operational cost for some deployments. Longer identifiers may also mean that using keys as a tracking vector is easier. (But as @ekr has pointed out, ECH does not assume DNS answer integrity, so there are other tracking mechanisms at play.) Let's use this issue to sort out this question. 

cc @davidben, @cjpatton, @martinthomson

chris-wood:
Closes #379.

cc @davidben, @martinthomson, @cjpatton, @cbartle891, @ekr


draft-ietf-tls-esni.md
```diff
@@ -261,7 +261,10 @@ key_config
 : A `HpkeKeyConfig` structure carrying the configuration information associated
 with the HPKE public key. Note that this structure contains the `config_id`
 field, which applies to the entire ECHConfigContents. Sites MUST NOT publish
```
cjpatton, line 263: "MUST NOT" -> "SHOULD NOT"? This sounds like it could be hard to enforce.

cjpatton (commented):
Looks good to me, though I think we should weaken "config_id MUST be unique"  to "config_id SHOULD be unique", since this seems impossible to enforce. (This comment pertains to the previous PR, unfortunately, but we could fix it here.)

draft-ietf-tls-esni.md
chris-wood: Who are you thinking might enforce this? Clients? (I definitely don't feel strongly about the choice of word here.)

cjpatton (approved):

draft-ietf-tls-esni.md
```diff
@@ -261,7 +261,10 @@ key_config
...
-two differet`ECHConfigContents` values with the same `HpkeKeyConfig` value.
+two different `ECHConfigContents` values with the same `HpkeKeyConfig` value.
+The RECOMMENDED technique for choosing `config_id` is to choose a random byte.
+This process should be repeated if this config_id matches that of any existing
+(and valid) ECHConfig.
```
martinthomson, line 267:
```suggestion
This process is repeated if this config_id matches that of any valid ECHConfig, which
could include any ECHConfig that has been recently removed from active use.
```

... or something along those lines.  I suspect that operations will want to avoid collisions over slightly longer timescales.

martinthomson (approved):

cbartle891 (approved):

ekr:
LGTM based on the discussion.

ekr (approved):

PR #383, opened 2021-02-17
Clarify privacy risk pertaining to resumption.

### LINKED ISSUE #380: Related Privacy Leaks suggests too strong of a correlation across resumption
(Filing this as a separate ticket so I don't forget to propose new text.)

#371 added this text:

> Attacks that rely on non-ECH traffic to infer server identity in an ECH connection are 
> out of scope for this document. For example, a client that connects to a particular host 
> prior to ECH deployment may later resume a connection to that same host after ECH 
> deployment, thereby linking the resulting ECH connection to the original non-ECH 
> connection.

I think this statement is a bit strong and overstates the issue. It reads as if resuming connections across ECH and non-ECH links the two connections.

If we're talking about someone other than the server, this isn't true. Tickets are issued encrypted and, in ECH, even sent encrypted. Even if they were sent in the clear, TLS 1.3's [single-use tickets](https://tools.ietf.org/html/rfc8446#appendix-C.4) protects against such correlations. Rather, the correlation in #369 is a much weaker one: based on either the size of the ClientHello or whether the server accepted the ticket, you learn a single bit for whether it was *some* previous connection made by that client. You don't know which previous connection it was, and you don't know if it was an ECH connection or a non-ECH connection. In fact what you learn... more-or-less aligns with the anonymity set. See https://github.com/tlswg/draft-ietf-tls-esni/issues/369#issuecomment-736083162.

(If we're talking about the server, yes, resuming allows the server to link the two connections. That's fundamental to resumption and is in the purview of https://github.com/tlswg/tls13-spec/pull/1205. I don't think most applications would consider ECH vs non-ECH a correlation boundary w.r.t. the server itself.)

cbartle891:
Resolves #380.

cc @davidben

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -1193,8 +1193,9 @@ validation.
 Attacks that rely on non-ECH traffic to infer server identity in an ECH
 connection are out of scope for this document. For example, a client that
 connects to a particular host prior to ECH deployment may later resume a
-connection to that same host after ECH deployment, thereby linking the resulting
-ECH connection to the original non-ECH connection.
+connection to that same host after ECH deployment. An adversary that observes
+this can deduce that the ECH-enabled connection was made to a host that the
+client previously connected to and which is within the same anonymity set.
 
 ## Attacks Exploiting Acceptance Confirmation
 


chris-wood (approved):
LGTM -- thanks!

davidben (approved):
Thanks!

PR #390, opened 2021-02-24
Update HPKE (and ECH) versions.

### LINKED ISSUE #387: Update to HPKE-08
This new version of the draft removes the version-vector dependency, so this should be the last time we update the pinned version.

chris-wood:
Closes #387, and also updates the ECH version. I didn't bump the extension codepoint for "ech_is_inner," since that seems harmless. 

cc @davidben, @cjpatton


cbartle891 (approved):

cjpatton (approved):
Works for me!

ekr (approved):

PR #391, opened 2021-02-26
s/ECHConfigs/ECHConfigList, for clarity.

chris-wood:
See [the source of confusion](https://github.com/MikeBishop/dns-alt-svc/issues/298).

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -850,8 +850,8 @@ If the client offers the "ech_is_inner" extension ({{is-inner}})
 in addition to the "encrypted_client_hello" extension, the server MUST abort
 with an "illegal_parameter" alert.
 
-First, the server collects a candidate ECHConfigList. This list is determined
-by one of the two following methods:
+First, the server collects a set of candidate ECHConfig values. This list is
+determined by one of the two following methods:
 
 1. Compare ClientECH.config_id against identifiers of each known ECHConfig
    and select the ones that match, if any, as candidates.
@@ -864,8 +864,8 @@ the second method should be used for matching ClientECH to known ECHConfig. See
 {{optional-configs}}. Unless specified by the application using (D)TLS or
 externally configured on both sides, implementations MUST use the first method.
 
-The server then iterates over the candidate ECHConfigList, attempting to decrypt
-the "encrypted_client_hello" extension:
+The server then iterates over the candidate ECHConfig values, attempting to
+decrypt the "encrypted_client_hello" extension:
 
 The server verifies that the ECHConfig supports the cipher suite indicated by
 the ClientECH.cipher_suite and that the version of ECH indicated by the client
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -850,21 +850,21 @@ If the client offers the "ech_is_inner" extension ({{is-inner}})
 in addition to the "encrypted_client_hello" extension, the server MUST abort
 with an "illegal_parameter" alert.
 
-First, the server collects a set of candidate ECHConfigs. This set is
```
dmcardle, line 853: Would it be possible to keep the "set of candidate $X" phrasing?

The phrase "candidate ECHConfigList" sounds to me like the whole ECHConfigList is a candidate, not that each element is a candidate. 

Perhaps

> First, the server collects a set of candidate ECHConfig values.

draft-ietf-tls-esni.md
```diff
@@ -850,21 +850,21 @@ If the client offers the "ech_is_inner" extension ({{is-inner}})
...
-determined by one of the two following methods:
+First, the server collects a candidate ECHConfigList. This list is determined
```
davidben, line 853: I think this section should just be "ECHConfigs" or "ECHConfig values". The server doesn't actually need to serialize them or anything.

Also now that we also have a better word, maybe {{real-ech}} should now be a bit more precise and say something like:
"To offer ECH, the client first chooses a suitable ECHConfig from the published ECHConfigList."

draft-ietf-tls-esni.md
chris-wood: Sure, that works equally well for me!

draft-ietf-tls-esni.md
chris-wood: (See latest change)

dmcardle (approved):
Thanks, Chris!

davidben (approved):
LGTM with one comment that got lost in the mixup. Now that we also have a better word, maybe {{real-ech}} should now be a bit more precise and say something like:

> To offer ECH, the client first chooses a suitable ECHConfig from the published ECHConfigList.

I figure the precision might help reenforce the path from published structure to client behavior.

chris-wood:
> I figure the precision might help reenforce the path from published structure to client behavior.

I left this out because I figure there may be cases where there is no "published" ECHConfigList. (Maybe the config is baked in the app?)

davidben:
Ah fair, that might be too precise. I think it still makes sense for that case to have the baked-in thing be an ECHConfigList than an ECHConfig, just for consistency. (I'm envisioning that ECHConfigList is the "public API" for client ECH and your TLS library might just that one byte string as input.)

How about s/published/server's/?

chris-wood:
> How about s/published/server's/?

That'd work :)

chris-wood:
@davidben done!

cbartle891 (approved):

davidben (approved):
LGTM

PR #409, opened 2021-03-30
Compress ECHConfigContents.

### LINKED ISSUE #403: public_name and maximum_name_length could be smaller
I think that we're committed now to public_name being a DNS name (there is no way to indicate the type of name (like in RFC 6066).  So I have to assume that the only type of name that is permitted is a DNS name.  And a DNS name can't be longer the 255 bytes.

Therefore, ECHConfig could be:

```tls
    struct {
        HpkeKeyConfig key_config;
        uint8 maximum_name_length;
        opaque public_name<1..255>;
        Extension extensions<0..2^16-1>;
    } ECHConfigContents;
```

chris-wood:
Closes #403.

cc @martinthomson

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -226,8 +226,8 @@ The ECH configuration is defined by the following `ECHConfig` structure.
 
     struct {
         HpkeKeyConfig key_config;
-        uint16 maximum_name_length;
-        opaque public_name<1..2^16-1>;
+        uint8 maximum_name_length;
+        opaque public_name<1..255>;
         Extension extensions<0..2^16-1>;
     } ECHConfigContents;
 


draft-ietf-tls-esni.md
```diff
@@ -226,8 +226,8 @@ The ECH configuration is defined by the following `ECHConfig` structure.
 
     struct {
         HpkeKeyConfig key_config;
-        uint16 maximum_name_length;
-        opaque public_name<1..2^16-1>;
+        uint8 maximum_name_length;
+        opaque public_name<1..255>;
```
davidben, line 230: (2^16-1 in `public_name` is probably my fault. I bet I got it from HostName in RFC6066.)

davidben (approved):

draft-ietf-tls-esni.md
chris-wood: No one's fault :-)

cbartle891 (approved):

cjpatton (approved):

martinthomson (approved):
Easy.  Thanks.

Good to see at least some of the problems of RFC 6066 aren't going to continue here.

ekr (approved):
LGTM

PR #410, opened 2021-03-30
Authenticate ClientECH extension position in ClientHelloOuterAAD.

### LINKED ISSUE #402: Not all of CHOuter is authenticated
The position of the ECH extension is not authenticated as part of ClientHelloOuterAAD.  This means that it is possible to construct multiple different ClientHello messages that will all be accepted.

This probably isn't that interesting, but it might be worth noting.

chris-wood:
Closes #402.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -466,10 +466,11 @@ The first three parameters are equal to, respectively, the
 `ClientECH.cipher_suite`, `ClientECH.config_id`, and `ClientECH.enc` fields of
 the payload of the "encrypted_client_hello" extension. The last parameter,
 `outer_hello`, is computed by serializing ClientHelloOuter with the
-"encrypted_client_hello" extension removed. Note this does not include the
-four-byte header included in the Handshake structure.
+"encrypted_client_hello" extension set to the empty string, i.e., the
+`extension_data` list has zero length. Note this serialization does not include
+the four-byte header included in the Handshake structure.
 
-Note the decompression process in {{encoding-inner}} forbids
+The decompression process in {{encoding-inner}} forbids
 "encrypted_client_hello" in OuterExtensions. This ensures the unauthenticated
 portion of ClientHelloOuter is not incorporated into ClientHelloInner.
 


draft-ietf-tls-esni.md
```diff
@@ -466,10 +466,11 @@ The first three parameters are equal to, respectively, the
 `ClientECH.cipher_suite`, `ClientECH.config_id`, and `ClientECH.enc` fields of
 the payload of the "encrypted_client_hello" extension. The last parameter,
 `outer_hello`, is computed by serializing ClientHelloOuter with the
-"encrypted_client_hello" extension removed. Note this does not include the
-four-byte header included in the Handshake structure.
+"encrypted_client_hello" extension set to the empty string, i.e., the
+`extension_data` list has zero length. Note this serialization does not include
+the four-byte header included in the Handshake structure.
```
davidben, line 471: If we're preserving the location of the "encrypted_client_hello" extension anyway, do we still need the `cipher_suite`, `config_id`, and `enc` separated out like this? Seems we can drop all of those, make ClientHelloOuterAAD the reserialized ClientHello, and say:

> [...] serializing ClientHelloOuter with the "payload" field of the "encrypted_client_hello" extension set to the empty string.

chris-wood: Hmm, that's what this text already says ("extension_data" is the payload of the extension).

davidben: Sorry, this was probably unclear. GitHub is terrible and doesn't let you add comments more than two lines away from the diff. :-/ ClientHelloOuterAAD is defined as:

~~~
   struct {
      HpkeSymmetricCipherSuite cipher_suite;
      uint8 config_id;
      opaque enc<1..2^16-1>;
      opaque outer_hello<1..2^24-1>;
   } ClientHelloOuterAAD;
~~~

The encrypted_client_hello extension normally contains a:

~~~
    struct {
       HpkeSymmetricCipherSuite cipher_suite;
       uint8 config_id;
       opaque enc<1..2^16-1>;
       opaque payload<1..2^16-1>;
    } ClientECH;
~~~

I'm saying that, rather than replace encrypted_client_hello with the empty string, and lift all but one of its fields to the outside, we may as well say ClientHelloOuterAAD is just the outer_hello field, and that encrypted_client_hello is a *non-empty* extension, containing a ClientECH whose payload field is empty.

chris-wood: I had that change first, but it seems unnecessarily complicated. Why not just zero out the extension entirely? All we care about is the position.

davidben: Heh, I figured zeroing out the field was simpler. Then we don't have to worry copying out the correct `enc` field (in the second ClientHelloOuterAAD, is it the empty string or the old value?), and ClientHello serializers don't need to support two different kinds of encrypted_client_hello payloads (empty vs ClientECH).

chris-wood: I totally do not feel strongly here. They are mostly equivalent in my view, so whatever folks think is best works for me!

cjpatton (approved):

martinthomson:
Having looked at this, I have a preference for a construction where the outer ClientHello has a placeholder that is empty, rather than a placeholder that has an empty payload.

Zeroing out the payload field is not easier in our implementation.  It is somewhat more difficult to implement.

This is a preference only; anything can be implemented.  But this version is quite a bit more code in our design.

draft-ietf-tls-esni.md
ekr: I think I'm in favor of this PR as-is but I don't feel strongly.

PR #411, opened 2021-03-30
Validate the contents of ClientHelloInner.

### LINKED ISSUE #406: Server enforcement of ClientHelloInner invariants
We currently have a number of invariants on the ClientHelloInner that the client must follow:

* ClientHelloInner must include ech_is_inner
* ClientHelloInner may not offer TLS 1.2 and below
* ClientHelloInner may not include encrypted_client_hello (somewhat implicit from the first rule and section 7)

But, syntactically, the client could violate this. How is the server expected to respond? In particular, a shared mode server directly observes all invariants, and it seems a little weird to ask it to send something it knows is inconsistent. But if we say to enforce, we either have a shared/split difference, or we put all the enforcement at the client-facing server (not wrong...) because the backend server doesn't have enough information to check these.

* If ClientHelloInner lacks ech_is_inner, should the server just accept ECH without reporting the confirmation signal, or should it report an error?
* If we end up negotiating TLS 1.2 with ClientHelloInner, is that an error or do we just proceed, without reporting the confirmation signal and sending a cleartext certificate?
* If ClientHelloInner contains another encrypted_client_hello extension, one possible implementation strategy would report retry keys, since backend server halves that see encrypted_client_hello typically assume the extension wasn't decrypted. Is that correct, or do you omit retry keys?

@dmcardle

chris-wood:
Closes #406.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -899,14 +899,15 @@ Upon determining the ClientHelloInner, the client-facing server then checks
 that the message includes the "ech_is_inner" extension, omits the
 "encrypted_client_hello" extension, and does not offer TLS 1.2 or below
 versions. If any of these checks fails, the client-facing server MUST
-abort with an "illegal_parameter" alert. If these checks succeed, the
-client-facing server then forwards the ClientHelloInner to the appropriate
-backend server, which proceeds as in {{backend-server}}. If the backend server
-responds with a HelloRetryRequest, the client-facing server forwards it,
-decrypts the client's second ClientHelloOuter using the procedure in
-{{server-hrr}}, and forwards the resulting second ClientHelloInner. The
-client-facing server forwards all other TLS messages between the client and
-backend server unmodified.
+abort with an "illegal_parameter" alert.
+
+If these checks succeed, the client-facing server then forwards the
+ClientHelloInner to the appropriate backend server, which proceeds as in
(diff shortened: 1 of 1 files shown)


cbartle891 (approved):

davidben (approved):

draft-ietf-tls-esni.md
```diff
@@ -895,13 +895,18 @@ Otherwise, the server reconstructs ClientHelloInner from
...
+"encrypted_client_hello" extension, and does not offer TLS 1.2 or below
+versions. If any of these checks fails, the client-facing server MUST
+abort with an "illegal_parameter" alert. If these checks succeed, the
+client-facing server then forwards the ClientHelloInner to the appropriate
+backend server, which proceeds as in {{backend-server}}. If the backend server
+responds with a HelloRetryRequest, the client-facing server forwards it,
+decrypts the client's second ClientHelloOuter using the procedure in
+{{server-hrr}}, and forwards the resulting second ClientHelloInner. The
+client-facing server forwards all other TLS messages between the client and
+backend server unmodified.
```
martinthomson, line 909: This is a [wall of text](https://en.uncyclopedia.co/wiki/Wall_of_Text) now.

It might be time to break this down a little more.  Pause for a breath occasionally.

cjpatton (approved):

PR #415, opened 2021-03-31
Mirror early_data in ClientHelloOuter when applicable.

### LINKED ISSUE #408: Early data and ClientHelloOuter interaction
Here's a fun case. Suppose the client talks to a 0-RTT-capable, ECH-capable server. It'll send early_data + pre_shared_key in ClientHelloInner, wrap that in _some_ ClientHelloOuter, and follow it up with early data records.

Now the server rejects ECH and handshakes with the ClientHelloOuter. But the client has already sent early data, so the server needs to know to skip past it. I believe this means we need to require the client mirror ClientHelloInner.early_data in ClientHelloOuter, even though it's not actually offering to resume anything. Sending early_data without pre_shared_key is slightly weird (if tolerated at all), so we might even want to recommend a fake outer pre_shared_key.

We _almost_ don't care about this: the client may as well stop the handshake at server Finished and skip the client Finished flight anyway, for purposes of the recovery flow. However:

* What's not what the draft currently says
* If the server rejects ECH, handshakes with ClientHelloOuter, *and* sends HelloRetryRequest, we do care about the client's second flight getting through.

That second case is extra fun because the client-facing server must look for the second ClientHello too.

chris-wood:
This should probably land after #414.

Closes #408.

cc @davidben

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -533,6 +533,10 @@ it does a standard ClientHello, with the exception of the following rules:
    "server_name" extension.
 1. It MUST NOT include the "pre_shared_key" extension. (See
    {{flow-clienthello-malleability}}.)
+1. When the client offers the "early_data" extension in ClientHelloInner, it
+   MUST also include the "early_data" extension in ClientHelloOuter. This
+   allows servers that reject ECH and use ClientHelloOuter to safely ignore any
+   early data sent by the client per {{RFC8446}}, Section 4.2.10.
 
 [[OPEN ISSUE: We currently require HRR-sensitive parameters to match in
 ClientHelloInner and ClientHelloOuter in order to simplify client-side
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -533,7 +533,7 @@ it does a standard ClientHello, with the exception of the following rules:
 1. The value of `ECHConfig.contents.public_name` MUST be placed in the
    "server_name" extension.
 1. When the client offers the "pre_shared_key" extension in ClientHelloInner, it
-   MUST also include a GREASE "pre_shared_key" extension in ClientHelloOuter,
+   SHOULD also include a GREASE "pre_shared_key" extension in ClientHelloOuter,
    generated in the manner described in {{grease-psk}}. The client MUST NOT use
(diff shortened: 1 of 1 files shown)


davidben (approved):

cjpatton (approved):

draft-ietf-tls-esni.md
```diff
@@ -533,6 +533,10 @@ it does a standard ClientHello, with the exception of the following rules:
    "server_name" extension.
 1. It MUST NOT include the "pre_shared_key" extension. (See
    {{flow-clienthello-malleability}}.)
+1. When the client offers the "early_data" extension in ClientHelloInner, it
+   MUST also include the "early_data" extension in ClientHelloOuter. This
+   allows servers that reject ECH and use ClientHelloOuter to safely ignore any
+   early data sent by the client per {{RFC8446}}, Section 4.2.10.
```
ekr, line 539: This also requires pre_shared_key per s 4.2.10, so this contradicts the previous reqt.

ekr (changes requested):

draft-ietf-tls-esni.md
chris-wood: This is based on #414, which clarifies that the PSK must also be present. This won't land until that does.

draft-ietf-tls-esni.md
```diff
@@ -533,11 +533,16 @@ it does a standard ClientHello, with the exception of the following rules:
 1. The value of `ECHConfig.contents.public_name` MUST be placed in the
    "server_name" extension.
 1. When the client offers the "pre_shared_key" extension in ClientHelloInner, it
-   SHOULD also include a GREASE "pre_shared_key" extension in ClientHelloOuter,
+   MUST also include a GREASE "pre_shared_key" extension in ClientHelloOuter,
```
chris-wood, line 536:
```suggestion
   SHOULD also include a GREASE "pre_shared_key" extension in ClientHelloOuter,
```

draft-ietf-tls-esni.md
chris-wood: @ekr this should now be resolved give that #414 landed. Can you please take another look?

cbartle891 (approved):

PR #416, opened 2021-04-06
Don't make a new handshake

martinthomson:
The idea that it might be OK to make a new handshake based on
unauthenticated messages is inherently insecure.


martinthomson:
This won't be needed if you take #417.  I just provided it because it is separable.

chris-wood:
This is overcome by #423. Closing.

PR #417, opened 2021-04-06
HRR applies to both

### LINKED ISSUE #233: Trial decryption after HelloRetryRequest
In a normal ECHO handshake with HelloRetryRequest there are 4 ClientHellos sent: ClientHelloOuter1, ClientHelloInner1, ClientHelloOuter2, ClientHelloInner2.

Encryption of ClientHelloInner2 is bound to echo_hrr_key so the server must have the ECHOConfig used for ClientHelloInner1 in order to process ECHO the second time.
And by the time the client receives encrypted handshake traffic it is not useful to signal if ClientHelloInner1 was successfully decrypted, what matters is to know if ECHO is globally successful or not.

So one aspect that I find not explained enough is whether the server replying with HRR should use ClientHelloInner1 in the transcript, or instead always takes ClientHelloOuter1 (data on the wire) regardless of the first decryption status. The ECHO status accept/reject can be conveyed based on distinction ClientHello2 inner/outer only.

This point impacts the number of transcripts the client may have to try, i.e. can be sure the combination (ClientHelloInner1, ClientHelloOuter2) is never valid.

### LINKED ISSUE #373: CHI vs. CHO when HRR is sent
While attempting to draft a change for https://github.com/tlswg/draft-ietf-tls-esni/issues/359, I was confused by this part of section 6.1.4:

"If the server sends a HelloRetryRequest in response to the ClientHello, the client sends a second updated ClientHello per the rules in [RFC8446]. However, at this point, the client does not know whether the server processed ClientHelloOuter or ClientHelloInner"

When does the client learn whether the server processed ClientHelloOuter or ClientHelloInner? If the client sees that the last 8 bytes of ServerHello.random were set to the first 8 bytes of `accept_confirmation`, does it just assume that the first ClientHello that the server processed was a ClientHelloInner? That seems like a big assumption. Couldn't the following happen?:

```
Client                           Client-facing server             Backend server
              CHO(CHI)
           ------------->
                                   (processes CHO)
                HRR
           <-------------
              CHO(CHI)
           ------------->
                                                          CHI
                                                      ----------->
                                                                  (processes CHI)
                            SH
          <-----------------------------------------------------

```

If this can happen, the client will assume, after it receives the SH, that the server processed the first ClientHelloInner, but this is in fact not the case, and the transcript will break. I mean, I guess that's okay if you don't want this behavior to succeed, but AFAICT, the draft doesn't explicitly say that the client-facing server shouldn't do this. Should it?

(Sorry if this is in fact already resolved somewhere in the draft. Also I'm pretty sure it's a different issue from https://github.com/tlswg/draft-ietf-tls-esni/issues/333.)

martinthomson:
The current document is *so* close to getting this right.  I think that
this is correct and sufficient.

Closes #233.
Closes #373 by doing nothing.
Issue #358 can be fixed in TLS 1.3 proper.
Issue #333 can be addressed separately.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -723,38 +723,41 @@ error code.
 
 ### Handling HelloRetryRequest {#client-hrr}
 
-As required in {{real-ech}}, clients offering ECH MUST ensure that all
-extensions or parameters that might change in response to receiving a
-HelloRetryRequest have the same values in ClientHelloInner and
-ClientHelloOuter. That is, if a HelloRetryRequest causes a parameter to be
-changed, the same change is applied to both ClientHelloInner and
-ClientHelloOuter. Applicable parameters include:
-
-1. TLS 1.3 {{!RFC8446}} ciphersuites in the ClientHello.cipher_suites list.
-1. The "key_share" and "supported_groups" extensions {{RFC8446}}. (These
-extensions may be copied from ClientHelloOuter into ClientHelloInner as
-described in {{real-ech}}.)
-1. Versions in the "supported_versions" extension, excluding TLS 1.2 and
-earlier. Note the ClientHelloOuter MAY include these older versions, while the
-ClientHelloInner MUST omit them.
-
-Future extensions that might change across first and second ClientHello messages
-in response to a HelloRetryRequest MUST have the same value.
-
 If the server sends a HelloRetryRequest in response to the ClientHello, the
-client sends a second updated ClientHello per the rules in {{RFC8446}}.
-However, at this point, the client does not know whether the server processed
-ClientHelloOuter or ClientHelloInner, and MUST regenerate both values to be
-acceptable. Note: if ClientHelloOuter and ClientHelloInner use different groups
-for their key shares or differ in some other way, then the HelloRetryRequest
-may actually be invalid for one or the other ClientHello, in which case a fresh
-ClientHello MUST be generated, ignoring the instructions in HelloRetryRequest.
-Otherwise, the usual rules for HelloRetryRequest processing apply.
+client updates both ClientHelloInner and ClientHelloOuter.  At this point, the
+client does not know whether the server processed ClientHelloOuter or
+ClientHelloInner, so the changes required by the HelloRetryRequest are applied
+to both.
 
-The client encodes the second ClientHelloInner as in {{encoding-inner}}, using
-the second ClientHelloOuter for any referenced extensions. It then encrypts
-the new EncodedClientHelloInner value as a second message with the previous
-HPKE context:
+To ensure that a HelloRetryRequest is successful, clients needs to ensure that
+any values that might change in response to HelloRetryRequest are consistent in
+the two messages.  Of the values defined in {{RFC8446}}, a client needs to
+include consistent values for ClientHello.cipher_suites, the value of
+"supported_versions" and "supported_groups" extensions, and which groups are
+included in the "key_share" extension.
+
+By consistent, this means that values in ClientHelloInner are identical to those
+in ClientHelloOuter, but with any values that only apply to TLS 1.2 or lower
+removed. ClientHelloOuter might be used to negotiate TLS 1.2, but
+ClientHelloInner cannot.  A client MAY include values in ClientHelloOuter that
+are only applicable to TLS 1.2 or lower versions, but any such options MUST NOT
+be included in ClientHelloInner.
+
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -723,46 +723,49 @@ error code.
...
+client does not know whether the server processed ClientHelloOuter or
+ClientHelloInner, so the changes required by the HelloRetryRequest are applied
+to both.
+
+To ensure that a HelloRetryRequest is successful, clients needs to ensure that
+any values that might change in response to HelloRetryRequest are consistent in
+the two messages.  Of the values defined in {{RFC8446}}, a client needs to
+include consistent values for ClientHello.cipher_suites, the value of
+"supported_versions" and "supported_groups" extensions, and which groups are
+included in the "key_share" extension.
```
davidben, line 737: This criteria still doesn't look right. It seems to make the same mistake as text elsewhere in the document. ClientHello.cipher_suites cannot change in response to HelloRetryRequest, so it does not match the stated criteria. But in order for HRR to successfully apply to both, it must match. The PR also says 'which groups are
included in the "key_share" extension', which is not supported by the proposed sentence. The proposed sentence would capture the entire key_share extension.

For me, I think the big question on matching vs. making a mismatch work (I now think we can do something much simpler than the HRRInner/HRROuter design, especially since we're discarding #333) is getting this criteria down. Whether we can write down clear and complete criteria that captures current and future[*] extensions and convince ourselves it'll work going forward. I don't think this PR is yet a successful existence proof.

[*] We don't have many of these, but RFC8492 is at least one data point.

martinthomson: It seems like you are fixating on being able to find a perfect solution (as @cjpatton is).  I'm suggesting that we give up on that.  We have the means to work out what the right answer is for any new extension.  And the default answer (it can't change) means that we won't often need to contend with that problem.  It's only when new key exchange methods arise that this becomes an issue.  Which is why RFC 8492 is potentially interesting, except that it really isn't[*].

[*] RFC 8492 is badly specified; it doesn't say that ClientHello needs to include password_salt in [Section 4.5.2.2](https://datatracker.ietf.org/doc/html/rfc8492#section-4.5.2.2).  But it seems fairly clear in its intent.  The extension exists to communicate a salt, which leads to a revision to the value of the key_share value.  This doesn't really change the calculus, except that key_share handling gets more complicated.  None of that is visible at the level at which this text operates though.

I think that I know where your objection comes from, but it's a fine point regarding wording rather than a point about its intent.  I can maybe see the shortcomings of the text, but I want to work through the fundamental point before I try to finagle the text.

davidben: You're right that this is mostly a comment on the text. I think you and I are on the same page on the PR's intent (arrange for an identical set of acceptable HRRs between CHInner and CHOuter). But I think the text doesn't accomplish the intent. I'm using the complexity and accuracy of the text somewhat as a proxy for how future-proof it is, though there's also the more immediate concern that the paragraph is self-contradictory. :-)

When you say figure out the right answer for any new extension, do you mean that each new extension where this matters should have some text relating to ECH? I think that's fine if we provide enough guidance (even if non-normative) that new extensions know what they need to do. (Do we need to update 8446 for this?) I don't think the general guidance in the PR works:

> To ensure that a HelloRetryRequest is successful, clients needs to ensure that any values that might change in response to HelloRetryRequest are consistent in the two messages. 

> By consistent, this means that values in ClientHelloInner are identical to those in ClientHelloOuter, but with any values that only apply to TLS 1.2 or lower removed.

>  Only extensions that change in response to HelloRetryRequest need to have any consideration for consistency between ClientHelloOuter and ClientHelloInner.

This general guidance isn't consistent with the goal of having an identical acceptable-HRR-set between the two. Suppose we added a parallel version of supported_groups + key_share tomorrow. By this guidance, supported_groups2 does not need to think about ECH, while key_share2 needs to be identical. But this is wrong on both counts. supported_groups2 *does* need to think about ECH, and only a subset of the information in key_share2 needs to match. (The latter isn't fatal. We could conservatively say key_share2 need to completely match, but I'll note that's not the strategy taken by this PR.)

I believe the actual general rule is if the extension impacts what HRR messages are acceptable.

chris-wood:
Closing as a solution to HRR since #423 was merged. (If we want to add text that suggests how clients construct CH1Inner/CH1Outer, that would be fine, but let's do it in a different PR!)

PR #419, opened 2021-04-06
Grease and HRR leak

### LINKED ISSUE #358: ECH contradicts second ClientHello consistency requirements in RFC8446

martinthomson:
But that is just how it is.  We get to choose between that leakage and a
known large amount of bustage.

Closes #358.


cbartle891 (approved):

cjpatton:
This solution is that it admits a straightforward, active distinguisher: just intercept CH1, respond with a fake HRR, and look at CH2. It's true that we "give up" on making HRR not stick out in #407. However, one of the design goals there is to prevent active attacks that trigger the HRR code path when it otherwise would not have happened.

cbartle891:
> This solution admits a straightforward, active distinguisher: just intercept CH1, respond with a fake HRR, and look at CH2. It's true that we "give up" on making HRR not stick out in #407. However, one of the design goals there is to prevent active attacks that trigger the HRR code path when it otherwise would not have happened. What we give up on is ensuring "don't stick out" when HRR is inevitable for the given client and server.

Ah, are you saying that this conflicts with #407 (and thus would have to be removed if #407 is merged)?

cjpatton:
I believe #407 resolves the same "OPEN ISSUE", yes.

chris-wood (approved):

draft-ietf-tls-esni.md
```diff
@@ -802,12 +802,13 @@ structure available for the server, it SHOULD send a GREASE {{?RFC8701}}
...
-
-[[OPEN ISSUE: The above doesn't match HRR handling for either ECH acceptance or
-rejection. See issue https://github.com/tlswg/draft-ietf-tls-esni/issues/358.]]
+When sending a second ClientHello in response to a HelloRetryRequest, the client
+copies the entire "encrypted_client_hello" extension from the first ClientHello.
+The identical value will reveal to an observer that the value of
+"encrypted_client_hello" was fake, but this only occurs if there is a
+HelloRetryRequest.  A client MAY generate a new value, however updating the
+value of the extension is known to cause compatibility issues with some deployed
+implementations.
```
chris-wood, line 811:
```suggestion
HelloRetryRequest. 
```

PR #420, opened 2021-04-14
Relax computation of the acceptance signal.

### LINKED ISSUE #399: The serverhello.random trick is ugly and might interact with ticket handling...
Encoding the "ECH worked" signal in the SH.random lower octets is, at best, ugly. It also arguably causes violation of internal API designs  in at least the OpenSSL code. And I'm not sure if there might be some interaction between that and ticket handling (just from the look of some code I've yet to understand really.)

Do we really need such a hacky solution?  Given the ECH sticks out in the CH, we could just define a non-encrypted extension in the SH to include the confirmation value maybe (though even that is pretty hacky).

chris-wood:
We previously derived the acceptance signal from the handshake secret.
This meant that clients which used the wrong ECHConfig might need to
process ServerHello extensions twice before computing the signal, which
can be problematic for some libraries. Given that the signal's secrecy
is entirely dependent on ClientHelloInner.random, we can relax the signal
computation and base it on the transcript alone, which includes
ClientHelloInner.random, rather than a secret derived from that transcript.

Closes #399.

cc @davidben, @cjpatton, @sftcd

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -536,8 +536,9 @@ it does a standard ClientHello, with the exception of the following rules:
    SHOULD also include a GREASE "pre_shared_key" extension in ClientHelloOuter,
    generated in the manner described in {{grease-psk}}. The client MUST NOT use
    this extension to advertise a PSK to the client-facing server. (See
-   {{flow-clienthello-malleability}}.) When the client includes a GREASE "pre_shared_key" extension, 
-   it MUST also copy the "psk_key_exchange_modes" from the ClientHelloInner into the ClientHelloOuter.
+   {{flow-clienthello-malleability}}.) When the client includes a GREASE
+   "pre_shared_key" extension, it MUST also copy the "psk_key_exchange_modes"
+   from the ClientHelloInner into the ClientHelloOuter.
 
 [[OPEN ISSUE: We currently require HRR-sensitive parameters to match in
 ClientHelloInner and ClientHelloOuter in order to simplify client-side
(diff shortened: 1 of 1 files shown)


davidben (approved):

sftcd:
On 14/04/2021 21:11, David Benjamin wrote:
> @davidben approved this pull request.

Yep, good change. Thanks,
S.

> 
> 
> 
> 
>

cjpatton (approved):

cbartle891 (approved):

draft-ietf-tls-esni.md
```diff
@@ -1003,21 +1004,22 @@ client by computing its ServerHello as described here.
...
+      8)
 ~~~
 
-where Derive-Secret and Handshake Secret are as specified in {{RFC8446}},
-Section 7.1, and ClientHelloInner...ServerHelloECHConf refers to the sequence of
-handshake messages beginning with the first ClientHello and ending with
+where HKDF-Expand-Label and Transcript-Hash are as defined in {{RFC8446}},
+Section 7.1, "0" indicates a string of Hash.length bytes set to zero, and
+ClientHelloInner...ServerHelloECHConf refers to the sequence of handshake
+messages beginning with the first ClientHello and ending with
```
martinthomson, line 1019: Not a problem with this PR, but...

I honestly don't understand this construction.  ServerHelloECHConf isn't part of the sequence listed in [Section 4.4.1 of RFC 8446](https://tools.ietf.org/html/rfc8446#section-4.4.1).  Maybe I can infer that it replaces ServerHello, but you need to *say* that.  And ClientHelloInner isn't listed there either.

I think that you need to be more careful about this part of the specification.

martinthomson (approved):
Seems fine to me.

draft-ietf-tls-esni.md
chris-wood: Yeah, that's the idea. I updated the text to make this explicit.

Issue #396, opened 2021-03-10
A MUST for inclusion of public_name in outer SNI seems wrong


sftcd:
For at least command line clients (e.g. curl) and perhaps for censorship circumvention tools it makes sense to allow the outer SNI to contain nothing or something that is not the public_name from the SVCB RR. It's possible that any tool that acquires an ECHConfig from a non-DNS source would benefit from that option too.

The offending (for me:-) text is in section 6.1, 2nd list, bullet 7 where it says: 'The value of ECHConfig.contents.public_name MUST be placed in the "server_name" extension.'

There may be clients that do need the above. I'm not sure how to chacterise those in a short phrase, but that's likely possible. At the IETF-110 TLS session mention was made of clients that need to fallback but I'm not sure what reference would be correct for that if e.g. we appended something like "....for clients that need to be able to fallback. [ref]"

cbartle891:
> and perhaps for censorship circumvention tools 

Censorship circumvention isn't one of the stated goals of ECH.

> It's possible that any tool that acquires an ECHConfig from a non-DNS source would benefit from that option too.

Even if an ECHConfig is acquired from a non-DNS source, it still has the ECHConfig.contents.public_name field, which, as specified in 6.1.3.3, is necessary because:

"The client MUST verify that the certificate is valid for ECHConfig.contents.public_name. If invalid, it MUST abort the connection with the appropriate alert."

chris-wood:
+1 -- for robustness reasons, I believe we should keep this as-is.

sayrer:
I also found this requirement odd. Is it really necessary with the changes to config_id in draft -10? If someone could explain why this should be mandatory, that would be fine. I just don't understand the requirement.

davidben:
When there is a mismatch between DNS and the server configuration, the server will handshake with the outer ClientHello. The expectation when that happens is that the connection is authenticated for the public name, which gives an authenticated recovery signal. That means the server needs to, e.g., select a certificate that covers the public name. Thus, the outer SNI selection.

sayrer:
OK, and a server that didn't need the outer SNI name would be identifiable by IP address anyway. Makes sense.

sftcd:
On 10/03/2021 23:13, David Benjamin wrote:
> When there is a mismatch between DNS and the server configuration,
> the server will handshake with the outer ClientHello. The expectation
> when that happens is that the connection is authenticated for the
> public name, which gives an authenticated recovery signal. That means
[...]

PR #425, opened 2021-04-19
Relax public_name to public_identity and describe how clients construct CHOuter based on it.

### LINKED ISSUE #424: Should IP addresses be allowed for the client-facing server?
In #413, there's been a lot of discussion around how and whether or not IP addresses should be allowed as a reference identity of the client-facing server. This issue tracks sorting this particular question out. 

- @bemasc suggested [allowing the SNI be empty](https://github.com/tlswg/draft-ietf-tls-esni/issues/405#issuecomment-812088744) in the case the reference identity is an IP address. (Whether or not this address is in the ECHConfig is separate.)
- @martinthomson suggested [prohibiting IP addresses in the ECHConfig](https://github.com/tlswg/draft-ietf-tls-esni/pull/413#issuecomment-822089084). (Whether or not this allows the outer SNI to be empty is separate.)

So there seems to be two questions we need to sort out, in relation to #396:

1. Should the ECHConfig be allowed to supply IP addresses? And if so, how do client stacks validate them?
2. Should clients be allowed to validate servers using IP addresses as a reference identity? (If yes, then the ClientHelloOuter.SNI must be allowed to be empty/omitted.)

I think (2) should be 'yes', given that's currently possible today introduces no new complexity. I think it should be feasible to spell out the requirements to make (1) feasible, so I also think that should be 'yes'.

What do others think?

chris-wood:
(This change assumes the answers to (1) and (2) from #424 are both "yes.")

Previously, public_name was assumed to be a DNS name. #413 added validation criteria for this value. This change relaxes public_name to be a reference identity for the client-facing server. As a result, clients use this identity when constructing CHOuter and authenticating the client-facing server. This identity may be either a DNS name or an IP address, depending on the particular client-facing server deployment.

I'm totally open to bike shedding the validation criteria, but would really like to hear high-level opinions on this change before doing so.

cc @bemasc, @davidben, @sftcd, @martinthomson, @dmcardle 

Closes #424, #405, #396

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -227,7 +227,7 @@ The ECH configuration is defined by the following `ECHConfig` structure.
     struct {
         HpkeKeyConfig key_config;
         uint8 maximum_name_length;
-        opaque public_name<1..255>;
+        opaque public_identity<1..255>;
         Extension extensions<0..2^16-1>;
     } ECHConfigContents;
 
@@ -274,20 +274,12 @@ padding scheme described below. That could happen if wildcard names are in use,
 or if names can be added or removed from the anonymity set during the lifetime
 of a particular ECH configuration.
 
-public_name
-: The non-empty name of the client-facing server, i.e., the entity trusted to
-update the ECH configuration. This is used to correct misconfigured clients, as
-described in {{handle-server-response}}. This value MUST NOT begin or end with
-an ASCII dot and MUST be parsable as a dot-separated sequence of LDH labels, as
-defined in {{!RFC5890}}, Section 2.3.1. Clients MUST ignore any `ECHConfig`
-structure whose `public_name` does not meet these criteria. Note that these
-criteria are incomplete; they incidentally rule out textual representations of
-IPv6 addresses (see {{!RFC3986}}, Section 3.2.2), but do not exclude IPv4
-addresses in standard dotted-decimal or other non-standard notations such as
-octal and hexadecimal (see {{RFC3986}}, Section 7.4). If `public_name` contains
-a literal IPv4 or IPv6 address, the client SHOULD ignore the `ECHConfig` to
-avoid sending a non-compliant "server_name" extension on the ClientHelloOuter
-(see {{!RFC6066}}, Section 3).
+public_identity
+: The identity of the client-facing server, i.e., the entity trusted to updat
+the ECH configuration. This is used to construct the ClientHelloOuter, as
+described in {{handle-server-response}}. This MUST be a valid DNS name or
+IP address as described in {{identity-validate}}. Clients MUST ignore any
+`ECHConfig` structure whose `public_identity` does not meet these criteria.
 
 extensions
 : A list of extensions that the client must take into consideration when
@@ -321,6 +313,20 @@ serialized as follows.
 The `ECHConfigList` structure contains one or more `ECHConfig` structures in
 decreasing order of preference. This allows a server to support multiple
 versions of ECH and multiple sets of ECH parameters.
+
+## Public Identity Validation {#identity-validate}
+
+The `public_identity` value corresponding to an ECHConfig MUST contain either
+a DNS name or IP address. Validation criteria for each of these values is below:
+
+- DNS names: The value MUST NOT begin or end with an ASCII dot and MUST be
+  parsable as a dot-separated sequence of LDH labels, as defined in
+  {{!RFC5890}}, Section 2.3.1.
+- IP addresses: The value MUST be as described in {{!RFC5280}}, Section
+  4.2.1.6.
+
+An ECHConfig containing a `public_identity` that does not meet either of these
+criteria MUST be ignored.
 
 ## Configuration Extensions {#config-extensions}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -274,20 +274,12 @@ padding scheme described below. That could happen if wildcard names are in use,
...
-structure whose `public_name` does not meet these criteria. Note that these
-criteria are incomplete; they incidentally rule out textual representations of
-IPv6 addresses (see {{!RFC3986}}, Section 3.2.2), but do not exclude IPv4
-addresses in standard dotted-decimal or other non-standard notations such as
-octal and hexadecimal (see {{RFC3986}}, Section 7.4). If `public_name` contains
-a literal IPv4 or IPv6 address, the client SHOULD ignore the `ECHConfig` to
-avoid sending a non-compliant "server_name" extension on the ClientHelloOuter
-(see {{!RFC6066}}, Section 3).
+public_identity
+: The identity of the client-facing server, i.e., the entity trusted to updat
```
dmcardle, line 278: s/updat/update

draft-ietf-tls-esni.md
```diff
@@ -540,8 +546,11 @@ it does a standard ClientHello, with the exception of the following rules:
    {{flow-client-reaction}}.)
 1. It MUST include an "encrypted_client_hello" extension with a payload
    constructed as described below.
-1. The value of `ECHConfig.contents.public_name` MUST be placed in the
-   "server_name" extension.
+1. The "server_name" extension MUST be constructed based on the reference
+   identity in `ECHConfig.contents.public_identity`. Specifically, if
+   `ECHConfig.contents.public_identity` is a DNS name then this name is
+   set in the "server_name" extension. Otherwise, the "server_name" extension
+   is omitted.
```
dmcardle, line 553: A name like `0x42.0x42.0x42.0x42` would pass the DNS name validation, yet still be treated as an IP address by application code. I think it's worth acknowledging this possibility, so that clients cannot be induced into sending noncompliant server_name extensions to unsuspecting servers.

draft-ietf-tls-esni.md
```diff
@@ -322,6 +314,20 @@ The `ECHConfigList` structure contains one or more `ECHConfig` structures in
...
+## Public Identity Validation {#identity-validate}
+
+The `public_identity` value corresponding to an ECHConfig MUST contain either
+a DNS name or IP address. Validation criteria for each of these values is below:
+
+- DNS names: The value MUST NOT begin or end with an ASCII dot and MUST be
+  parsable as a dot-separated sequence of LDH labels, as defined in
+  {{!RFC5890}}, Section 2.3.1.
+- IP addresses: The value MUST be as described in {{!RFC5280}}, Section
+  4.2.1.6.
```
dmcardle, line 326: Subject Alternative Name actually leans on GeneralName, which is a pretty big sledgehammer, as discussed in #405.

dmcardle: Even if we only use `iPAddress` here, it will require some carveouts.  For starters, it actually defines a subnet, rather than a single address.

I think we're better off inferring DNS-or-address from the textual representation in `public_identity`.

draft-ietf-tls-esni.md
chris-wood, line 278:
```suggestion
: The identity of the client-facing server, i.e., the entity trusted to update
```

draft-ietf-tls-esni.md
chris-wood: > yet still be treated as an IP address by application code.

It would? 🤔

draft-ietf-tls-esni.md
chris-wood: I'm not leaning on GeneralName, but specifically on what's allowed for iPAddress, since that's what's relevant for certificate verification. I think the text here adequately covers both cases. I'm happy to take concrete suggestions if people have them.

bemasc:
As I mentioned in #424, I think RFC 8738's [approach](https://tools.ietf.org/html/rfc8738#section-6) is worth considering:

> As [RFC6066] does not permit IP
>    addresses to be used in the Server Name Indication (SNI) extension
>    HostName field, the server MUST instead use the IN-ADDR.ARPA
>    [RFC1034] or IP6.ARPA [RFC3596] reverse mapping of the IP address as
>    the HostName field value instead of the IP address string
>    representation itself.  For example, if the IP address being
>    validated is 2001:db8::1, the SNI HostName field should contain "1.0.
>    0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.8.b.d.0.1.0.0.2.ip6.arpa"

We would leave the SNI empty for the IP case, but use this as the encoding of IP addresses in `public_name`.  This encoding is not very efficient, but it's nicely unambiguous and widely implemented.

If we decide to stick with this ambiguous encoding, I would suggest relying on the URI `host` syntax from [RFC 3986](https://tools.ietf.org/html/rfc3986#section-3.2.2), rather than cloning it as this text does.

chris-wood:
(Noting here that we're bike shedding on encoding, and not on whether or not the identity can be an IP address.)

> We would leave the SNI empty for the IP case, but use this as the encoding of IP addresses in public_name. This encoding is not very efficient, but it's nicely unambiguous and widely implemented.

If folks think this is adequate and easy to implement, I'm fine with it. (I don't know how widely implemented this is.)

> If we decide to stick with this ambiguous encoding, I would suggest relying on the URI host syntax from RFC 3986, rather than cloning it as this text does.

I do not feel strongly, but, any particular reason why you'd lean on RFC 3986?

draft-ietf-tls-esni.md
dmcardle: Haha yes, this is an odd anti-feature, but I can verify it exists in Chrome and Safari. Visit http://0xac.0xd9.0xa.0xe, which is translated to google.com's IPv4 address: 172.217.10.14.

dmcardle: (To clarify, I just think it's plausible that application code could get confused by something like this. The above example shows lax client-side interpretation, not any server-side confusion.)

chris-wood: Wow, interesting. Thanks for clarifying!

PR #426, opened 2021-04-19
Move public_name to a mandatory extension.

chris-wood:
And in doing so, allow for future extensions to permit IP-based identities with a separate extension.

Alternate to #425.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -227,7 +227,6 @@ The ECH configuration is defined by the following `ECHConfig` structure.
     struct {
         HpkeKeyConfig key_config;
         uint8 maximum_name_length;
-        opaque public_name<1..255>;
         Extension extensions<0..2^16-1>;
     } ECHConfigContents;
 
@@ -273,21 +272,6 @@ can be set to zero, in which case clients SHOULD use the inner ClientHello
 padding scheme described below. That could happen if wildcard names are in use,
 or if names can be added or removed from the anonymity set during the lifetime
 of a particular ECH configuration.
-
-public_name
-: The non-empty name of the client-facing server, i.e., the entity trusted to
-update the ECH configuration. This is used to correct misconfigured clients, as
-described in {{handle-server-response}}. This value MUST NOT begin or end with
-an ASCII dot and MUST be parsable as a dot-separated sequence of LDH labels, as
-defined in {{!RFC5890}}, Section 2.3.1. Clients MUST ignore any `ECHConfig`
-structure whose `public_name` does not meet these criteria. Note that these
-criteria are incomplete; they incidentally rule out textual representations of
-IPv6 addresses (see {{!RFC3986}}, Section 3.2.2), but do not exclude IPv4
-addresses in standard dotted-decimal or other non-standard notations such as
-octal and hexadecimal (see {{RFC3986}}, Section 7.4). If `public_name` contains
-a literal IPv4 or IPv6 address, the client SHOULD ignore the `ECHConfig` to
-avoid sending a non-compliant "server_name" extension on the ClientHelloOuter
-(see {{!RFC6066}}, Section 3).
 
 extensions
 : A list of extensions that the client must take into consideration when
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -339,6 +323,40 @@ Clients MUST parse the extension list and check for unsupported mandatory
...
+or end with an ASCII dot and MUST be parsable as a dot-separated sequence of LDH
+labels, as defined in {{!RFC5890}}, Section 2.3.1. Clients MUST ignore any
+`ECHConfig` structure whose `public_name` does not meet these criteria. Note
+that this criteria incidentally permits a IPv4 addresses in standard
+dotted-decimal or other non-standard notations such as octal and hexadecimal
+(see {{!RFC3986}}, Section 7.4). Clients MUST ignore `ECHConfig` structures
+whose "public_name" payloads contain such addresses.
+
+The value of the "public_name" extension is hereafter referred to as the public
+identity. ECHConfig MUST contain this extension. Future extensions MAY override
```
davidben, line 355: Perhaps: "ECHConfig MUST contain this extension, and clients MUST ignore ECHConfigs missing the extension."

draft-ietf-tls-esni.md
```diff
@@ -724,44 +745,45 @@ an "illegal_parameter" alert.
...
 client MUST NOT enable the False Start optimization {{RFC7918}} for this
 handshake. If authentication or the handshake fails, the client MUST return a
 failure to the calling application. It MUST NOT treat this as a secure signal to
 disable ECH.
 
-Otherwise, when the handshake completes successfully with the public name
-authenticated, the client MUST abort the connection with an "ech_required"
-alert. The client can then regard ECH as securely disabled by the server. It
-SHOULD retry the handshake with a new transport connection and ECH disabled.
+Otherwise, when the handshake completes successfully with the reference
```
davidben, line 755: s/reference/public/?

draft-ietf-tls-esni.md
```diff
@@ -1557,18 +1580,33 @@ for ExtensionType (defined in {{!RFC8446}}):
...
+## ECHConfig Extension Registry
+
+This document defines a new ECHConfig extension registry that is maintained
+by IANA. Values with the first byte in the range 0-253 (decimal) are assigned
+via Specification Required {{!RFC8126}}.  Values with the first byte 254 or 255
+(decimal) are reserved for Private Use {{!RFC8126}}.
+
+This registry has a "Recommended" and "Mandatory" column. The registry is
+initially populated with a single entry for the "public_name" extension with
+"Recommended" and "Mandatory" values set to "Y" and "Y", respetively.
```
davidben, line 1603: (Just starting a discussion. I don't feel strongly either way, so whatever you think is best.)

It's interesting that we have a "Mandatory" column when it's purely a function of the value. On the one hand, it avoids readers needing to look at the constant. On the other hand, it might give the impression you can arbitrarily assign that column and you can't.

Maybe if the registry had a note at the top explaining how the mandatory bit works?

WDYT?

draft-ietf-tls-esni.md
```diff
@@ -1557,18 +1580,33 @@ for ExtensionType (defined in {{!RFC8446}}):
...
+registry for Alerts (defined in {{!RFC8446}}), with the "DTLS-OK" column set to
 "Y".
 
-# ECHConfig Extension Guidance {#config-extensions-guidance}
+## ECHConfig Extension Registry
+
+This document defines a new ECHConfig extension registry that is maintained
+by IANA. Values with the first byte in the range 0-253 (decimal) are assigned
+via Specification Required {{!RFC8126}}.  Values with the first byte 254 or 255
+(decimal) are reserved for Private Use {{!RFC8126}}.
```
davidben, line 1599: s/254/127/? Otherwise all Private Use extensions are also Mandatory.

draft-ietf-tls-esni.md
chris-wood: That's a good point -- I tried to add some text which clarifies that the value has implications on users.

draft-ietf-tls-esni.md
chris-wood: I did a slight variation on that -- thanks!

PR #436, opened 2021-05-26
Prohibit IP addresses in ECHConfig.public_name.

### LINKED ISSUE #396: A MUST for inclusion of public_name in outer SNI seems wrong

chris-wood:
This is yet another option to deal with public_name encoding, reference identity, and ClientHelloOuter construction issues. 

Closes #396, #405, #424

cc @martinthomson, @davidben, @cjpatton, @sftcd


draft-ietf-tls-esni.md
```diff
@@ -275,19 +275,15 @@ or if names can be added or removed from the anonymity set during the lifetime
...
-a literal IPv4 or IPv6 address, the client SHOULD ignore the `ECHConfig` to
-avoid sending a non-compliant "server_name" extension on the ClientHelloOuter
-(see {{!RFC6066}}, Section 3).
+: The non-empty DNS name of the client-facing server, i.e., the entity trusted
+to update the ECH configuration. This is used to correct misconfigured clients,
+as described in {{handle-server-response}}. This value MUST NOT begin or end
+with an ASCII dot and MUST be parsable as a dot-separated sequence of LDH
+labels, as defined in {{!RFC5890}}, Section 2.3.1. Clients MUST ignore any
+`ECHConfig` structure whose `public_name` does not meet these criteria. Note
+that these criteria allow IPv4 addresses in standard dotted-decimal or other
```
cjpatton, line 284:
```suggestion
that these criteria disallow IPv4 addresses in standard dotted-decimal or other
```
?

chris-wood: Sadly, no :( The format admits IP addresses, but the following sentence says they should be ignored in those cases.

sftcd:
Hiya,

On 26/05/2021 20:48, Christopher Wood wrote:
> Sadly, no :( The format admits IP addresses, but the following
> sentence says they should be ignored in those cases.
I forget why we wanted to even consider all this TBH:-)

I'm not sure if I'd want to write code to enforce those
rules on the client or not TBH.

One nit: FQDNs can (and implicitly do) end in a dot. I
bet the current text would annoy some DNS personages and
might even break something e.g. if the string were copied
from (some library that reads in) a zone file where that
trailing dot is significant when present, but often
absent. (I dunno if a library that'd make that bug likely
exists but I can easily imagine the code that'd cause a
dot at the end of a public_name.)

Cheers,
S.

chris-wood:
> I forget why we wanted to even consider all this TBH:-)

Do you mean consider dropping IP address support? The rationale was that encoding rules for addresses here is overly complicated, and most folks should be able to get a certificate with a name.

> I'm not sure if I'd want to write code to enforce those rules on the client or not TBH.

By rules, do you mean "ignore if IP address"?

Regarding the FQDN trailing dot, yeah, I figure we'd just assume that's implicit. We could also change the text to say the trailing dot is allowed. I don't feel strongly about that. I'm just trying to find a path forward.

draft-ietf-tls-esni.md
```diff
@@ -275,19 +275,15 @@ or if names can be added or removed from the anonymity set during the lifetime
...
+non-standard notations such as octal and hexadecimal (see {{?RFC3986}}, Section
+7.4). Clients MUST ignore the `ECHConfig` if it contains an encoded IP address.
```
martinthomson, line 286: So this says "work out if it is an IP address and don't allow it".  That's too imprecise for me.

The [parsing algorithm for IP addresses in the whatWG URL spec](https://url.spec.whatwg.org/#concept-ipv4-parser) is very precise.  You could reference that and say that if that returns a value then the ECHConfig can be ignored.

A MUST here without that level of precision is not good.  I can't test this.  Well, I can make something up, but that's not how this works.

martinthomson (changes requested):

draft-ietf-tls-esni.md
chris-wood: Thanks -- I don't think any of us (beyond you) knew that reference existed. We can certainly cite it to make this algorithm more precise.

chris-wood: @martinthomson I referenced it directly as a possible way to make this determination. I'm not sure we want to require use of this specific algorithm (especially since that spec is a living document and may change). We could alternatively try to pull that algorithm into this document, but that seems... terrible.

sftcd:
On 26/05/2021 21:28, Christopher Wood wrote:
>> I forget why we wanted to even consider all this TBH:-)
> 
> Do you mean consider dropping IP address support? 

I just meant I literally forget why we need any IP address
support at all within ECHConfigs.

> The rationale was
> that encoding rules for addresses here is overly complicated, and
> most folks should be able to get a certificate with a name.

Yep.

> 
>> I'm not sure if I'd want to write code to enforce those rules on
>> the client or not TBH.
> 
> By rules, do you mean "ignore if IP address"?

So it might be ok to delve into the public_name and do such
checks whilst encoding or decoding an ECHConfig but I'd not
wanna include that code into the OpenSSL extension handling.

> Regarding the FQDN trailing dot, yeah, I figure we'd just assume
> that's implicit. We could also change the text to say the trailing
> dot is allowed. I don't feel strongly about that. I'm just trying to
> find a path forward.

Sure, it was a nit, but one that might be worth handling. (I
didn't check the whatwg thing though, so maybe that handles
it too.)

Cheers,
S.


>

sftcd:
On 27/05/2021 02:23, Christopher Wood wrote:
>> So it might be ok to delve into the public_name and do such checks
>> whilst encoding or decoding an ECHConfig but I'd not wanna include
>> that code into the OpenSSL extension handling.
> 
> Would this be part of the extension handling? I imagine it would be
> underneath some API where you pass ECHConfig from elsewhere (DNS) to
> the TLS stack.

Right, that's what I was trying to say:-)

> 
> In any case, isn't some sort of processing required for ECHConfig? If
> it's not a DNS name, what would you do? I would expect OpenSSL, like
> any major stack, to implement the ECH rejection path using
> public_name, absent an application-specific knob to disable that. Am
> I misunderstanding you?

I'd need to look at the code but from memory the library
defaults to just providing a function to let the application
say to enforce the check (e.g. to dive into the SANs and see
if a string-form name matches some SAN). I've not stumbled
over any detailed/strict DNS name syntax checks that I recall
e.g. checking for <=63 octet label names or whatever else is
needed, though there may be such code already in order to
support DANE. Would have to look and see TBH - but adding
DNS name syntax checks that are only for ECH public_name
wouldn't be so attractive if it didn't do the same as was
already done elsewhere in the library or by a bunch of
applications using the library (that latter being v. hard to
know of course).

Cheers,
S.


>

draft-ietf-tls-esni.md
davidben: I agree with Martin that any MUST-level requirements should be precise. Though the WHATWG algorithm is... definitely hairy. :-/ I imagine most TLS libraries, including ours, will not have a version of it readily available, so we'd have to reimplement it.

I currently lean towards not trying too hard to filter IPs and instead saying:

* The public name is a DNS name. DNS names have such-and-such syntax. Anything outside that syntax is invalid. (We could, optionally, exclude the tighter [RFC3986 definition](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2), but see note below.)
* When your application evaluates whether a certificate is valid for a public name, you must interpret the public name as a DNS name. That means, you should match it against DNS entries in the SAN list.
* Note, however, that the syntax for IPv4 literals in URLs are also valid DNS names. If your application would otherwise interpret the public name as an IP literal, you should make sure you don't go down the IP address matching logic.
* At the TLS level, we consider the embedding of IPv4 literals in strings an application-specific thing and assume different applications may do this subtly differently.

The consequence of this is:

* If your application's certificate logic already separates DNS strings and IP strings, you can just send the string down the DNS string path and don't worry too hard.
* If your application's certificate logic takes a string and expects to parse out DNS vs IP address out of one string for web stuff, you should detect this (hopefully now your application's IP parser is available)
* Servers that live in an ecosystem where "127.0.0.1" is not a valid DNS name should not use "127.0.0.1" as a public name because it won't work.

In other words, the web-specific `MakeOriginFromDNSName("https", "127.0.0.1", 443)` function should fail because `https://127.0.0.1` gives back an origin that's interpreted as an IP literal, not a DNS name. It's not possible to express "127.0.0.1", the DNS name in the web.

This is slightly disappointing because, practically, that WHATWG-IP-literals are invalid DNS names is really a universal constraint. But we seem to have made enough of a mess here that I don't think we can guarantee the application won't think it is an IPv4 literal. (RFC3986 has [another definition](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2) of IPv4 literal, but if we used it, we'd still allow a bunch of strings that, on the web, is interpreted as an IPv4 literal. So applications would need to account for this `MakeOriginFromDNSName` error case regardless.)

chris-wood: @davidben two quick comments:

1. If we say "anything outside that syntax is invalid," without giving a function to determine if it's invalid, then we're back at Martin's criticism, right?
2. If the check is deferred to the certificate check, doesn't this run the risk of including an IP address in the outer SNI extension?

davidben: > If we say "anything outside that syntax is invalid," without giving a function to determine if it's invalid, then we're back at Martin's criticism, right?

Sorry, when I say "that syntax", I mean just the syntax that DNS defines. The one that is sadly ambiguous with IPv4 literals.

> If the check is deferred to the application when validating the certificate, doesn't this run the risk of including an IP address in the outer SNI extension?

It does, but I think that's fine. I think the two risks here are that the DNS can induce the client to send some string in SNI to some server, and it technically violates RFC6066. The former doesn't haven't a clear threat model, more fuzzy "well random private servers are often buggy". I think it's reasonable to say we'll validate the DNS-level syntax and assume the server behaves equally well given "a.b.c.d" or "0x1.0x2.0x3.0x4". The latter is true, but RFC6066 doesn't actually define the syntax anyway. I'm not really sure how one is meant to interpret that prohibition apart beyond "look deep into your heart and if this *feels* like an IP address, don't do it". And under this interpretation, TLS truly,  deeply in its heart, wants a DNS name of "0x1.0x2.0x3.0x4". That the application will never accept any certificate for that *DNS* name is the application's problem. :-)

(This is not a *great* interpretation. It's just barely self-consistent enough to say with a straight face, in hopes of getting out of this hole.)

chris-wood: Gotcha, thanks for clarifying! I'm fine with this as it avoids the parsing complexity altogether and moves it to other places in the stack -- "not my problem" =)

martinthomson (approved):
So I think that this is - unfortunately - the only sensible answer here.

The original sin here was the overloading of names and IP addresses in every interface that has ever accepted input from humans.  That's convenient, but the fact that an IPv4 address and domain name have overlapping syntax is a real problem.  However bad that is, that's the situation we have to deal with.

(For the record, 0x1.0x2.0x3.0x4 is a domain name always; the x makes it so.  IP addresses are always digits.  The real trick is in realizing that 127.1 and 2130706433 and 017700000001 are all localhost.)

As much as @davidben's idea here is appealing (i.e., the [SEP](https://en.wikipedia.org/wiki/Somebody_else%27s_problem) cloaking field), that is a great way to get real security problems.  If you have an input that results in different parts of the system disagreeing about semantics, the time to next vulnerability (MTTV?) goes right down.  I know that we are in the business of expanding the attack surface, but that's a bit to easy.

We can't rely on "usage" context to protect us either.  Say protocol X says "we don't accept IP addresses" or protocol Y says "we only accept strict decimal dotted quad".  That doesn't work if someone forgets to validate inputs.  That a problem is someone else's fault doesn't absolve us.

We might be able to rely on ICANN.  Might.  They haven't enabled all-numeric TLDs.  Their current rules exclude the possibility.  If you trust in that, then looking for the last label being all digits might work, but I'm not going to advocate for trust like that.  I'll just write the code.

I'll let you know how that goes.

davidben:
> (For the record, 0x1.0x2.0x3.0x4 is a domain name always; the x makes it so. IP addresses are always digits. The real trick is in realizing that 127.1 and 2130706433 and 017700000001 are all localhost.)

Is it? WHATWG parser happily accepts hex digits. Empirically, Chrome, Firefox, and Safari all happily accept, say, http://0x7f000001:8080/ as 127.0.0.1. I'm not seeing anything in WHATWG that says IPv4 literals are always digits. RFC3986 does, but RFC3986 is much more restrictive than what the web uses.
https://url.spec.whatwg.org/#ipv4-number-parser

martinthomson:
Yeah, I misread that.  I just went ahead and implemented this in rust, just to see what the algorithm might need to look like.  It's a gigantic mess.  Another lesson in how software engineering practices of the 1980s should be left there.

But you can all see how it worked out: https://github.com/martinthomson/url-ipv4

davidben:
Yeah, that gigantic mess, combined with ambiguity as to IP syntaxes (we have at least two so far), is why I was inclined to go the Somebody Else's Problem route. :-) But I can also live with a well-defined IP literal syntax to reject. Though, in order to be useful for browsers, that syntax must be the messy WHATWG one or a superset.

It's a lot more complexity for ECH (lots of people now need to reimplement the WHATWG parser), but I agree it does avoid reasoning about `DNSName` -> `DNSOrIP` type transitions downstream... provided your `DNSOrIP` type is compatible with the WHATWG parser. `DNSOrIP<WHATWG>` and `DNSOrIP<RFC3986>` are fine. If your random protocol's `DNSOrIP<BrandNewVariant>` is even more permissive than WHATWG's IP parser, you're out of luck and need the SEP provisions.

I don't think the SEP provisions are _quite_ that dire. Anyone transiting a `DNSOrIP` through their X.509 code has to be aware of the syntax *somewhere* because X.509 uses different name types for DNS vs. IP. Whatever you use to resolve that is the right parser to use when going from ECH's `DNSName` to your `DNSOrIP` type. And, hopefully, at that transition, the IP parser is readily available so we don't have duplicate code. Another thing to keep in mind is the public name doesn't need to interact with the *entire* system, just the TLS to certificate verification path. ClientHelloOuter handshakes are discarded.

That said, realistically, everyone just passes things around as strings, and you're right that this is something people need to remember to do. I dunno, it's a mess. I guess we pick whichever evil folks generally think is lesser.

martinthomson:
Yeah, the thing that concerns me is that one common system architecture, one that we use at least, has logic that makes three independent decisions.  That is, the `DNSOrIP` call is made for name resolution, TLS ECH, and certificate validation.  That's nasty duplication, so it's gross, but they could - at least in theory - share an implementation.  As you say, if one of those is more permissible than another in the wrong direction, we get bad outcomes.  To be clear, a lot else has to fail before this becomes a genuinely exploitable problem, but I don't want us to be making security holes.

I put up a patch for NSS that does this validation in case anyone wants to cheat. I probably should have been doing other work, but it's done now.  The LDH validation was easier, but still non-trivial.

chris-wood:
Given that the primary risk in disagreement between DNS or IP here seems to be interopability, and not security, perhaps we can downgrade the "MUST ignore" to a "SHOULD ignore" and clarify the risk? Failing early and consistently is certainly better, but the (possibly new) parsing requirement this imposes on the stack may not be worth the complexity. @davidben, @martinthomson: what do you think?

richsalz:
I know that the parsing in OpenSSL is naive, doesn't easily handle fallbacks/alternatives.  Further, it just calls the local platform code, which means there could be inconsistencies.

davidben:
Re SHOULD, if we do that, I think we should have some (probably less verbose) version of https://github.com/tlswg/draft-ietf-tls-esni/pull/436#discussion_r640744959 as guidance.

Whatever we do, we should say "ECH believes this is a DNS name, so please verify accordingly" because we need to define semantics for things we expose out of TLS. The IP address thing is how to square "DNS name" with all systems built around "DNS or IP" serializations. We can either:

* Say "ECH _promises_ not to give you something that looks like an IP" and define how to do that. But they do not use consistent or simple syntaxes, so this is tricky.
* Or we can say "you might get something that looks like an IP, deal with it" with rough guidance on how to deal with it. But this is punting up a layer which may or may not forget and misbehave. (How much do we care?)

I'm not sure there's a way out of picking one. Anything short of a MUST that covers the "DNS or IP" syntaxes we care about puts us in the second bucket.

chris-wood:
> Re SHOULD, if we do that, I think we should have some (probably less verbose) version of [#436 (comment)](https://github.com/tlswg/draft-ietf-tls-esni/pull/436#discussion_r640744959) as guidance.

Indeed — this is what I mean by clarifying the risk. I’d also keep the pointer to the WHATWG parser in case someone wants to enforce that and otherwise doesn’t have a parser to rely upon. 

> 
> Whatever we do, we should say "ECH believes this is a DNS name, so please verify accordingly" because we need to define semantics for things we expose out of TLS. The IP address thing is how to square "DNS name" with all systems built around "DNS or IP" serializations. We can either:
> 
> * Say "ECH _promises_ not to give you something that looks like an IP" and define how to do that. But "DNS or IP" systems do not use consistent or simple syntaxes, so this is tricky. (But probably doable since the WHATWG syntax is likely a superset of every other one.)
> * Or we can say "you might get something that looks like an IP, deal with it" with rough guidance on how to deal with it. But this is punting up a layer which may or may not forget and misbehave. (How much do we care?)

The sheer number of inconsistencies makes me lean towards the latter, with an _example_ of how to deal with the inconsistency. (That is, guidance for how to deal with it in ECH, and guidance for how to deal with it outside of ECH.)

> I'm not sure there's a way out of picking one. Anything short of a MUST that covers the "DNS or IP" syntaxes we care about puts us in the second bucket.

Yeah, I’m not suggesting we do nothing. I’m just proposing we relax the MUST for stacks that might otherwise fall back on a check elsewhere. This seems to balance things nicely.

davidben:
Ah, sure. Works for me, though I was already okay with the Somebody Else's Problem version. :-) (Would we even need a SHOULD at that point, or should we just leave IP parsing out of ECH's hands entirely? We'd be always punting it, but consistently so.)

@martinthomson, WDYT? I wasn't sure whether to read your most recent comment as changing your mind here or still preferring ECH apply an IP parser.

chris-wood:
> Would we even need a SHOULD at that point, or should we just leave IP parsing out of ECH's hands entirely? We'd be always punting it, but consistently so.

I learn towards keeping it for now and removing it later if, after we all implement the thing, consider it unnecessary. There seem to be enough strange edge cases with how applications might use the `DNSOrIp` value in verifying client-facing servers, and removing sharp edges seems useful. That said, if all of the major stacks hate this, then certainly that says a lot too. :-)

cbartle891:
I agree that we should just change MUST to SHOULD and add some verbiage about the risks. It strikes me that we're trying to close the barn doors after the horses are out. The fact is that most systems already have some sort of DNS and IP parser, some of which are probably not consistent with each other, and to now say "you have to be consistent and here are the rules you must abide by" just isn't realistic.

We had a case recently where we weren't allowing SNIs with an underscore--because we were following the spec--but when someone complained (https://twitter.com/ifette/status/1375559666340618242?s=20), we discovered that Chrome and Firefox were allowing underscores and changed our policy to reflect this, even though it went against the spec. I'm sure If we can't all follow the official DNS parsing rules now, without ECH, what makes you think we would all follow the official DNS and IP parsing rules later?

If we just say "use the parsers you already have" (and point to the WHATWG parser for those who don't already have one as @chris-wood suggested), at least we'll be gaining consistency within a system.

martinthomson:
Given that I've already written the code, there is a bit of sunk-cost that influences my thinking now...  

That said, I am thinking that the worst that can happen here if someone gets this wrong.  We get either:

* an ECH configuration is ignored because we falsely believe that it is an IP address
* a bad ECH configuration with an IP address is used because we falsely believe that it is a DNS name

For the former, if this happens, then we might not use ECH when we could.  That's not great, but we might get bug reports and can fix the parser (or people can use better names).

For the latter, we will put an IP in SNI and maybe the IP will make it as far as certificate validation.  If the server has a valid IP certificate for that address, then maybe it just works.  Maybe the server chokes on the SNI.  Maybe the client code chokes on the bizarre IP address and fails (this is what moz::pkix would do, it is picky about the format of addresses as it assumes that the caller normalizes them).  If the server doesn't have a valid certificate, then the connection fails.

In this latter case, it seems like failure to validate the ECH config correctly more often leads to connection failures.  That suggests that the SHOULD is worthwhile, because we don't want some clients connecting and others failing to connect.

Carrick is right to point to our collective differences, but I see that as unhealthy rather than inevitable.  We should fix the underscore thing by updating specifications to match reality (Anne van Kesteren is right about this...again), but I know the IETF well enough to be wary of proposing that, even if it would be the right thing to do.  I do think that the suggestion to reuse existing parsers is sensible.

So I'm down for making this change with a "SHOULD" in place of the "MUST".

chris-wood:
OK, I pushed an attempt at clarifying risk without being overly verbose. @davidben, @martinthomson, please have a look! @sftcd, would this current framing work for your implementation? (It seems like it would, especially if the amount of parsing is to be minimized.)

draft-ietf-tls-esni.md
```diff
@@ -275,25 +281,30 @@ or if names can be added or removed from the anonymity set during the lifetime
...
+as described in {{handle-server-response}}. This value MUST NOT begin or end
+with an ASCII dot and MUST be parsable as a dot-separated sequence of LDH
+labels, as defined in {{!RFC5890}}, Section 2.3.1. Clients MUST ignore any
+`ECHConfig` structure whose `public_name` does not meet these criteria. Note
+that these criteria allow IPv4 addresses in standard dotted-decimal or other
+non-standard notations such as octal and hexadecimal (see {{?RFC3986}}, Section
+7.4). Clients SHOULD ignore the `ECHConfig` if it contains an encoded IPv4
+address. (To determine if a public_name value is an IPv4 address, clients can
+invoke the IPv4 parser algorithm in {{WhatWGURLAddressParser}}. It returns a
+value when the input is an IPv4 address.) Failure to validate that public_name
```
martinthomson, line 295:
```suggestion
value when the input is an IPv4 address.) Failure to validate that a public_name
```

draft-ietf-tls-esni.md
```diff
@@ -275,25 +281,30 @@ or if names can be added or removed from the anonymity set during the lifetime
...
+is not an IPv4 address may cause connection failures to the client-facing
+server. However, it is not expected to induce security failures. Clients will
+use this name to authenticate the client-facing server and will fail closed if
+the client-facing server cannot produce a certificate bound to the same name.
```
martinthomson, line 299: "to the same name" might be tricky - if there is a disagreement about whether the name is an IP or DNS name.

How about:

```suggestion
server. However, it is not expected to induce security failures. Clients will
fail closed if the client-facing server cannot produce a certificate bound
to the a reference identity based on the public_name.
```

draft-ietf-tls-esni.md
```diff
@@ -554,6 +565,9 @@ it does a standard ClientHello, with the exception of the following rules:
    allows servers that reject ECH and use ClientHelloOuter to safely ignore any
    early data sent by the client per {{RFC8446}}, Section 4.2.10.
 
+Note that these rules may change in the presence of an application profile
+standard specifying otherwise.
```
martinthomson, line 569:
```suggestion
specifying otherwise.
```

martinthomson (approved):

draft-ietf-tls-esni.md
chris-wood: Much better!

draft-ietf-tls-esni.md
```diff
@@ -275,25 +281,30 @@ or if names can be added or removed from the anonymity set during the lifetime
...
+that these criteria allow IPv4 addresses in standard dotted-decimal or other
+non-standard notations such as octal and hexadecimal (see {{?RFC3986}}, Section
+7.4). Clients SHOULD ignore the `ECHConfig` if it contains an encoded IPv4
+address. (To determine if a public_name value is an IPv4 address, clients can
+invoke the IPv4 parser algorithm in {{WhatWGURLAddressParser}}. It returns a
+value when the input is an IPv4 address.) Failure to validate that a public_name
+is not an IPv4 address may cause connection failures to the client-facing
+server. However, it is not expected to induce security failures. Clients will
+fail closed if the client-facing server cannot produce a certificate bound
+to the a reference identity based on the public_name.
```
davidben, line 299: s/the a/the/

draft-ietf-tls-esni.md
```diff
@@ -554,6 +565,9 @@ it does a standard ClientHello, with the exception of the following rules:
    allows servers that reject ECH and use ClientHelloOuter to safely ignore any
    early data sent by the client per {{RFC8446}}, Section 4.2.10.
 
+Note that these rules may change in the presence of an application profile
+specifying otherwise.
```
davidben, line 569: No objection to the sentence, though I'm unsure what it's trying to do in the context of this PR.

davidben (commented):
*shrug* This works for me, although it still is unclear to me what we envision a TLS implementation doing here. I expect for most TLS implementations, "use the parsers you already have" isn't actionable because the parser you have is not available in the TLS library. It sounds like that's also true for @martinthomson, just there's a sunk cost thing where he's already implemented it. :-)

I worry we'll end up with a random mix of some TLS libraries doing a lot of work to reimplement the WHATWG algorithm (hopefully correctly!) and other libraries ignoring the text, depending on what they read between the lines of the SHOULD.

draft-ietf-tls-esni.md
```diff
@@ -42,6 +42,12 @@ normative:
   RFC2119:
   RFC7918:
 
+informative:
+  WhatWGURLAddressParser:
+   title: "URL Living Standard - IPv4 Parser"
+   target: https://url.spec.whatwg.org/#concept-ipv4-parser
```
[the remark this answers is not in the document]
chris-wood: Flagged as an open issue for now.

draft-ietf-tls-esni.md
chris-wood, line 299:
```suggestion
to the reference identity based on the public_name.
```

draft-ietf-tls-esni.md
chris-wood: I threw this in to allow clients to omit the public name in CHO.

richsalz:
> I worry we'll end up with a random mix of some TLS libraries doing a lot of work to reimplement the WHATWG algorithm (hopefully correctly!) and other libraries ignoring the text, depending on what they read between the lines of the SHOULD.

I share this concern.  See my comment above about OpenSSL.

chris-wood:
> we'll end up with a random mix of some TLS libraries doing a lot of work to reimplement the WHATWG algorithm (hopefully correctly!) and other libraries ignoring the text, depending on what they read between the lines of the SHOULD.

@davidben, @richsalz: it's true, but this seems inevitable. (We could omit the reference entirely, and I could imagine stacks trying to implement a parser and verifier to abide by the SHOULD.) What's the downside of different libraries choosing to handle this SHOULD differently?

chris-wood:
Paging @sftcd, @bemasc, and @cbartle891 for reviews!

draft-ietf-tls-esni.md
```diff
@@ -275,25 +281,30 @@ or if names can be added or removed from the anonymity set during the lifetime
...
+that these criteria allow IPv4 addresses in standard dotted-decimal or other
+non-standard notations such as octal and hexadecimal (see {{?RFC3986}}, Section
+7.4). Clients SHOULD ignore the `ECHConfig` if it contains an encoded IPv4
+address. (To determine if a public_name value is an IPv4 address, clients can
+invoke the IPv4 parser algorithm in {{WhatWGURLAddressParser}}. It returns a
+value when the input is an IPv4 address.) Failure to validate that a public_name
+is not an IPv4 address may cause connection failures to the client-facing
+server. However, it is not expected to induce security failures. Clients will
+fail closed if the client-facing server cannot produce a certificate bound
+to the reference identity based on the public_name.
```
bemasc, line 299: This assertion requires a bunch of assumptions:
* A server has published an IPv4 address in the public_name even though they're not supposed to, presumably because it's permitted by a future draft.
* public_name is delivered "securely".
* ICANN won't let anyone register a numeric TLD.
* Clients don't parse IPv4 strings into 4 octets, and then try to interpret that as a domain name, resulting in a memory safety violation.

I find these assumptions less than totally obvious, and that makes this text confusing.

Here's some alternative explanatory text.
```suggestion
value when the input is an IPv4 address.) Ignoring IPv4 addresses in `public_name`
allows their use to be defined in a future update to ECH without breaking
compatibility.
```

bemasc (approved):

draft-ietf-tls-esni.md
davidben: > Ignoring IPv4 addresses in `public_name` allows their use to be defined in a future update to ECH without breaking compatibility.

I don't think this is true. If that's the goal, we'd need a MUST-level requirement. With a SHOULD, some TLS implementations will ignore it, while others will go ahead and use that ECHConfig and likely get confused.

chris-wood: @bemasc clarifying question: by "ignoring IPv4 addresses in `public_name`," do you mean "ignore ECHConfigs that contain an IPv4 address," or "allow ECHConfigs that contain an IPv4 address (by ignoring the fact that the config contains one)"? (I'm parsing your suggestion as the latter, which seems counter to what this PR is trying to do.)

davidben:
>> we'll end up with a random mix of some TLS libraries doing a lot of work to reimplement the WHATWG algorithm (hopefully correctly!) and other libraries ignoring the text, depending on what they read between the lines of the SHOULD.
>
> @davidben, @richsalz: it's true, but this seems inevitable. (We could omit the reference entirely, and I could imagine stacks trying to implement a parser and verifier to abide by the SHOULD.) What's the downside of different libraries choosing to handle this SHOULD differently?

I don't think this is inevitable. While https://github.com/tlswg/draft-ietf-tls-esni/pull/436#discussion_r640744959 (i.e. no requirement or recommendation here at all) punts the problem to the application, it punts the problem consistently. The application can then use the parser they already have. Especially since we seem to agree it won't cause security problems, it seems we should just stick with TLS not having an opinion, with non-normative text to point out this means DNS name.

The downside of libraries choosing to handle the SHOULD differently is that the ecosystem behaves differently. We'd have to consider both kinds of stacks when reasoning about the protocol. It's also less clear for any given implementor what the right behavior would be. We seem to essentially be saying "you can reimplement the WHATWG protocol, which is just code, or you do less work and skip it, so do you want more code or less code?".

(I'll go put together some text for the "consistently punt it" option, so there's something concrete here.)

chris-wood:
> (I'll go put together some text for the "consistently punt it" option, so there's something concrete here.)

I understand that the text can consistently punt the problem elsewhere, unlike what this PR does, but that doesn't mean implementations will consistently punt it. 

In any case, if you could target your change against this PR to minimize deltas, that would be great.

draft-ietf-tls-esni.md
bemasc: @davidben Perhaps, but that's how I read the intention of the PR nonetheless.

@chris-wood The former.  I updated my comment to clarify.

draft-ietf-tls-esni.md
```diff
@@ -275,25 +281,30 @@ or if names can be added or removed from the anonymity set during the lifetime
...
-defined in {{!RFC5890, Section 2.3.1}}. Clients MUST ignore any `ECHConfig`
-structure whose `public_name` does not meet these criteria. Note that these
-criteria are incomplete; they incidentally rule out textual representations of
-IPv6 addresses (see {{!RFC3986, Section 3.2.2}}), but do not exclude IPv4
-addresses in standard dotted-decimal or other non-standard notations such as
-octal and hexadecimal (see {{RFC3986, Section 7.4}}). If `public_name` contains
-a literal IPv4 or IPv6 address, the client SHOULD ignore the `ECHConfig` to
-avoid sending a non-compliant "server_name" extension on the ClientHelloOuter
-(see {{!RFC6066, Section 3}}).
+: The non-empty DNS name of the client-facing server, i.e., the entity trusted
```
davidben, line 284: Couple minor nitpicks, just to make this less long-winded: I think we can drop "non-empty". It's already captured by both the struct definition and "DNS name". We can probably also replace "standard dotted-decimal or other non-standard notations such as octal and hexadecimal" with "dotted-decimal or other notations", since this PR ends up citing WHATWG later anyway. (Also it seems better not to get into the "standard" vs "non-standard" rabbithole.)

draft-ietf-tls-esni.md
```diff
@@ -42,6 +42,12 @@ normative:
   RFC2119:
   RFC7918:
 
+informative:
+  WhatWGURLAddressParser:
```
davidben, line 46: Perhaps s/WhatWGURLAddressParser/WHATWG-IPV4/ to be slightly shorter? Also, while not explicitly stated, the example in https://datatracker.ietf.org/doc/html/rfc7322#section-4.8.6.6 implies we like SHOUTY-CASE-WITH-HYPHENS for these tags.

richsalz:
>  it seems we should just stick with TLS not having an opinion, with non-normative text to point out this means DNS name.

Works for me.

davidben:
> (I'll go put together some text for the "consistently punt it" option, so there's something concrete here.)

https://github.com/tlswg/draft-ietf-tls-esni/pull/447

> In any case, if you could target your change against this PR to minimize deltas, that would be great.

Done. I've also left comments on this PR for the incidental stuff I noticed while putting that together.

draft-ietf-tls-esni.md
chris-wood: 👍 The variant in your PR is cleaner, so we can just use that if we decide to land this.

draft-ietf-tls-esni.md
chris-wood: 👍 I'm fine with this, too :)

sftcd:
On 07/06/2021 14:49, Christopher Wood wrote:
> Paging @sftcd, @bemasc, and @cbartle891 for reviews!
> 

Assuming I figured out the actual latest text, that
seems fine to me, except that DNS names can end with
an explicit dot, (which is always implicitly there)
as I think I pointed out  before. I don't know if TLS
stacks handle that well or badly but we should probably
either allow the trailing dot, or, if disallowing it
is really better, just say that we know we're not quite
perfectly correct, in DNS terms.

Cheers,
S.

draft-ietf-tls-esni.md
cbartle891: If someone is allowed to change these rules, then how can they really be MUSTs?

chris-wood:
Closing in favor of #456.

PR #438, opened 2021-05-29
Remove redundant length prefix from ClientHelloOuterAAD.

davidben:
ClientHello is already a defined structure. No need to wrap it in an opaque byte array. This avoids needing to fill in an extra length prefix in front.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -468,7 +468,7 @@ the following structure:
       HpkeSymmetricCipherSuite cipher_suite;
       uint8 config_id;
       opaque enc<1..2^16-1>;
-      opaque outer_hello<1..2^24-1>;
+      ClientHello outer_hello;
    } ClientHelloOuterAAD;
 ~~~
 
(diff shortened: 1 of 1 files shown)


eligrey:
Isn't the length prefix there in order to prevent ossification?

davidben:
Could you elaborate? Note AAD values aren't parsed or even sent over the wire. Rather, the requirement on them is injectivity.

davidben:
If the thinking is one might want to change what goes in the AAD later, the length prefix doesn't do anything. You'd need an actual distinguishing field like a type or version. We get that via labels and such when setting up the ECH context itself. By the time we've settled on speaking a particular ECH-like protocol, the AAD structure is fixed.

eligrey:
Thanks for the clarification. I was confused about the specifics of this prefix.

sayrer:
`Note this serialization uses the ClientHello structure from Section 4.1.2 of {{RFC8446}}, which does not include...`

Can probably just say `This serialization...` instead of leading with `Note`, which is almost always unnecessary.

draft-ietf-tls-esni.md
```diff
@@ -477,7 +477,8 @@ The first three parameters are equal to, respectively, the
 the payload of the "encrypted_client_hello" extension. The last parameter,
 `outer_hello`, is computed by serializing ClientHelloOuter with the
 "encrypted_client_hello" extension set to the empty string, i.e., the
-`extension_data` list has zero length. Note this serialization does not include
+`extension_data` list has zero length. Note this serialization uses the
+ClientHello structure from Section 4.1.2 of {{RFC8446}}, which does not include
```
chris-wood, line 481:
```suggestion
`extension_data` list has zero length. This serialization uses the
ClientHello structure from Section 4.1.2 of {{RFC8446}}, which does not include
```

chris-wood (approved):

PR #442, opened 2021-06-04
Preserve length prefixes in ClientHelloOuterAAD

davidben:
After having written so many variations of ClientHelloOuterAAD construction (client and server, test and production), I think this may be simpler. A little bit less ClientHello rewriting. The cost is predicting the payload length on the client (same as PSK binders), but that prediction is just adding the tag length to len(EncodedClientHelloInner). The server, in particular, needs to parse and then re-encode an arbitrary incoming ClientHelloOuter with transforms.

Thoughts?


martinthomson (commented):
This wouldn't be easier for us to implement; we insert the extension.  Yes, with `memmove()`.  We use extension insertion in a few places (padding), so it's not that difficult.

It's doable, and not that much more work to do it this way.

Did I miss the rationale for not including the HPKE parameters?  I know that some of these are in HPKE and so they are redundant here, but they are encoded in the e_c_h extension, so maybe there is a different spelling of this that only drops/zeros-out the encrypted portion.

davidben:
> Did I miss the rationale for not including the HPKE parameters? I know that some of these are in HPKE and so they are redundant here, but they are encoded in the e_c_h extension, so maybe there is a different spelling of this that only drops/zeros-out the encrypted portion.

That's exactly what this PR does. `payload` is the encrypted portion. (The whole extension body is `extension_data` in RFC8446. Guessing that was the disconnect?)

martinthomson:
Ah, precise terminology is useful up until the point that you aren't intimately familiar with each and every term.  That would be OK, but as I said, it's not an improvement for us.

davidben:
Hehe. I'm mostly hoping it'd tidy up the server side. When you decide to do ECH, you've already got to have parsed the ClientHello and found the ECH extension. Instead of going back, trimming out the ECH extension, preserving the surrounding extensions, and then updating the extensions length prefix, you can just take the offset of the payload field you just found, zero that and leave the rest alone.

martinthomson:
Yeah, if it is a server-side improvement (we don't heavily optimize any of this stuff, so we don't care much), then I won't stand in the way.

chris-wood (approved):
I've no strong preference here. This folds the same data into the AAD, just differently, so if it's easier to implement, 👍

briansmith:
The benefit of this change, IIRC, is that you don't need to reserialize anything. But the new proposed text still talks about reserializing. It would be good to find a way to reword it so it doesn't imply that reserialization is needed.

davidben:
@briansmith Rephrased.

briansmith:
@davidben LGTM! Thanks for making that change.

chris-wood:
@davidben would you mind resolving conflicts here?

davidben:
> would you mind resolving conflicts here?

Rebased.

chris-wood:
Thanks! Paging @cjpatton, @sftcd, and @cbartle891 for additional reviews.

draft-ietf-tls-esni.md
```diff
@@ -459,27 +459,27 @@ it.
...
-the four-byte header included in the Handshake structure.
+passing ClientHelloOuterAAD as the associated data for HPKE sealing and opening
+operations. The ClientHelloOuterAAD is a serialized ClientHello structure,
+defined in {{Section 4.1.2 of RFC8446}}, which matches the ClientHelloOuter
+except the `payload` field of the "encrypted_client_hello" is replaced with a
+byte string of the same length but whose contents are zeros. This value does
+not include the four-byte header from the Handshake structure.
+
+The client constructs ClientHelloOuterAAD first, followed by ClientHelloOuter.
+First, the client predicts the length of the `payload` based on the HPKE AEAD
```
cbartle891, line 470: I'm kind of thrown off by the use of the word "predict." It suggests to me that there's a chance that the prediction might be wrong. Should it not be "compute" instead? Or is there in fact a chance it might be wrong?

cbartle891:
Apart from my question about clarification, makes sense to me.

cbartle891 (approved):

sftcd:
On 08/06/2021 00:57, Christopher Wood wrote:
> Thanks! Paging @cjpatton, @sftcd, and @cbartle891 for additional reviews.
> 

LGTM

S.

davidben:
> I'm kind of thrown off by the use of the word "predict." It suggests to me that there's a chance that the prediction might be wrong. Should it not be "compute" instead? Or is there in fact a chance it might be wrong?

Good idea. I went to make that more precise and, in the process, realized a whole lot of the client procedure was specified funny so b7c3f61bfeb2e396f3408c0d15ee9d01be58416b fixes that up. (Those issues apply to the existing text too, so if we end up not merging this PR, I'll write an equivalent PR with just those bits.)

draft-ietf-tls-esni.md
```diff
@@ -569,42 +566,58 @@ implementations need to take care to ensure that sensitive extensions are not
...
 
 If optional configuration identifiers (see {{optional-configs}}) are used,
-`config_id` SHOULD be set to a randomly generated byte. Unless specified by the
-application using (D)TLS or externally configured on both sides,
-implementations MUST set the field as specified in
-{{encrypted-client-hello}}.
+`config_id` SHOULD be set to a randomly generated byte in the first
+ClientHelloOuter and MUST be left unchanged for the second ClientHelloOuter.
+
+The client serializes this structure compute the ClientHelloOuterAAD. It then
```
chris-wood, line 605:
```suggestion
The client serializes this structure to construct the ClientHelloOuterAAD. It then
```

draft-ietf-tls-esni.md
```diff
@@ -569,42 +566,58 @@ implementations need to take care to ensure that sensitive extensions are not
...
+computes the payload as:
+
+~~~
+    payload = context.Seal(ClientHelloOuterAAD,
```
chris-wood, line 609: Maybe we call this enc_payload (or final_payload or something) to distinguish between the payload extension value? The following sentence read a bit weird to me, where it refers to "payload" and "this value."

chris-wood (approved):

cjpatton (approved):
I'm happy with this.

draft-ietf-tls-esni.md
davidben: Done

davidben: (Went with final_payload because enc_payload seemed potentially confusing with enc.)

PR #443, opened 2021-06-04
Move ClientHello padding to the encoding.

### LINKED ISSUE #433: Embedded ClientHello padding complicates whole-input padding schemes
Filing this as a topic for discussion. I think "we don't like these padding strategies" or "that's not too complicated" are perfectly plausible answers.

So, we need to pad the ECH payload to avoid revealing information via length. It seems to me there are two plausible strategies for padding, and you could do one or both of them. (You could do both by applying the first, then applying the second.)

_NB: I don't know what the overall consensus is on padding strategies. My assumption has been that there isn't one and we're building flexibility into protocols so we can figure it out. The below is a summary of my current thinking, but there are probably lots of well-established ideas I don't know._ 

**Pad particular fields you are trying to hide**
There may be a particular field in your message with a length distribution you would like to hide. Perhaps some anonymity set has server name lengths up to 96 bytes. So you arrange for the client to know this, and the client adds 96 - len(server_name) extra bytes to the ClientHello somewhere. This is easy to reason about in isolation, but only considers individual fields. Information about other fields still leaks.

**Generically reduce the range of possible lengths**
You may also just try to generically reduce the range of the function. Maybe you round to a multiple of 32, or round to the next power of some R > 1. This may still leak information: padding a multiple of 32 still distinguishes (big server name, big ALPN list) and (small server name, small ALPN list) when the variance is more than 32 bytes. But it pretty uniformly reduces the information leaked in general, which can help increase anonymity sets. For example, an HTTP client will always sends an ALPN of {h2, http/1.1}, but a tiny amount of overall padding can help combine HTTP ClientHellos with a non-HTTP client that might have a slightly different ALPN set. It can also mitigate mistakes. For instance, TLS tickets often embed the server name. If the server forgets to pad that when minting the ticket, the first strategy alone will leak this.

The recommended padding scheme currently only does the first kind. We add padding for server_name and largely ignore the other fields. Suppose some client wanted to additionally use the second strategy. The order of operations here is fussy and interleaves bits of ClientHelloInner and EncodedClientHelloInner:

* EncodedClientHelloInner has a PSK binder
* ...which depends on the non-PSK extensions of ClientHelloInner
* ...which includes the padding extension
* ...which depends on the length of the non-padding extensions of EncodedClientHelloInner
* ...which depends on the length of the PSK binder, the non-PSK non-padding extensions of ClientHelloInner, which of them will be compressed

This is not actually a circular dependency. But it's a place the ClientHelloInner and EncodedClientHelloInner are awkward linked together. Probably the way to do it is to maintain a count of how much smaller EncodedClientHelloInner will be and subtract that from the ClientHelloInner as you construct and pad it.

*If* we think this is a problem, the fix is pretty straightforward: add padding to EncodedClientHelloInner itself. Something like:

```
struct {
  ClientHello client_hello;
  opaque padding<0..2^16-1>;
} EncodedClientHelloInner;
```

Now the ClientHelloInner no longer depends on the length of the EncodedClientHelloInner. Is this worth doing?

davidben:
Closes #433. This PR does four things:

1. Apply the EncodedClientHelloInner change suggested in #433.

2. Adjust the recommended padding scheme to round the overall length to
   a multiple of 32, now that this is fairly easy to compute.

3. It fixes an mistake in the recommended padding scheme. The server
   sets maximum_name_length to the longest *name*, but the client
   computes padding based on the server_name *extension*. We're
   miscounting 7 bytes of extension overhead and effectively are using
   maximum_name_length - 7 instead of maximum_name_length.

4. It defines what to do if you aren't sending SNI. While unlikely to
   happen in practice (no way to get an ECHConfig), implementations will
   need to handle this. The application may well call SetECHConfigList()
   without SetServerName() and it's rude to crash. It's pretty obvious what
   to do, so just define it.

(If folks would rather the PR only do a subset of these, I can defer the more controversial bits to other PRs. They all touched the same text, so it was easier to just do them all.)

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -455,9 +455,9 @@ extensions MUST be ordered consecutively in ClientHelloInner. The list of outer
 extensions, OuterExtensions, includes those which were removed from
 EncodedClientHelloInner, in the order in which they were removed.
 
-Finally, the client determines the number of padding bytes, N, and sets the
-`zeros` field to a byte string whose contents are N zeros. {{padding}}
-describes a recommended padding scheme.
+Finally, the client pads the message by setting the `zeros` field to a byte
+string whose contents are all zeros and whose length is the amount of padding
+to add. {{padding}} describes a recommended padding scheme.
 
 The client-facing server computes ClientHelloInner by reversing this process.
 First it parses EncodedClientHelloInner, interpreting all bytes after
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -426,11 +426,11 @@ which is the following structure:
     } EncodedClientHelloInner;
 ~~~
 
-The `client_hello` field is computed by first making a copy of
-ClientHelloInner with the `legacy_session_id` field with an empty string.
-Note this field uses the ClientHello structure, defined in {{Section 4.1.2
-of RFC8446}} which does not include the Handshake structure's four byte
-header. The `padding` field MUST be all zeroes.
+The `client_hello` field is computed by first making a copy of ClientHelloInner
+and setting the `legacy_session_id` field to the empty string. Note this field
+uses the ClientHello structure, defined in {{Section 4.1.2 of RFC8446}} which
+does not include the Handshake structure's four byte header. The `padding`
+field MUST be all zeroes.
 
 The client then MAY substitute extensions which it knows will be duplicated in
 ClientHelloOuter. To do so, the client removes and replaces extensions from
@@ -444,8 +444,11 @@ are all zeros. {{padding}} describes a recommended padding scheme.
 
 The client-facing server computes ClientHelloInner by reversing this process.
 First it parses EncodedClientHelloInner, interpreting all bytes after
-`client_hello` as padding. Next it makes a copy of the `client_hello` field and
-copies the `legacy_session_id` field from ClientHelloOuter. It then looks for an
+`client_hello` as padding. If any padding byte is non-zero, the server MUST
+abort the connection with an "illegal_parameter" alert.
+
+Next it makes a copy of the `client_hello` field and copies the
+`legacy_session_id` field from ClientHelloOuter. It then looks for an
 "ech_outer_extensions" extension. If found, it replaces the extension with the
 corresponding sequence of extensions in the ClientHelloOuter. If any referenced
 extensions are missing or if "encrypted_client_hello" appears in the list, the
@@ -702,8 +705,8 @@ Finally, the client SHOULD pad the entire message as follows:
    computed so far.
 2. Let P = 31 - ((L - 1) % 32) and add P bytes of padding.
 
-This rounds the overall structure up to a multiple of 32 bytes. This reduces
-the range of lengths of the EncodedClientHelloInner across all clients.
+This rounds the length of EncodedClientHelloInner up to a multiple of 32 bytes,
+reducing the set of possible lengths across all clients.
 
 In addition to padding ClientHelloInner, clients and servers will also need to
 pad all other handshake messages that have sensitive-length fields. For example,
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -438,15 +438,15 @@ which is the following structure:
 ~~~
     struct {
         ClientHello client_hello;
-        opaque padding[N];
+        uint8 zeros[length_of_padding];
     } EncodedClientHelloInner;
 ~~~
 
 The `client_hello` field is computed by first making a copy of ClientHelloInner
 and setting the `legacy_session_id` field to the empty string. Note this field
 uses the ClientHello structure, defined in {{Section 4.1.2 of RFC8446}} which
-does not include the Handshake structure's four byte header. The `padding`
-field MUST be all zeroes.
+does not include the Handshake structure's four byte header. The `zeros` field
+MUST be all zeroes.
 
 The client then MAY substitute extensions which it knows will be duplicated in
 ClientHelloOuter. To do so, the client removes and replaces extensions from
@@ -455,9 +455,9 @@ extensions MUST be ordered consecutively in ClientHelloInner. The list of outer
 extensions, OuterExtensions, includes those which were removed from
 EncodedClientHelloInner, in the order in which they were removed.
 
-Finally, the client determines the number of padding bytes, N, and sets
-the `padding` field to a byte string whose contents are N zeros.
-{{padding}} describes a recommended padding scheme.
+Finally, the client determines the number of padding bytes, N, and sets the
+`zeros` field to a byte string whose contents are N zeros. {{padding}}
+describes a recommended padding scheme.
 
 The client-facing server computes ClientHelloInner by reversing this process.
 First it parses EncodedClientHelloInner, interpreting all bytes after
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -696,7 +696,7 @@ value.
 1. If the ClientHelloInner contained a "server_name" extension with a name of
    length D, add max(0, L - D) bytes of padding.
 2. If the ClientHelloInner did not contain a "server_name" extension (e.g., if
-   the client is connecting to an IP address), add L + 7 bytes of padding. This
+   the client is connecting to an IP address), add L + 9 bytes of padding. This
    is the length of a "server_name" extension with an L-byte name.
 
 Finally, the client SHOULD pad the entire message as follows:
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -691,7 +691,7 @@ value.
 
 1. If the ClientHelloInner contained a "server_name" extension with a name of
    length D, add max(0, L - D) bytes of padding.
-2. If the ClientHelloInner did not contain a "server_name" extension (e.g. if
+2. If the ClientHelloInner did not contain a "server_name" extension (e.g., if
    the client is connecting to an IP address), add L + 7 bytes of padding. This
    is the length of a "server_name" extension with an L-byte name.
 
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -422,7 +422,7 @@ which is the following structure:
 ~~~
     struct {
         ClientHello client_hello;
-        opaque padding<0..2^16-1>;
+        opaque padding[N];
     } EncodedClientHelloInner;
 ~~~
 
@@ -443,8 +443,9 @@ Finally, the client sets the `padding` field to a byte string whose contents
 are all zeros. {{padding}} describes a recommended padding scheme.
 
 The client-facing server computes ClientHelloInner by reversing this process.
-First it makes a copy of the `client_hello` field and copies the
-`legacy_session_id` field from ClientHelloOuter. It then looks for an
+First it parses EncodedClientHelloInner, interpreting all bytes after
+`client_hello` as padding. Next it makes a copy of the `client_hello` field and
+copies the `legacy_session_id` field from ClientHelloOuter. It then looks for an
 "ech_outer_extensions" extension. If found, it replaces the extension with the
 corresponding sequence of extensions in the ClientHelloOuter. If any referenced
 extensions are missing or if "encrypted_client_hello" appears in the list, the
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -416,9 +416,21 @@ OuterExtensions consists of one or more ExtensionType values, each of which
...
-any PSK binders. It then computes a new value, the EncodedClientHelloInner, by
-first making a copy of ClientHelloInner. It then replaces the
-legacy\_session\_id field with an empty string.
+any PSK binders. It then computes a new value, the EncodedClientHelloInner,
+which is the following structure:
+
+~~~
+    struct {
+        ClientHello client_hello;
+        opaque padding<0..2^16-1>;
```
martinthomson, line 425: This would be easier if this wasn't length-prefixed.  Every padding scheme that ever used length has caused a lot of pain.  I realize that this isn't necessarily security-critical, but just extending with zeros should be easier to implement AND you don't have two extra bytes of overhead.

draft-ietf-tls-esni.md
```diff
@@ -662,15 +672,24 @@ extension across all application profiles. The target padding length of most
...
-2. Otherwise, let P = 31 - ((D - 1) % 32), and add P bytes of padding, plus an
-   additional 32 bytes if D + P < L + 32. This rounds D up to the nearest
-   multiple of 32 bytes that permits at least 32 bytes of length ambiguity.
+server's anonymity set without server input. Clients SHOULD use the ECHConfig's
+`maximum_name_length` field as follows, where L is the `maximum_name_length`
+value.
+
+1. If the ClientHelloInner contained a "server_name" extension with a name of
+   length D, add max(0, L - D) bytes of padding.
+2. If the ClientHelloInner did not contain a "server_name" extension (e.g. if
```
martinthomson, line 681:
```suggestion
2. If the ClientHelloInner did not contain a "server_name" extension (e.g., if
```

martinthomson (commented):
This isn't how I had planned to implement this, but it seems preferable.  Mostly because the padding calculation is clearer.  I had planned to do much the same as this, but with a padding extension.  This is less fuss.

...If only HPKE had padding.

draft-ietf-tls-esni.md
davidben: Works for me. Since every TLS structure is self-delimiting, we can just append stuff. It shifts some work from presentation language to prose, but I agree that it's easier to implement.

davidben: Done. I kinda cheated a bit around the presentation language, but hopefully that's clear enough?

davidben:
> This isn't how I had planned to implement this, but it seems preferable. Mostly because the padding calculation is clearer. I had planned to do much the same as this, but with a padding extension. This is less fuss.

Yeah, I'd actually already implemented the draft-10 construction, but I prefer this one. As you say, less fuss, especially if you want to do whole-message rounding. I also prefer rounding to a multiple of 32 at the end. If we decrease the range of possible values, it's easier to GREASE.

draft-ietf-tls-esni.md
```diff
@@ -416,9 +416,21 @@ OuterExtensions consists of one or more ExtensionType values, each of which
...
+    struct {
+        ClientHello client_hello;
+        opaque padding[N];
+    } EncodedClientHelloInner;
+~~~
+
+The `client_hello` field is computed by first making a copy of
+ClientHelloInner with the `legacy_session_id` field with an empty string.
+Note this field uses the ClientHello structure, defined in Section 4.1.2
+of {{RFC8446}} which does not include the Handshake structure's four byte
```
martinthomson, line 432:
```suggestion
Note this field uses the ClientHello structure, defined in {{Section 4.1.2
of RFC8446}} which does not include the Handshake structure's four byte
```

draft-ietf-tls-esni.md
```diff
@@ -416,9 +416,21 @@ OuterExtensions consists of one or more ExtensionType values, each of which
...
+
+~~~
+    struct {
+        ClientHello client_hello;
+        opaque padding[N];
+    } EncodedClientHelloInner;
+~~~
+
+The `client_hello` field is computed by first making a copy of
+ClientHelloInner with the `legacy_session_id` field with an empty string.
```
chris-wood, line 430:
```suggestion
ClientHelloInner and sets the `legacy_session_id` field to an empty string.
```

draft-ietf-tls-esni.md
```diff
@@ -416,9 +416,21 @@ OuterExtensions consists of one or more ExtensionType values, each of which
...
+Note this field uses the ClientHello structure, defined in Section 4.1.2
+of {{RFC8446}} which does not include the Handshake structure's four byte
+header.
```
chris-wood, line 433:
```suggestion
header. The `padding` field MUST be all zeroes.
```

draft-ietf-tls-esni.md
```diff
@@ -427,13 +439,13 @@ extensions MUST be ordered consecutively in ClientHelloInner. The list of outer
...
-defined in Section 4.1.2 of {{RFC8446}}. Note this does not include the
-four-byte header included in the Handshake structure.
+Finally, the client sets the `padding` field to a byte string whose contents
+are all zeros. {{padding}} describes a recommended padding scheme.
 
 The client-facing server computes ClientHelloInner by reversing this process.
-First it makes a copy of EncodedClientHelloInner and copies the
-legacy_session_id field from ClientHelloOuter. It then looks for an
+First it parses EncodedClientHelloInner, interpreting all bytes after
+`client_hello` as padding. Next it makes a copy of the `client_hello` field and
```
chris-wood, line 447: Should we require the server check that padding is all zeroes?

draft-ietf-tls-esni.md
```diff
@@ -662,15 +673,24 @@ extension across all application profiles. The target padding length of most
...
+   is the length of a "server_name" extension with an L-byte name.
+
+Finally, the client SHOULD pad the entire message as follows:
+
+1. Let L be the length of the EncodedClientHelloInner with all the padding
+   computed so far.
+2. Let P = 31 - ((L - 1) % 32) and add P bytes of padding.
+
+This rounds the overall structure up to a multiple of 32 bytes. This reduces
+the range of lengths of the EncodedClientHelloInner across all clients.
```
chris-wood, line 693:
```suggestion
This rounds the length of EncodedClientHelloInner up to a multiple of 32 bytes,
reducing lengths of this structure across all clients.
```

chris-wood (approved):
LGTM pending one question about server-side checks for all-zero padding.

draft-ietf-tls-esni.md
davidben: YOU CAN DO THAT?!?!!? Wow, I need to go update all my documents now.

draft-ietf-tls-esni.md
davidben: Agreed on the first half. I worry "reducing lengths" reads like making the lengths smaller. I guess "range of lengths" also has a different meaning... I was thinking the range of a function. :-)

How about "reducing the set of possible lengths across all clients".

chris-wood: That works for me too 👍

draft-ietf-tls-esni.md
davidben: Done, though s/sets/setting/

draft-ietf-tls-esni.md
davidben: Sure. Done.

draft-ietf-tls-esni.md

davidben:
Oops. Pushed one more fixup to correct the server_name extension overhead. I missed the ServerNameList length prefix, so the correct overhead is 9, not 7. See 95b91f0d0465d5fcee3cd0d73c84dc293f4146b5 for derivation.

draft-ietf-tls-esni.md
```diff
@@ -427,13 +439,16 @@ extensions MUST be ordered consecutively in ClientHelloInner. The list of outer
 extensions, OuterExtensions, includes those which were removed from
 EncodedClientHelloInner, in the order in which they were removed.
 
-Finally, EncodedClientHelloInner is serialized as a ClientHello structure,
-defined in Section 4.1.2 of {{RFC8446}}. Note this does not include the
-four-byte header included in the Handshake structure.
+Finally, the client sets the `padding` field to a byte string whose contents
+are all zeros. {{padding}} describes a recommended padding scheme.
```
cjpatton, line 443: We ought to define N in this paragraph. Something like: Finally, the client sets the `padding` field to a length-N all-zero byte string, where N is computed as described in {{padding}}.`

draft-ietf-tls-esni.md
```diff
@@ -662,15 +676,24 @@ extension across all application profiles. The target padding length of most
...
+   length D, add max(0, L - D) bytes of padding.
+2. If the ClientHelloInner did not contain a "server_name" extension (e.g., if
+   the client is connecting to an IP address), add L + 9 bytes of padding. This
+   is the length of a "server_name" extension with an L-byte name.
+
+Finally, the client SHOULD pad the entire message as follows:
+
+1. Let L be the length of the EncodedClientHelloInner with all the padding
+   computed so far.
+2. Let P = 31 - ((L - 1) % 32) and add P bytes of padding.
```
cjpatton, line 693: s/P/N/g

cjpatton (changes requested):
Just some editorial comments.

draft-ietf-tls-esni.md
chris-wood, line 693:
```suggestion
2. Let N = 31 - ((L - 1) % 32) and add N bytes of padding.
```

draft-ietf-tls-esni.md
chris-wood, line 443:
```suggestion
Finally, the client sets the `padding` field to a byte string of length N whose
contents are all zeros. {{padding}} describes a recommended padding scheme.
```

draft-ietf-tls-esni.md

draft-ietf-tls-esni.md
davidben: Done, although tweaked slightly since {{padding}} is merely recommended, not required.

davidben: Done, although tweaked slightly since {{padding}} is merely recommended, not required.

chris-wood:
@davidben one last thing -- would you mind rebasing to resolve conflicts?

davidben:
Rebased.

(Man, who is this jerk who keeps causing merge conflicts with my PRs? Let's see... oh, it's @davidben. That guy is the worst! :-D)

draft-ietf-tls-esni.md
```diff
@@ -416,9 +416,21 @@ OuterExtensions consists of one or more ExtensionType values, each of which
...
-any PSK binders. It then computes a new value, the EncodedClientHelloInner, by
-first making a copy of ClientHelloInner. It then replaces the
-legacy\_session\_id field with an empty string.
+any PSK binders. It then computes a new value, the EncodedClientHelloInner,
+which is the following structure:
+
+~~~
+    struct {
+        ClientHello client_hello;
+        opaque padding[N];
```
cjpatton, line 425: nit: `uint8 zeros[length_of_padding];` aligns a little nicer with [8446](https://datatracker.ietf.org/doc/html/rfc8446#section-5.2).

davidben: Done. It does make it no longer align with the requested variable name change in {{padding}}, but I agree with you aligning with 8446 makes sense.

cjpatton (approved):

draft-ietf-tls-esni.md
```diff
@@ -443,13 +455,17 @@ extensions MUST be ordered consecutively in ClientHelloInner. The list of outer
 extensions, OuterExtensions, includes those which were removed from
 EncodedClientHelloInner, in the order in which they were removed.
 
-Finally, EncodedClientHelloInner is serialized as a ClientHello structure,
-defined in {{Section 4.1.2 of RFC8446}}. Note this does not include the
-four-byte header included in the Handshake structure.
+Finally, the client determines the number of padding bytes, N, and sets the
```
cjpatton, line 458: nit (here and below):
```suggestion
Finally, the client determines the number of padding bytes, length_of_padding, and sets the
```

davidben: Well, N is referenced immediately afterwards and elsewhere we've used capital letters for random variables in prose. Note RFC8446 never utters length_of_padding in prose either. (Really what's going on here is that this is a questionably valid TLS message. The TLS message syntax is typically self-delimiting. But it's silly to add a length prefix here.)

I'll rewrite this to not reference a variable name.

chris-wood:
Given positive support here and on the list, I'm merging this to include in the next version of the draft.

PR #447, opened 2021-06-07
 Leave IPv4 address parsing to the application

davidben:
This sits on top of #436 (per https://github.com/tlswg/draft-ietf-tls-esni/pull/436#issuecomment-855992106) and is an implementation of https://github.com/tlswg/draft-ietf-tls-esni/pull/436#discussion_r640744959.
    
This is the "somebody else's problem" option, on the theory that IP parsers are typically outside the TLS stack, and, sadly, IPv4 address syntaxes are a bit inconsistent across applications.

This is papering over the fact that what DNS believes is a DNS name doesn't match reality. Almost everyone sticks IP literals and DNS names into one syntax, in slightly different variations. But we appear to have dug ourselves into that hole, so this PR attempts rationalize it by saying those applications accept a subset of the DNS syntax, and it's on the applications to manage their subsets.


draft-ietf-tls-esni.md
```diff
@@ -757,6 +764,15 @@ authenticate the connection with the public name, as follows:
...
   empty Certificate message, denoting no client certificate.
 
+When verifying the certificate, the client uses its application-specific
+certificate profile, e.g. {{Section 4.3.4 of
+?I-D.draft-ietf-httpbis-semantics}}, with a DNS-based reference identity. If
+the public name value is not a valid DNS name for the application,
+the client MUST reject the certificate. In particular, applications that
+incorporate DNS names and IP addresses into the same syntax (e.g. {{?RFC3986,
+Section 7.4}} and {{WHATWG-IPV4}}) MUST reject names that would be interpreted
+as IPv4 addresses.
```
chris-wood, line 774:
```suggestion
?I-D.draft-ietf-httpbis-semantics}}, with the public name as a DNS-based
reference identity. Applications that incorporate DNS names and IP addresses
into the same syntax (e.g. {{?RFC3986, Section 7.4}} and {{WHATWG-IPV4}}) MUST
reject names that would be interpreted as IPv4 addresses.
```

chris-wood: (simplifying to further punt on particulars of what certificate validators do with DNS names)

davidben: Hrm. I guess I was viewing "I think this is actually an IP" as a special case of "I don't accept that DNS name". But maybe that's an unnecessary generalization?

chris-wood: Yeah, that's what I'm thinking. Less words overall, and it seems to have the same effect.

martinthomson (commented):
I would greatly prefer #436.  I don't think that making this someone else's problem is the responsible thing to do.  Most applications take IPv4 and domain names in the same field.  That's how the Internet works, for better or worse, and this approach only creates a negative externality.  As minor as it might seem, that's not how I would like us to operate.

davidben:
Yeah, that we're really papering over DNS failing to match reality is a pretty solid argument. :-) I think the main thing that makes me nervous about #436 is not wanting to commit to a MUST. I feel it is better to definitively try to resolve the problem, or try to punt it, rather than leave it ambiguous. That is:

* Either, we decide that DNS's syntax rules are insufficient and we write down explicitly what we mean. Perhaps that's the WHATWG parser. That means we are deciding every TLS stack must have an implementation of that parser, that this implementation may well be redundant in many applications due to layering, and that we are okay with this.
* Or, we decide that we don't want to deal with this at the TLS level. This means we prefer that there not be redundant parsers, that TLS implementations not parse this messy format, and that we are okay with punting this to application, with all the risks that entails of the application failing to handle this.

I'm not opposed to either, but the more I think about it, the more unhappy I am with the SHOULD option. SHOULD seems to basically leave it to interpretation which side we're on.

chris-wood:
Closing in favor of #456.

PR #455, opened 2021-06-09
Require compressed extensions appear in the same relative order.

### LINKED ISSUE #378: Naive outer_extensions decoding is a DoS risk
A naive implementation of decoding the inner ClientHello can easily go quadratic when processing outer_extensions. Thus far, nothing in TLS has required searching for O(N) extensions, only O(implemented extensions). It's likely a lot of TLS implementations handle the extensions list linearly most of the time.

outer_extensions is now the first time you have to search for extensions you may not have implemented. A naive implementation would then go quadratic, which is a server DoS risk. It doesn't need to go quadratic (for instance, you can sort ClientHelloOuter's extensions and then binary search per lookup), but it's probably worth at least calling this out, if not switching to a simpler decoding scheme.

(For a simpler decoder, we already require CHInner group the compressed extensions consecutively. We could do the same with CHOuter, at which point we only need to reference one byte slice, be it by offset/length or some goofier scheme.)

davidben:
Closes #378. This is an alternative to #452, which constrains the client such this suggestion is feasible:
https://github.com/tlswg/draft-ietf-tls-esni/pull/452#discussion_r647858091

@martinthomson, thoughts? This does constrain the client behavior, though our draft implementation would not be affected by it. (We run through all the extensions in our usual order and compute the inner value + whether to compress it. If we intend to compress it, we write it to a separate buffer instead of the pending ClientHelloInner. Then we drop in all the buffered extensions, so they're contiguous and, in the EncodedClientHelloInner, write ech_outer_extensions instead of the buffer. This process keeps the relative order of compressed extensions, so we follow this rule anyway.)

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -431,13 +431,13 @@ may use the "ech_outer_extensions" extension.
 OuterExtensions consists of one or more ExtensionType values, each of which
 reference an extension in ClientHelloOuter. The extensions in OuterExtensions
 MUST appear in ClientHelloOuter in the same relative order, however, there is
-no requirement that they be continguous. For examples, OuterExtensions may
+no requirement that they be continguous. For example, OuterExtensions may
 contain extensions A, B, C, while ClientHelloOuter contains extensions A, D, B,
 C, E, F.
 
 The "ech_outer_extensions" extension is only used for compressing the
-ClientHelloInner. It MUST NOT be sent in either ClientHelloOuter or
-ClientHelloInner.
+ClientHelloInner. It can only be included in EncodedClientHelloInner, and MUST
+NOT be sent in either ClientHelloOuter or ClientHelloInner.
 
 When sending ClientHello, the client first computes ClientHelloInner, including
 any PSK binders. It then computes a new value, the EncodedClientHelloInner, by
@@ -1661,17 +1661,14 @@ The following procedure processes the "ech_outer_extensions" extension (see
    * If E is "encrypted_client_hello", abort the connection with an
      "illegal_parameter" alert and terminate this procedure.
 
-   * While the I is less than N and the Ith extension of
+   * While I is less than N and the I-th extension of
      ClientHelloOuter does not have type E, increment I.
 
    * If I is equal to N, abort the connection with an "illegal_parameter"
      alert and terminate this procedure.
 
-   * Otherwise, the Ith extension of ClientHelloOuter has type E. Copy
+   * Otherwise, the I-th extension of ClientHelloOuter has type E. Copy
      it to the EncodedClientHelloInner and increment I.
-
-Implementations that retain the ClientHelloOuter extension list in another
-iterable form can equivalently replace I with an iterator.
 
 # Acknowledgements
 
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -472,31 +472,11 @@ are true:
 * The extensions in ClientHelloOuter corresponding to those in OuterExtensions
   do not occur in the same order.
 
-If the cost to decode an EncodedClientHelloInner is disproportionately large in
-comparison to the input size, a malicious client could exploit this in a denial
-of service attack. Implementations are RECOMMENDED to use the following
-decoding procedure, which runs in linear time:
-
-1. Let `i` be zero and `n` be the number of extensions in ClientHelloOuter.
-
-1. For each extension type, `ext`, in OuterExtensions:
-
-   * If `ext` is "encrypted_client_hello", abort the connection with an
-     "illegal_parameter" alert and terminate this procedure.
-
-   * While the `i` is less than `n` and the `i`th extension of
-     ClientHelloOuter does not have type `ext`, increment `i`.
-
-   * If `i` is equal to `n`, abort the connection with an "illegal_parameter"
-     alert and terminate this procedure.
-
-   * Otherwise, the `i`th extension of ClientHelloOuter has type `ext`. Copy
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -435,25 +443,44 @@ The client-facing server computes ClientHelloInner by reversing this process.
...
+   * If `ext` is "encrypted_client_hello", abort the connection with an
+     "illegal_parameter" alert and terminate this procedure.
+
+   * While the `i` is less than `n` and the `i`th extension of
+     ClientHelloOuter does not have type `ext`, increment `i`.
+
+   * If `i` is equal to `n`, abort the connection with an "illegal_parameter"
+     alert and terminate this procedure.
+
+   * Otherwise, the `i`th extension of ClientHelloOuter has type `ext`. Copy
```
martinthomson, line 477: One nitpick: "`i`th" renders really badly in text.  Consider just avoiding the backticks.

martinthomson (approved):
I'm not going to nitpick this, but I like it.

We currently have the same triply-nested loop here:
* the outer loop iterates over extensions in CHInner
* for most extensions we just copy them into a new structure
* for ech_outer we have two more layers of loop
* the middle loop iterates over the items in ech_outer
* the inner loop iterates over CHOuter, looking for a match

This wouldn't really reduce code complexity, but it would mean that we wouldn't need to start over every time for that innermost loop.  So we would be able to run in linear time as this says.

Like you, as we iterate over extensions in the same order when constructing messages, this wouldn't have any effect on generating a ClientHello.

My only concern with this is that it is somewhat more prescriptive than I would generally like for a specification.  Moving this algorithmic stuff to an appendix might be a good balance.  Either way, if you are adding this much text, it needs a section heading.

sftcd:
On 09/06/2021 07:38, Martin Thomson wrote:
> I'm not going to nitpick this, but I like it.

Same here (wrt liking:-).

Cheers,
S.

davidben:
> My only concern with this is that it is somewhat more prescriptive than I would generally like for a specification. Moving this algorithmic stuff to an appendix might be a good balance. Either way, if you are adding this much text, it needs a section heading.

Moved to appendix.

chris-wood:
@davidben can you please rebase?

cjpatton (approved):
Works for me!

davidben:
Rebased

draft-ietf-tls-esni.md
```diff
@@ -429,7 +429,15 @@ may use the "ech_outer_extensions" extension.
 ~~~
 
 OuterExtensions consists of one or more ExtensionType values, each of which
-reference an extension in ClientHelloOuter.
+reference an extension in ClientHelloOuter. The extensions in OuterExtensions
+MUST appear in ClientHelloOuter in the same relative order, however, there is
+no requirement that they be continguous. For examples, OuterExtensions may
```
chris-wood, line 434:
```suggestion
no requirement that they be continguous. For example, OuterExtensions may
```

draft-ietf-tls-esni.md
```diff
@@ -429,7 +429,15 @@ may use the "ech_outer_extensions" extension.
...
+contain extensions A, B, C, while ClientHelloOuter contains extensions A, D, B,
+C, E, F.
+
+The "ech_outer_extensions" extension is only used for compressing the
+ClientHelloInner. It MUST NOT be sent in either ClientHelloOuter or
+ClientHelloInner.
```
chris-wood, line 440:
```suggestion
ClientHelloInner. It can only be included in EncodedClientHelloInner, and MUST
NOT be sent in either ClientHelloOuter or ClientHelloInner.
```

chris-wood: I was really confused by this at first, so hopefully this helps clarify for others.

draft-ietf-tls-esni.md
```diff
@@ -1641,6 +1649,30 @@ and (3) obtaining the generic SNI to bootstrap the connection. In contrast,
...
+{{encoding-inner}}) in linear time:
+
+1. Let I be zero and N be the number of extensions in ClientHelloOuter.
+
+1. For each extension type, E, in OuterExtensions:
+
+   * If E is "encrypted_client_hello", abort the connection with an
+     "illegal_parameter" alert and terminate this procedure.
+
+   * While the I is less than N and the Ith extension of
```
chris-wood, line 1664:
```suggestion
   * While I is less than N and the I-th extension of
```

draft-ietf-tls-esni.md
```diff
@@ -1641,6 +1649,30 @@ and (3) obtaining the generic SNI to bootstrap the connection. In contrast,
...
+     ClientHelloOuter does not have type E, increment I.
+
+   * If I is equal to N, abort the connection with an "illegal_parameter"
+     alert and terminate this procedure.
+
+   * Otherwise, the Ith extension of ClientHelloOuter has type E. Copy
```
chris-wood, line 1670:
```suggestion
   * Otherwise, the I-th extension of ClientHelloOuter has type E. Copy
```

draft-ietf-tls-esni.md
```diff
@@ -1641,6 +1649,30 @@ and (3) obtaining the generic SNI to bootstrap the connection. In contrast,
...
+     it to the EncodedClientHelloInner and increment I.
+
+Implementations that retain the ClientHelloOuter extension list in another
+iterable form can equivalently replace I with an iterator.
+
```
chris-wood, line 1675:
```suggestion
```
This seems really language and library specific, so I might just drop it.

chris-wood (approved):
LGTM pending suggestions!

PR #456, opened 2021-06-11
Require that someone validate public_name correctly.

### LINKED ISSUE #396: A MUST for inclusion of public_name in outer SNI seems wrong

chris-wood:
This combines #436 and #447, requiring that validation of public_name happen _somewhere_, providing a reference parser (if one does not exist), and allowing it to happen when parsing ECHConfig.

Closes #396, #405, #424

cc @martinthomson, @davidben, @cjpatton, @sftcd

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -301,6 +301,8 @@ extensions
 generating a ClientHello message. These are described below
 ({{config-extensions}}).
 
+[[OPEN ISSUE: determine if clients should enforce a 63-octet label limit for
+public_name]]
 [[OPEN ISSUE: fix reference to WHATWG-IPV4]]
 
 The `HpkeKeyConfig` structure contains the following fields:
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -284,16 +284,14 @@ public_name
 to update the ECH configuration. This is used to correct misconfigured clients,
 as described in {{handle-server-response}}.
 
-: Clients MUST ignore any `ECHConfig` structure
-whose public_name does is not parsable as a
-dot-separated sequence of LDH labels, as defined in
-{{!RFC5890, Section 2.3.1}} or which
-begins or end with an ASCII dot.
-:  Clients SHOULD ignore the
-`ECHConfig` if it contains an encoded IPv4 address. To determine if a
-public_name value is an IPv4 address, clients can invoke the IPv4 parser
-algorithm in {{WHATWG-IPV4}}. It returns a value when the input is an IPv4
-address.
+: Clients MUST ignore any `ECHConfig` structure whose public_name is not
+parsable as a dot-separated sequence of LDH labels, as defined in
+{{!RFC5890, Section 2.3.1}} or which begins or end with an ASCII dot.
+
+: Clients SHOULD ignore the `ECHConfig` if it contains an encoded IPv4 address.
+To determine if a public_name value is an IPv4 address, clients can invoke the
+IPv4 parser algorithm in {{WHATWG-IPV4}}. It returns a value when the input is
+an IPv4 address.
 
 : See {{auth-public-name}} for how the client interprets and validates the
 public_name.
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -289,8 +289,7 @@ whose public_name does is not parsable as a
 dot-separated sequence of LDH labels, as defined in
 {{!RFC5890, Section 2.3.1}} or which
 begins or end with an ASCII dot.
-: These criteria allow IPv4 addresses in dotted-decimal or other notations,
-e.g. {{?RFC3986, Section 7.4}} and {{WHATWG-IPV4}}. Clients SHOULD ignore the
+:  Clients SHOULD ignore the
 `ECHConfig` if it contains an encoded IPv4 address. To determine if a
 public_name value is an IPv4 address, clients can invoke the IPv4 parser
 algorithm in {{WHATWG-IPV4}}. It returns a value when the input is an IPv4
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -42,6 +42,12 @@ normative:
   RFC2119:
   RFC7918:
 
+informative:
+  WHATWG-IPV4:
+   title: "URL Living Standard - IPv4 Parser"
+   target: https://url.spec.whatwg.org/#concept-ipv4-parser
+   date: May 2021
+
 --- abstract
 
 This document describes a mechanism in Transport Layer Security (TLS) for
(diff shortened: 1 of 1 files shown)


davidben (approved):

cjpatton (approved):

draft-ietf-tls-esni.md
```diff
@@ -274,25 +280,27 @@ constrain server name lengths. Names may exceed this length if, e.g.,
...
+with an ASCII dot and MUST be parsable as a dot-separated sequence of LDH
+labels, as defined in {{!RFC5890, Section 2.3.1}}. Clients MUST ignore any
+`ECHConfig` structure whose public_name does not meet these criteria. Note
+that these criteria allow IPv4 addresses in dotted-decimal or other notations,
+e.g. {{?RFC3986, Section 7.4}} and {{WHATWG-IPV4}}. Clients SHOULD ignore the
+`ECHConfig` if it contains an encoded IPv4 address. (To determine if a
+public_name value is an IPv4 address, clients can invoke the IPv4 parser
+algorithm in {{WHATWG-IPV4}}. It returns a value when the input is an IPv4
+address.) See {{auth-public-name}} for how the client interprets and validates
+the public_name.
```
martinthomson, line 295: Try splitting this up a little.  You can start multiple sections with ": " to make separate paragraphs (I think).

martinthomson (approved):

draft-ietf-tls-esni.md
chris-wood: Oh, nice. I didn't know that was possible. I broke up the description into logical units.

draft-ietf-tls-esni.md
```diff
@@ -274,25 +280,32 @@ constrain server name lengths. Names may exceed this length if, e.g.,
...
-criteria are incomplete; they incidentally rule out textual representations of
-IPv6 addresses (see {{!RFC3986, Section 3.2.2}}), but do not exclude IPv4
-addresses in standard dotted-decimal or other non-standard notations such as
-octal and hexadecimal (see {{RFC3986, Section 7.4}}). If `public_name` contains
-a literal IPv4 or IPv6 address, the client SHOULD ignore the `ECHConfig` to
-avoid sending a non-compliant "server_name" extension on the ClientHelloOuter
-(see {{!RFC6066, Section 3}}).
+: The DNS name of the client-facing server, i.e., the entity trusted
+to update the ECH configuration. This is used to correct misconfigured clients,
+as described in {{handle-server-response}}.
```
ekr, line 285: Let's separate out the conformance requirements and the check. I think you want to say "This value MUST be a valid domain name in accordance with XXX".  I believe that would still allow trailing (but not leading) dot, so you may also have to say "The value MUST NOT end with an ASCII dot".

draft-ietf-tls-esni.md
```diff
@@ -274,25 +280,32 @@ constrain server name lengths. Names may exceed this length if, e.g.,
...
+
+: This value MUST NOT begin or end with an ASCII dot and MUST be parsable as a
+dot-separated sequence of LDH labels, as defined in
+{{!RFC5890, Section 2.3.1}}. Clients MUST ignore any `ECHConfig` structure
+whose public_name does not meet these criteria.
+
```
ekr, line 291:
```suggestion
: Clients MUST ignore any `ECHConfig` structure
whose public_name does is not parsable as a
dot-separated sequence of LDH labels, as defined in
{{!RFC5890, Section 2.3.1}} or which
begins or end with an ASCII dot.
```

draft-ietf-tls-esni.md
```diff
@@ -274,25 +280,32 @@ constrain server name lengths. Names may exceed this length if, e.g.,
...
+: These criteria allow IPv4 addresses in dotted-decimal or other notations,
+e.g. {{?RFC3986, Section 7.4}} and {{WHATWG-IPV4}}. Clients SHOULD ignore the
+`ECHConfig` if it contains an encoded IPv4 address. To determine if a
+public_name value is an IPv4 address, clients can invoke the IPv4 parser
+algorithm in {{WHATWG-IPV4}}. It returns a value when the input is an IPv4
+address.
```
ekr, line 297:
```suggestion
:  Clients SHOULD ignore the
`ECHConfig` if it contains an encoded IPv4 address. To determine if a
public_name value is an IPv4 address, clients can invoke the IPv4 parser
algorithm in {{WHATWG-IPV4}}. It returns a value when the input is an IPv4
address.
```

draft-ietf-tls-esni.md
ekr: Note that I think this still allows for bogus names, like ones that are super-long. Do we want to allow clients to check for that?

ekr (changes requested):

draft-ietf-tls-esni.md
chris-wood: The public_name field is already limited to 255 characters, so I think we're fine.

sftcd:
On 14/06/2021 20:26, Christopher Wood wrote:
> @chris-wood commented on this pull request.
> 
> 
> 
>> +: This value MUST NOT begin or end with an ASCII dot and MUST be parsable as a
> +dot-separated sequence of LDH labels, as defined in
> +{{!RFC5890, Section 2.3.1}}. Clients MUST ignore any `ECHConfig` structure
> +whose public_name does not meet these criteria.
> +
> 
> The public_name field is already limited to 255 characters, so I think we're fine.

The max for a label in a DNS name is 63 octets though isn't
it?

>

chris-wood:
> The max for a label in a DNS name is 63 octets though isn't it?

Yeah, but RFC6066 has no similar requirements (that I can tell), so I don't see we need to impose anything new here. Maybe I'm missing something? And if so, can you please propose text?

draft-ietf-tls-esni.md
chris-wood: Good call -- I believe your suggestions rephrased things to make the requirements and check read easier.

sftcd:
On 14/06/2021 20:33, Christopher Wood wrote:
>> The max for a label in a DNS name is 63 octets though isn't it?
> 
> Yeah, but RFC6066 has no similar requirements (that I can tell), so I
> don't see we need to impose anything new here. Maybe I'm missing 
> something? And if so, can you please propose text?

Fair ask:-)

I guess the public_name mostly has to go in a cert SAN so
would it help to use or refer to rfc5280 text?

The most relevant bit there seems to be:

    When the subjectAltName extension contains a domain name
    system label, the domain name MUST be stored in the
    dNSName (an IA5String). The name MUST be in the "preferred
    name syntax", as specified by Section 3.5 of [RFC1034] and
    as modified by Section 2.1 of [RFC1123].  Note that while
    uppercase and lowercase letters are allowed in domain
    names, no significance is attached to the case.  In
    addition, while the string " " is a legal domain name,
    subjectAltName extensions with a dNSName of " " MUST NOT
    be used.

(That text is near [1]).

That said, that " " is a legal domain name surprises me (and
I'm a co-author of 5280!)

BTW, I'm fine that that all gets fixed later if that's
better.

S.

[1] https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.6

chris-wood:
> BTW, I'm fine that that all gets fixed later if that's better.

Sounds good -- I dropped an OPEN ISSUE to note this for later. @sftcd, would you have cycles to take that issue and propose a PR to address it?

sftcd:
On 14/06/2021 20:52, Christopher Wood wrote:
>> BTW, I'm fine that that all gets fixed later if that's better.
> 
> Sounds good -- I dropped an OPEN ISSUE to note this for later.
> @sftcd, would you have cycles to take that issue and propose a PR to
> address it?

Sneaky eh:-)

Sure, I can look at it this week but don't wait on me.

S

>

chris-wood:
> Sure, I can look at it this week but don't wait on me.

Yeah, there's no rush here, I think. This seems like a smaller issue to resolve at some point.

PR #459, opened 2021-06-14
Add 10-to-11 change log.

chris-wood:
Since Padding is still being discussed, I figured we'd leave related issues out of this version and take this opportunity to mark the completion of most HRR issues.

cc @davidben, @martinthomson, @cjpatton, @cbartle891, @ekr

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -1662,7 +1662,7 @@ Issue and pull request numbers are listed with a leading octothorp.
 
 - Relax computation of the acceptance signal (#420, #449)
 
-- Fix ClientHelloOuterAAD generation (#438, #442)
+- Simplify ClientHelloOuterAAD generation (#438, #442)
 
 - Allow empty enc in ClientECH (#444)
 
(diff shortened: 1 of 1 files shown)


cjpatton (approved):
I would note which changes are wire breaking, but otherwise looks good.

draft-ietf-tls-esni.md
```diff
@@ -1656,6 +1656,31 @@ important ideas and contributions.
 
 Issue and pull request numbers are listed with a leading octothorp.
 
+## Since draft-ietf-tls-esni-10
+
+- Make HRR confirmation and ECH acceptance explicit (#422, #423)
+
+- Relax computation of the acceptance signal (#420, #449)
+
+- Fix ClientHelloOuterAAD generation (#438, #442)
```
davidben, line 1665: Maybe s/Fix/Simplify/? The old version wasn't broken. Just this new one's hopefully simpler.

davidben (approved):

draft-ietf-tls-esni.md
chris-wood: Yep, that's better!

chris-wood, line 1665:
```suggestion
- Simplify ClientHelloOuterAAD generation (#438, #442)
```

Issue #466, opened 2021-06-23
Client behavior if the HRR signal is wrong


davidben:
{{client-hrr}} says the client assumes ECH was accepted if the signal matches, but it doesn't say what to do if the signal doesn't match. Is it an error, or do you ignore it?

Related to #450, though whether we decide to do that, we need to write down *something* in the meantime.

cbartle891:
Seems reasonable to presume rejection, as if the extension weren't present at all, particularly if we end up GREASEing the extension.

davidben:
Right now the server's not supposed to send it at all. Normally if the peer sends invalid things, we make it an error. But, yeah, allowing it and presuming rejection is plausible when we GREASE it. But that's not currently in the draft, and I'm not sure how coherent GREASEing it actually is. Right now, draft-11 as currently published is ambiguous, which makes implementing it hard.

cjpatton:
I agree we need to say something here. The way I interpreted it in my implementation is: presume ECH acceptance if the ECH extension is present and its payload matches the acceptance signal; otherwise, presume rejection. I think we should try to spell this behavior in a way that make sense.

chris-wood:
+1 to assuming rejection, since this is exactly what we do in the non HRR case. @davidben, will you be able to prep a PR for this today? We can include this clarification in -12.

davidben:
Sorry, was OOO yesterday and today was weekend email catchup. I might have time tomorrow?

Alright, sounds like everyone likes presuming rejection, so presuming rejection it is. :-)

PR #469, opened 2021-06-28
Add -12 change log.

chris-wood:
cc @cjpatton, @ekr, @davidben, @sftcd

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -1721,6 +1721,8 @@ Issue and pull request numbers are listed with a leading octothorp.
 
 - Move ClientHello padding to the encoding (#443)
 
+- Align codepoints (#464)
+
 - Relax OuterExtensions checks for alignment with RFC8446 (#467)
 
 - Editorial improvements (#468, #465, #462, #461)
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -1717,6 +1717,14 @@ important ideas and contributions.
...
 Issue and pull request numbers are listed with a leading octothorp.
 
+## Since draft-ietf-tls-esni-11
+
+- Move ClientHello padding to the encoding (#443)
+
+- Relax OuterExtensions checks for alignment with RFC8446 (#467)
+
+- Editorial improvements (#468, #465, #462, #461)
+
```
cjpatton, line 1727: I think we should call out #464 as a bugfix. ECH-11 isn't implementable.

chris-wood: Well, not really. It just broke the convention.

chris-wood: Fixed anyway.

chris-wood:
Happy to park this if we want to land a change for #466, but I don't think that's absolutely needed for a new version.

draft-ietf-tls-esni.md
davidben: Well, it was a convention break along with an incompatibility with ECH-10. So I guess it's implementable, but you don't want to implement it the way it was written. :-)

davidben (approved):
LGTM, though I think a fix for #466 would be prudent since it's currently ambiguous.

cjpatton (approved):

PR #475, opened 2021-07-15
Revise middlebox section.

### LINKED ISSUE #474: Clarification of section 8.2. Middleboxes
Hello,

While reading this specification, section 8.2. states that "MITM proxies" are required to remove any extensions they do not understand. 

> A more serious problem is MITM proxies which do not support this
   extension.  [RFC8446], Section 9.3 requires that such proxies remove
   any extensions they do not understand.  The handshake will then
   present a certificate based on the public name, without echoing the
   "encrypted_client_hello" extension to the client.

However, looking at RFC 8446 Section 9.3, the first section you encounter on forwarding middleboxes explicitly states the opposite: the middlebox is not allowed to change or modify any parameters in the handshake and subsequent traffic if it doesn't understand some parameter in the ClientHello:

> A middlebox which forwards ClientHello parameters it does not
   understand MUST NOT process any messages beyond that ClientHello.
   It MUST forward all subsequent traffic unmodified.  Otherwise, it
   may fail to interoperate with newer clients and servers.

Reading further, we find the bullet point the warning in section 8.2 refers to:

> A middlebox which terminates a TLS connection MUST behave as a
   compliant TLS server (to the original client), including having a
   certificate which the client is willing to accept, and also as a
   compliant TLS client (to the original server), including verifying
   the original server's certificate.  In particular, it MUST
   generate its own ClientHello containing only parameters it
   understands, and it MUST generate a fresh ServerHello random
   value, rather than forwarding the endpoint's value.

Here, the middlebox isn't so much required to "remove any extension they do not understand", but to "generate its own ClientHello containing only parameters it understands". After carefully reading over these sections multiple times, I do get the message, but I feel like it could be better worded as not to create confusion. The words "MITM proxy" could refer to the forwarding middlebox as well. Changing the wording in, for example, the following way makes the reference more clear in my opinion:

> A more serious problem is MITM proxies which do not support this
extension.  [RFC8446], Section 9.3 requires that such proxies generate
their own ClientHello containing only parameters they understand. The
middlebox will then present a certificate based on the public name to
the client, without echoing the "encrypted_client_hello" extension.

What do you think about this?

davidben:
Closes #474. This makes a few changes to the middlebox section:

First, using "MITM proxies" when RFC8446 talks about terminating TLS was
confusing, so align with the RFC8446 terminology.

Next, RFC8446 does not quite require dropping unknown extensions, only
generating your own ClientHello. Dropping unknown extensions happens if
the proxy tries to intersect its true capabilities with that of the
client. I've also dropped the "a more serious problem" introduction
since the point of this section is that it's not a serious problem.

Finally, I've dropped the "non-conformant MITM proxy" paragraph. It's
slightly out of date now that we've unified everything around trial
decryption. I believe, depending on what the proxy does with the
server's EncryptedExtensions response, things will either break, work on
accident, or retry loop and hit the limit. It doesn't seem worth
expending a whole lot of text on something that is the proxy's fault.

This does drop the "Clients SHOULD NOT attempt to repair the connection
in this case" line, but the client couldn't really detect this case
anyway. It looks like any old connection failure without a proxy, and
the usual considerations for changing your TLS config on connection
failure apply.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -1128,21 +1128,16 @@ DNS results, if one is provided.
 
 ## Middleboxes
 
-A more serious problem is MITM proxies which do not support this extension.
-{{RFC8446, Section 9.3}} requires that such proxies remove any extensions they
-do not understand. The handshake will then present a certificate based on the
-public name, without echoing the "encrypted_client_hello" extension to the
-client.
+{{RFC8446, Section 9.3}} requires TLS-terminating proxies ignore unknown
+parameters, and generate their own ClientHello containing only parameters they
+understand. Thus, when presenting a certificate to the client or sending a
+ClientHello to the server, the proxy will act as if connecting to the public
+name, without echoing the "encrypted_client_hello" extension.
 
 Depending on whether the client is configured to accept the proxy's certificate
 as authoritative for the public name, this may trigger the retry logic described
 in {{handle-server-response}} or result in a connection failure. A proxy which
 is not authoritative for the public name cannot forge a signal to disable ECH.
-
-A non-conformant MITM proxy which instead forwards the ECH extension,
-substituting its own KeyShare value, will result in the client-facing server
-recognizing the key, but failing to decrypt the SNI. This causes a hard failure.
-Clients SHOULD NOT attempt to repair the connection in this case.
 
 # Compliance Requirements {#compliance}
 
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -1128,11 +1128,13 @@ DNS results, if one is provided.
 
 ## Middleboxes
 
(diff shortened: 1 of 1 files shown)


martinthomson (commented):
The middlebox could also support ECH, right? It would need a way to get a config, and probably would need to be involved in feeding a config too the client, but it is at least possible for interception to consume and use ECH.

Maybe qualify with "unless the middlebox understands ECH and is able to interpret the extension (which likely means being involved in providing the client with an ECHConfig)..."

davidben:
Ah good point. I dropped the "which do not support this extension" clause on accident. How's this new iteration?

(Seems best not to get into middleboxes which do support the extension. "which likely means being involved in" feels too handwaivy, and I'd rather not expend a lot of text around how DNS works with HTTP CONNECT proxies, CDNs where the "proxy" is on the server's side and controls DNS, etc.)

chris-wood (approved):

PR #478, opened 2021-07-23
Rephrase the overview

davidben:
The overview was a bit long, and slightly inaccurate. The
ClientHelloInner is encrypted after (most) of the ClientHelloOuter is
constructed. I've also merged the ECH-naive server case with the ECH
rejection case. This is more consistent with the client text, and most
other TLS extensions, where we do not distinguish between reasons why
the extension could not be negotiated.

I've also generally tightened up the wording to make it shorter. As part
of that, I'm trimmed the repeated list of sensitive ClientHello
extensions. ECH itself is (mostly) agnostic to which extensions you
believe are sensitive, and we already gave examples in the introduction.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -148,11 +148,10 @@ and backend servers are physically separated.
 
 A client-facing server enables ECH by publishing an ECH configuration, which
 is an encryption public key and associated metadata. The server must publish
-this for all the domains it serves directly or indirectly (via Split Mode).
-This document defines the ECH configuration's format, but delegates DNS
-publication details to {{!HTTPS-RR=I-D.ietf-dnsop-svcb-https}}. Other delivery
-mechanisms are also possible. For example, the client may have the ECH
-configuration preconfigured.
+this for all the domains it serves via Shared or Split Mode. This document
+defines the ECH configuration's format, but delegates DNS publication details
+to {{!HTTPS-RR=I-D.ietf-dnsop-svcb-https}}. Other delivery mechanisms are also
+possible. For example, the client may have the ECH configuration preconfigured.
 
 When a client wants to establish a TLS session with some backend server, it
 constructs a private ClientHello, referred to as the ClientHelloInner.
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -146,44 +146,35 @@ and backend servers are physically separated.
 
 ## Encrypted ClientHello (ECH)
 
-ECH allows the client to encrypt sensitive ClientHello extensions, e.g., SNI,
-ALPN, etc., under the public key of the client-facing server. This requires the
-client-facing server to publish the public key and metadata it uses for ECH for
-all the domains for which it serves directly or indirectly (via Split Mode).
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -163,44 +163,35 @@ and backend servers are physically separated.
...
-   extension, then it terminates the handshake using the ClientHelloOuter. This
-   is referred to as "ECH rejection". When ECH is rejected, the client-facing
-   server sends an acceptable ECH configuration in its EncryptedExtensions
-   message.
-1. If it supports ECH and decrypts the extension, it forwards the
-   ClientHelloInner to the backend server, who terminates the connection. This
-   is referred to as "ECH acceptance".
+A client-facing server enables ECH by publishing an ECH configuration, which
+is an encryption public key and associated metadata. The server must publish
+this for all the domains it serves directly or indirectly (via Split Mode).
```
chris-wood, line 168:
```suggestion
this for all the domains it serves directly or indirectly via Shared or Split Mode, respectively.
```

draft-ietf-tls-esni.md
```diff
@@ -163,44 +163,35 @@ and backend servers are physically separated.
...
+This document defines the ECH configuration's format, but delegates DNS
+publication details to {{!HTTPS-RR=I-D.ietf-dnsop-svcb-https}}. Other delivery
+mechanisms are also possible. For example, the client may have the ECH
+configuration preconfigured.
+
+When a client wants to establish a TLS session with some backend server, it
+constructs a secret ClientHello, referred to as the ClientHelloInner.
```
chris-wood, line 175:
```suggestion
constructs a private ClientHello, referred to as the ClientHelloInner.
```

draft-ietf-tls-esni.md
```diff
@@ -163,44 +163,35 @@ and backend servers are physically separated.
...
+   to the backend server, which completes the handshake. This is referred to
+   as accepting ECH.
 
 Upon receiving the server's response, the client determines whether or not ECH
-was accepted and proceeds with the handshake accordingly. (See
-{{client-behavior}} for details.)
+was accepted ({{handle-server-response}}) and proceeds with the handshake
+accordingly. When ECH is rejected, the connection is not directly usable for
+the client, but allows the client to retry with up-to-date configuration
+({{rejected-ech}}).
```
chris-wood, line 194:
```suggestion
accordingly. When ECH is rejected, the resulting connection is not usable by
the client for application data. Instead, ECH rejection allows the client to
retry with up-to-date configuration ({{rejected-ech}}).
```

chris-wood (approved):
This is very clean :-) Nice work!

draft-ietf-tls-esni.md
davidben: Oh, I got this from the earlier version of the text. I wonder whether both OR clauses are even necessary. WDYT about:

> for all the domains it serves via Shared or Split Mode.

davidben: Applied that version and rebased.

chris-wood: Lovely :-) thanks!

PR #481, opened 2021-07-29
Allow the client-facing server to GREASE the extension.

chris-wood:
This assumes the formulation where, on ECH reject, the server is the client-facing server. In other words, backend servers will never get a CH without ECH having been accepted, so GREASE doesn't apply there. 

cc @davidben

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -998,6 +998,9 @@ If the client-facing server rejected ECH, or if the first ClientHello did not
 include an "encrypted_client_hello" extension, the client-facing server
 proceeds with the connection as usual. The server does not decrypt the
 second ClientHello's ECHClientHello.payload value, if there is one.
+If the client-facing server rejected ECH and the first ClientHello included an
+"encrypted_client_hello" extension, the client-facing server MAY include
+the "encrypted_client_hello" extension with a payload of 8 random bytes.
 
 Note that a client-facing server that forwards the first ClientHello cannot
 include its own "cookie" extension if the backend server sends a
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -948,11 +948,17 @@ messages between the client and backend server unmodified.
 
 Otherwise, if all candidate ECHConfig values fail to decrypt the extension, the
 client-facing server MUST ignore the extension and proceed with the connection
-using ClientHelloOuter. This connection proceeds as usual, except the server
-MUST include the "encrypted_client_hello" extension in its EncryptedExtensions
-with the "retry_configs" field set to one or more ECHConfig structures with
-up-to-date keys. Servers MAY supply multiple ECHConfig values of different
-versions. This allows a server to support multiple versions at once.
+using ClientHelloOuter. If this does not result in sending a HelloRetryRequest,
+the connection proceeds as usual, except the server MUST include the
+"encrypted_client_hello" extension in its EncryptedExtensions with the
+"retry_configs" field set to one or more ECHConfig structures with up-to-date
+keys. Servers MAY supply multiple ECHConfig values of different versions. This
+allows a server to support multiple versions at once. Otherwise, if this results
+in sending a HelloRetryRequest, and the corresponding ClientHello included an
+"encrypted_client_hello" extension, the client-facing server MAY include the
+"encrypted_client_hello" extension in the HelloRetryRequest with a payload of
+8 random bytes; see {{dont-stick-out}} for details. Afterwards, the
+client-facing server proceeds as described in {{client-facing-server-hrr}}.
 
 Note that decryption failure could indicate a GREASE ECH extension (see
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -998,6 +998,9 @@ If the client-facing server rejected ECH, or if the first ClientHello did not
 include an "encrypted_client_hello" extension, the client-facing server
 proceeds with the connection as usual. The server does not decrypt the
 second ClientHello's ECHClientHello.payload value, if there is one.
+If the client-facing server rejected ECH and the first ClientHello included an
+"encrypted_client_hello" extension, the client-facing server MAY include
+the "encrypted_client_hello" extension with a payload of 8 random bytes.
```
cjpatton, line 1003: * In which message may the extension appear? ServerHello, HelloRetryRequest, or both?
* I think we should make it clear that the purpose is to grease. Maybe point to Security Considerations and expand the relevant section to talk about this?

cjpatton (commented):
I'm good with the change, but I think the spec needs more context for it.

draft-ietf-tls-esni.md
chris-wood: Good suggestions -- I moved some stuff around to try and clarify.

cjpatton (approved):

draft-ietf-tls-esni.md
```diff
@@ -998,6 +1004,9 @@ If the client-facing server rejected ECH, or if the first ClientHello did not
 include an "encrypted_client_hello" extension, the client-facing server
 proceeds with the connection as usual. The server does not decrypt the
 second ClientHello's ECHClientHello.payload value, if there is one.
+Moreover, the server MUST include the "encrypted_client_hello" extension in its
+EncryptedExtensions with the "retry_configs" field set to one or more ECHConfig
+structures with up-to-date keys, as described in {{client-facing-server}}.
```
davidben, line 1009: This doesn't have anything to do with HRR. Do we need to restate it from the section here? (I wonder if the server half could also do with an editorial pass to make it more chronological.)

draft-ietf-tls-esni.md
```diff
@@ -948,11 +948,17 @@ messages between the client and backend server unmodified.
...
+the connection proceeds as usual, except the server MUST include the
+"encrypted_client_hello" extension in its EncryptedExtensions with the
+"retry_configs" field set to one or more ECHConfig structures with up-to-date
+keys. Servers MAY supply multiple ECHConfig values of different versions. This
+allows a server to support multiple versions at once. Otherwise, if this results
+in sending a HelloRetryRequest, and the corresponding ClientHello included an
+"encrypted_client_hello" extension, the client-facing server MAY include the
+"encrypted_client_hello" extension in the HelloRetryRequest with a payload of
+8 random bytes; see {{dont-stick-out}} for details. Afterwards, the
+client-facing server proceeds as described in {{client-facing-server-hrr}}.
```
davidben, line 961: This is a little long. How about:

```
Otherwise, if all candidate ECHConfig values fail to decrypt the extension, the
client-facing server MUST ignore the extension and proceed with the connection
using ClientHelloOuter, with the following modifications:

* If sending a HelloRetryRequest, the server MAY include an
  "encrypted_client_hello" extension with a payload of 8 random bytes; see
  {{dont-stick-out}} for details.

* The server MUST include the "encrypted_client_hello" extension in its
  EncryptedExtensions message with ....
```

This also fixes the issue where we need to duplicate the EncryptedExtensions text (see comment below).

Though, it occurs to me, even that's a little odd because even non-ECH servers could send HRR.ECH for GREASE. Maybe we can handle that be deleting a few words. Suppose you are a server with zero ECHConfigs. The dispatch in {#server-behavior} still sends you to this section. The candidate ECHConfig values is then empty, so you'll hit this case. That means we actually should replace:

> The server MUST include the "encrypted_client_hello" extension in its EncryptedExtensions message with [...]

with:

> If the server is configured with any ECHConfigs, it MUST include the "encrypted_client_hello" extension in its EncryptedExtensions message with [...]

chris-wood: That modification is good in both places -- I'll take it!

draft-ietf-tls-esni.md
chris-wood: We should do such a pass, but let's do it separately.

chris-wood: (The server still includes configs upon rejection, even in the event of HRR, right?)

chris-wood:
cc @martinthomson, too, since he had thoughts on this

draft-ietf-tls-esni.md
davidben: It does. The point of my suggestion with the bullet points was instead to have the one sentence apply to both cases.

chris-wood: Gotcha. In my reading this section it seemed clear to include here, but we can drop it later on if not needed.

PR #484, opened 2021-08-04
HPKE isn't draft-08 any more

sftcd:


davidben (approved):
Hopefully HPKE is done enough that we no longer need to pin a draft version.

chris-wood (approved):

PR #485, opened 2021-08-04
note experiment

sftcd:
Seems useful to note that -13 should be around for a while but still might change later

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -59,7 +59,9 @@ encrypting a ClientHello message under a server public key.
 
 DISCLAIMER: This draft is work-in-progress and has not yet seen significant (or
 really any) security analysis. It should not be used as a basis for building
-production systems.
+production systems. As of the time of publication, this draft is being used 
+for ongoing experimentation and so may remain stable for some time. Changes
+are to be expected though once those experiments and analyses are completed.
 
 Although TLS 1.3 {{!RFC8446}} encrypts most of the handshake, including the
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -59,7 +59,9 @@ encrypting a ClientHello message under a server public key.
 
 DISCLAIMER: This draft is work-in-progress and has not yet seen significant (or
 really any) security analysis. It should not be used as a basis for building
-production systems.
+production systems. As of the time of publication, this draft is being used 
+for ongoing experimentation and so may remain stable for some time. Changes
+are to be expected though once those experiments and analyses are completed.
```
chris-wood, line 64:
```suggestion
production systems. This published version of the draft has been designated
an "implementation draft" for testing and interop purposes.
```

chris-wood (approved):
How about this simplification based on the outcome from 111? I think it's clear that future changes may occur, so I just dropped that.

sftcd:
On 05/08/2021 22:56, Christopher Wood wrote:
> @chris-wood approved this pull request.
> 
> How about this simplification based on the outcome from 111? I think it's clear that future changes may occur, so I just dropped that.
> 
>> +production systems. As of the time of publication, this draft is being used
> +for ongoing experimentation and so may remain stable for some time. Changes
> +are to be expected though once those experiments and analyses are completed.
> 
> ```suggestion
> production systems. This published version of the draft has been designated
> an "implementation draft" for testing and interop purposes.

Yeah that's fine.

Ta,
S.
> ```
>

PR #487, opened 2021-08-04
note potential config_id collisions

sftcd:
MUST->SHOULD for not publishing colliding config_id values.

This is a little more than editorial maybe but later text in any case seems to handle config_id collisions so given the MUST here is really on DNS provisioning code and not ECH code anyway, I'll claim it's editorial:-) I'd also be fine with s/MUST/should/ if that were thought better.

I also suggest deleting a sentence that looks like it no longer belongs. ("This process...") It could be that ought be shifted to later where trial decryption is discussed, not sure.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -239,12 +239,12 @@ The `ECHConfigContents` structure contains the following fields:
 key_config
 : A `HpkeKeyConfig` structure carrying the configuration information associated
 with the HPKE public key. Note that this structure contains the `config_id`
-field, which applies to the entire ECHConfigContents. Sites MUST NOT publish
+field, which applies to the entire ECHConfigContents. Sites SHOULD NOT publish
 two different `ECHConfigContents` values with the same `HpkeKeyConfig` value.
 The RECOMMENDED technique for choosing `config_id` is to choose a random byte.
-This process is repeated if this config_id matches that of any valid ECHConfig,
-which could include any ECHConfig that has been recently removed from active
-use.
+This could lead to `config_id` collisions, especially if a backend server
+is provided with ECHConfig values by more than one client-facing server and
+if each of those support multiple KEMs, hence the SHOULD above.
 
 maximum_name_length
 : The longest name of a backend server, if known. If not known, this value can


draft-ietf-tls-esni.md
```diff
@@ -239,12 +239,12 @@ The `ECHConfigContents` structure contains the following fields:
 key_config
 : A `HpkeKeyConfig` structure carrying the configuration information associated
 with the HPKE public key. Note that this structure contains the `config_id`
-field, which applies to the entire ECHConfigContents. Sites MUST NOT publish
+field, which applies to the entire ECHConfigContents. Sites SHOULD NOT publish
 two different `ECHConfigContents` values with the same `HpkeKeyConfig` value.
 The RECOMMENDED technique for choosing `config_id` is to choose a random byte.
-This process is repeated if this config_id matches that of any valid ECHConfig,
-which could include any ECHConfig that has been recently removed from active
-use.
```
davidben, line 247: This sentence is important to the recommendation, otherwise it doesn't actually avoid collisions.

draft-ietf-tls-esni.md
```diff
@@ -239,12 +239,12 @@ The `ECHConfigContents` structure contains the following fields:
...
+This could lead to `config_id` collisions, especially if a backend server
+is provided with ECHConfig values by more than one client-facing server and
+if each of those support multiple KEMs, hence the SHOULD above.
```
davidben, line 247: I don't think this example is right. If a backend server is supported by multiple client-facing servers (A and B), they need to set up their DNS properly, otherwise the connection will fail. For instance, multiple HTTPS RRs, one A's IP + A's ECHConfigList, and another with B's IP + B's  ECHConfigList.

Indeed this is why we switched from the old ECH DNS record: we got feedback from CDN operators that IPs and ECHConfigLists need to be correlated to avoid multi-CDN problems.

davidben (commented):
Agreed on MUST -> SHOULD, since it's just a performance thing. If you have a collision, trial decryption would repair it. It's just more server load. But I don't think the other two changes are right.

draft-ietf-tls-esni.md
sftcd: I'm not clear how CDN-A knows that CDN-B (also used by the backend) will in future not collide. So while CDN-A can repeat the process as often as it wants, and the ECHConfig value from CDN-A won't collide with one another, the backend can still get a collision. Or where am I wrong?

draft-ietf-tls-esni.md
sftcd: "They need to setup their DNS properly" is the issue - it's the backend's DNS not the client-facing server's DNS.

draft-ietf-tls-esni.md
davidben: They don't know, and don't have to know. You only hit trial decryption if, within a set of active configs *on the client-facing server*, there is a config ID collision. The backend doesn't deal with configs in the first place.

sftcd: Sure. trial decryption is fine. The issue is that the sentence I suggest removing seems to me to claim that one can avoid config_id collisions in the DNS if you iterate the random picking. But that's not the case.

sftcd: If one changed that sentence via s/any valid ECHConfig/any known ECHConfig/ it'd be correct I guess.

draft-ietf-tls-esni.md
davidben: Yes, the backend and client-facing servers do need to work out how they delegate this information. This has nothing to do with config_ids. The keys are a property of the client-facing server, one that will rotate over time, and the backend's DNS needs to somehow present those keys.

If the backend server and client-facing server share enough infrastructure (Shared Mode), maybe the client-facing server's provisioning process is able to directly configures the backend server's DNS. More likely, I expect the backend server will delegate the name with a CNAME or AliasForm HTTPS record.

This is the same sort of deployment decisions folks already go through with A/AAAA records.

sftcd: Sorry, but doesn't the protocol need to support the case where the split-mode backend's DNS operator is independent of all client-facing servers? If not, I've a problem with that. If so, then we can get config_id collisions, even if those aren't much of a problem. (We are after all only arguing here about the example to illustrate a SHOULD so it's not the end of the world:-)

draft-ietf-tls-esni.md
davidben: The DNS isn't the scope we care about for collisions in the first place. It sounds like there's some misunderstanding here as to what the config_id is even for. I agree the current text is not right. However, this PR does not fix it. "Any known ECHConfig" is also weird after you've rotated 256 times.

Let's take a step back here: we're encrypting a message for the client-facing server, which then dispatches to a backend server. An ECHConfig is a property of the client-facing server. It's the one generating the keys, etc. In fact, the backend server may not even have the key.

In the DNS, we *typically* only need one ECHConfig per route, because the client is ultimately going to just pick one. The multiplicity in the ECHConfigList is purely so a client-facing server can simultaneously support draft-13 and the final draft, or X25519 and fancy-future-post-quantum-KEM.

However, we never design systems with unrotatable keys. That means, at some point, the client-facing server is going to generate a new a key and update DNS. DNS is cached, TTLs are lies, etc., so old DNS entries may still be used for a while after rotation.

That means the client-facing server needs to support a _window_ of ECHConfigs, perhaps the last week or so's worth, to avoid service interrupts. This window is *larger* than what's in DNS because, again, DNS only really needs the current ones.

When the client-facing server gets a ClientHello, it needs to know which of those to use. It can either trial-decrypt, or it can filter on config_id. Trial decryption means, for every config you keep around, you have to do that many HPKE operations. That is poor performance and a DoS risk. Thus, we have a config_id, enough to disambiguate however many entries the client-facing server needs to keep alive. You *can* reuse config_ids, just, if you want to avoid trial decryption, you must have retired the old one first.

That means, when a client-facing server generates new keys, it should do so in a way that avoids collisions with any existing *active* ones. Rejection sampling is one strategy. Just incrementing and cycling IDs is another. (TBH, cycling seems simpler.)

All of this is relative to one client-facing server. If you use both CDN-A and CDN-B, A and B have separate sets of active keys and separate config ID spaces. There is no need for them to coordinate because we don't care about config_id collisions. CDN-A is never going to try one of CDN-B's keys.

I'll take a stab at rewriting this.

draft-ietf-tls-esni.md
davidben: > Sorry, but doesn't the protocol need to support the case where the split-mode backend's DNS operator is independent of all client-facing servers?

That's what CNAMEs and AliasForm records are for, no? Or do you mean independent as in without any pointers to the client-facing server at all? Even without ECHConfigs, you need to point at the client-facing server's A/AAAA records, no?

>  If so, then we can get config_id collisions, even if those aren't much of a problem?

Again, collisions across client-facing servers aren't relevant to the protocol at all. See my comment above. I think you are misunderstanding what this is even doing.

draft-ietf-tls-esni.md
sftcd: I agree with all the above except the misunderstanding bit:-) The "no need for them to co-ordinate" at the end is important IMO and one reason the original "MUST not publish collisions" was wrong. s/MUST/SHOULD/ and s/valid/known active/ would be enough of a fix for me fwiw if you prefer minimalism:-)

draft-ietf-tls-esni.md
sftcd: Collisions across client-facing servers would be relevant if the MUST not publish remained. I don't believe I'm misunderstanding config_id thanks:-)

CNAMEs etc are one way of doing things and likely will be most used.  But a backend is not in a position to say to a likely much larger CDN - hey please change your config_id because that other big CDN used the same one;-) So if we want the protocol to allow the backend to choose any DNS operator (unrelated to all CDNs) then the MUST NOT was wrong but so was the assumption that one CDN could entirely avoid config_id collisions *in the DNS* via iteration at key gen time.

davidben: Ah yeah, the sentence is just wrong all around. :-) It both should be a SHOULD, not a MUST, but the SHOULD is also wrong because it should be about an individual client-facing server, and not the backend at all.

(I'm reworking that text now, which hopefully avoids this. Will upload a PR when that's done.)

sftcd:
Closed - #490 is a better fix.

PR #488, opened 2021-08-04
sensible-pickers

sftcd:
Not sure if this is needed, but I guess it would be another way to generate failures;-)

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -293,10 +293,7 @@ public_key
 
 cipher_suites
 : The list of HPKE KDF and AEAD identifier pairs clients can use for encrypting
-ClientHelloInner.  Note that clients are intended to choose one pair, and not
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -293,7 +293,10 @@ public_key
 
 cipher_suites
 : The list of HPKE KDF and AEAD identifier pairs clients can use for encrypting
-ClientHelloInner.
+ClientHelloInner.  Note that clients are intended to choose one pair, and not
+independent KDF and AEAD identifiers, that is, choosing the first KDF with the
+second AEAD is liable to fail if the provider has not pubished an ECHConfig
+with those two as a single cipher_suite pair.
```
chris-wood, line 299:
```suggestion
ClientHelloInner. See {{real-ech}} for how clients choose from this list.
```

chris-wood (approved):
Good clarification! This was already somewhat in the real-ech section, so I just replaced this with a forward pointer there.

PR #493, opened 2021-08-04
A bit more text to introduce EncodedClientHelloInner

sftcd:
Could probably be done lots of other ways but something like this seems needed to introduce EncodedClientHelloInner.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -374,8 +374,8 @@ payload
 : The serialized and encrypted ClientHelloInner structure, encrypted using HPKE
 as described in {{real-ech}}.
 
-When the client offers the "encrypted_client_hello" extension, if the payload is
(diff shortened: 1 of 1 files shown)


sftcd:
oops - ignore the 1st block of changes those are from another PR

davidben:
Nice catch! I actually think we should go further and just straight-up define the structure first. How do you feel about https://github.com/tlswg/draft-ietf-tls-esni/pull/503?

sftcd:
On 05/08/2021 18:03, David Benjamin wrote:
> Nice catch! I actually think we should go further and just
> straight-up define the structure first. How do you feel about
> https://github.com/tlswg/draft-ietf-tls-esni/pull/503?

LGTM

Cheers,
S.

>

chris-wood:
Closing in favor of #503.

PR #495, opened 2021-08-05
optional configids seems wrong, maybe random is better?

sftcd:
minor suggested change


draft-ietf-tls-esni.md
```diff
@@ -621,9 +621,10 @@ variant of ECHClientHello with the following fields:
 - `enc`, as given above; and
 - `payload`, a placeholder byte string containing L zeros.
 
-If optional configuration identifiers (see {{optional-configs}}) are used,
+If random configuration identifiers (see {{random-configs}}) are used,
 `config_id` SHOULD be set to a randomly generated byte in the first
-ClientHelloOuter and MUST be left unchanged for the second ClientHelloOuter.
+ClientHelloOuter and, in the event of HRR, MUST be left unchanged for 
+the second ClientHelloOuter.
```
chris-wood, line 627: The intent was to say that optional IDs are to be implemented as random IDs, i.e., if you don't care about IDs (hence optional), then just set them to a random byte. So I'm not inclined to take this change.

I _do_ like the clarification to include "in the event of HRR" -- can we split that out into a separate PR?

chris-wood (changes requested):

sftcd:
Hiya,

On 05/08/2021 22:50, Christopher Wood wrote:
> @chris-wood requested changes on this pull request.
> 
> 
> 
>> `config_id` SHOULD be set to a randomly generated byte in the
>> first
> -ClientHelloOuter and MUST be left unchanged for the second
> ClientHelloOuter. +ClientHelloOuter and, in the event of HRR, MUST be
> left unchanged for +the second ClientHelloOuter.
> 
> The intent was to say that optional IDs are to be implemented as
> random IDs, i.e., if you don't care about IDs (hence optional), 

For me, for a protocol element, "optional" means "not
included" whereas here we're talking about "optionally
ignoring the proposed value." (Yeah, I probably read
too much ASN.1 once, or even twice:-)

"Random" might not be right, but "optional" could maybe
briefly confuse an implementer.

> then
> just set them to a random byte. So I'm not inclined to take this
> change.
> 
> I _do_ like the clarification to include "in the event of HRR" -- can
> we split that out into a separate PR?

Sure. (Not that I know how to do that easily, but feel free
to just do it, or wait for me to figure it out tomorrow;-)

Cheers,
S.


>

cbartle891:
> For me, for a protocol element, "optional" means "not
> included" whereas here we're talking about "optionally
> ignoring the proposed value."

I agree.

chris-wood:
It's not possible to omit the config ID (it's a byte). It's only possible to not use it. Optional seems like a fine word to capture that. We could also say that it's unused or ignored? For example:

```
If `config_id` is unused as specified by an application profile or otherwise 
externally configured, `config_id` SHOULD be set to a randomly generated byte
in the first ClientHelloOuter and, in the event of HRR, MUST be left unchanged
for the second ClientHelloOuter.
```

Or something, with similar updates to the text and section titles. Would that work?

sftcd:
On 06/08/2021 13:45, Christopher Wood wrote:
> It's not possible to omit the config ID (it's a byte). It's only
> possible to not use it. Optional seems like a fine word to capture
> that.

Interesting, (but unimportant:-) my take on those facts
is the opposite.

> We could also say that it's unused or ignored? For example:
> 
> ``` If `config_id` is unused as specified by an application profile
> or otherwise externally configured, `config_id` SHOULD be set to a
> randomly generated byte in the first ClientHelloOuter and, in the
> event of HRR, MUST be left unchanged for the second
> ClientHelloOuter. ```
> 
> Or something, with similar updates to the text and section titles.
> Would that work?

For me: "ignored" is ok, "unused" is as odd as "optional."

Cheers,
S.

chris-wood:
@sftcd updated the proposal above to use "ignored". Does that work?

sftcd:
On 06/08/2021 17:17, Christopher Wood wrote:
> @sftcd updated the proposal above to use "ignored". Does that work?

Sure. Thanks,
S

>

chris-wood:
Great! Could you please update this PR to match?

sftcd:
I think I've made those changes now.

chris-wood (approved):

draft-ietf-tls-esni.md
```diff
@@ -637,9 +637,10 @@ variant of ECHClientHello with the following fields:
 - `enc`, as given above; and
 - `payload`, a placeholder byte string containing L zeros.
 
-If optional configuration identifiers (see {{optional-configs}}) are used,
+If configuration identifiers (see {{ignored-configs}}) are to be ignored,
 `config_id` SHOULD be set to a randomly generated byte in the first
-ClientHelloOuter and MUST be left unchanged for the second ClientHelloOuter.
+ClientHelloOuter and, in the event of HRR, MUST be left unchanged for
```
chris-wood, line 642:
```suggestion
ClientHelloOuter and, in the event of HRR, MUST be left unchanged for
```

draft-ietf-tls-esni.md
```diff
@@ -1249,22 +1250,22 @@ for extended periods of time, e.g., using per-client ECHConfig structures
...
 identifiers may exacerbate DoS attacks. Specifically, an adversary may send
 malicious ClientHello messages, i.e., those which will not decrypt with any
 known ECH key, in order to force wasteful decryption. Servers that support this
 feature should, for example, implement some form of rate limiting mechanism to
-limit the damage caused by such attacks.
+limit the potential damage caused by such attacks.
 
-Unless specified by the application using (D)TLS or externally configured on
-both sides, implementations MUST NOT use this mode.
+Unless specified by the application using (D)TLS or externally configured,
```
chris-wood, line 1267:
```suggestion
Unless specified by the application using (D)TLS or externally configured,
```

PR #496, opened 2021-08-05
Clarify cardinality via the use of the (non-redudant:-) length field

sftcd:
The goal here is really to provide more clarify about cardinality, if there's a better way or place to do that, that'd be fine.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -214,7 +214,7 @@ The ECH configuration is defined by the following `ECHConfig` structure.
         uint16 version;
         uint16 length;
         select (ECHConfig.version) {
-          case 0xfe0c: ECHConfigContents contents;
+          case 0xfe0d: ECHConfigContents contents;
         }
     } ECHConfig;
 ~~~~
@@ -331,7 +331,7 @@ ClientHelloInner.
 
 ~~~
     enum {
-       encrypted_client_hello(0xfe0c), (65535)
+       encrypted_client_hello(0xfe0d), (65535)
     } ExtensionType;
 ~~~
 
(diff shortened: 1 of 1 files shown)


ekr:
I don't believe this is necessary. This idiom is well understood in TLS and this just appears to be excess verbiage.

draft-ietf-tls-esni.md
```diff
@@ -228,7 +228,13 @@ draft-08, the version is the same as the code point for the
...
 
 length
-: The length, in bytes, of the next field.
+: The length, in bytes, of the next field. One primary method of
+distributing ECHConfig values is via the DNS
+{{!HTTPS-RR=I-D.ietf-dnsop-svcb-https}} where they may be found in HTTPS or
+SVCB records that can (and will commonly) contain a list of ECHConfig values
+using the ECHConfigList syntax defined below.  This length field allows
+implementations to skip over the elements in such a list where they cannot
+parse the specific version of ECHConfig.
```
chris-wood, line 237:
```suggestion
: The length, in bytes, of the next field. This length field allows
implementations to skip over the elements in such a list where they cannot
parse the specific version of ECHConfig.
```

chris-wood (approved):

sftcd:
Hiya,

(As with all these PRs, I'm ok if they're not accepted...)

On 05/08/2021 22:48, Christopher Wood wrote:
> @chris-wood approved this pull request.
> 
> 
> 
>> +: The length, in bytes, of the next field. One primary method of
> +distributing ECHConfig values is via the DNS
> +{{!HTTPS-RR=I-D.ietf-dnsop-svcb-https}} where they may be found in HTTPS or
> +SVCB records that can (and will commonly) contain a list of ECHConfig values
> +using the ECHConfigList syntax defined below.  This length field allows
> +implementations to skip over the elements in such a list where they cannot
> +parse the specific version of ECHConfig.
> 
> ```suggestion
> : The length, in bytes, of the next field. This length field allows
> implementations to skip over the elements in such a list where they cannot
> parse the specific version of ECHConfig.
> ```

For me, that is better than the original, but I do think we
ought state the expected cardinality that implementers need
to handle somewhere. I reckon it's better to not assume
implementers are very familiar with SVCB as for libraries at
least, the DNS stuff happens elsewhere. They will have to
figure that out eventually, but better to say it somewhere
explicitly I think, or we may face some important code base
that doesn't handle >1 public key when it ought.

Cheers,
S.

PR #497, opened 2021-08-05
Try clarify an odd MUST in 6.1.5

sftcd:
6.1.5 says "This message MUST be syntactically valid." I'm not entirely sure what was intended there, so took a stab at clarifying. In any case as-is that sentence doesn't really seem to say anything useful.

My suggestion assumes that enc is included as part of the ECH in the AAD used with HRR. Given we omit enc from the 2nd ECH as-sent that seems a bit wrong so may need a few more words.

Note that the next bullet talks about using the "emtpy string for enc" - not sure if "empty string" precise enough but if we omit enc from both AAD and ECH (during HRR) we may need align how we describe things.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -757,10 +757,10 @@ ClientHello message as follows:
 1. It constructs EncodedClientHelloInner as described in {{encoding-inner}}.
 
 1. It constructs a second partial ClientHelloOuterAAD. This MUST be constructed
-   as described in {{encrypting-clienthello}} with an empty `enc` field. The
-   extensions MAY be copied from the original ClientHelloOuter unmodified, or
-   omitted. If not sensitive, the client MAY copy updated extensions from the
-   second ClientHelloInner for compression.
+   afresh as described in {{encrypting-clienthello}}. The extensions MAY be copied
+   from the original ClientHelloOuter unmodified, or omitted. If not sensitive,
+   the client MAY copy updated extensions from the second ClientHelloInner for
+   compression.
 
 1. It encrypts EncodedClientHelloInner as described in
    {{encrypting-clienthello}}, using the second partial ClientHelloOuterAAD, to
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -214,7 +214,7 @@ The ECH configuration is defined by the following `ECHConfig` structure.
         uint16 version;
         uint16 length;
         select (ECHConfig.version) {
-          case 0xfe0c: ECHConfigContents contents;
+          case 0xfe0d: ECHConfigContents contents;
         }
     } ECHConfig;
 ~~~~
@@ -331,7 +331,7 @@ ClientHelloInner.
 
 ~~~
     enum {
-       encrypted_client_hello(0xfe0c), (65535)
+       encrypted_client_hello(0xfe0d), (65535)
     } ExtensionType;
 ~~~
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -756,10 +756,11 @@ ClientHello message as follows:
 
 1. It constructs EncodedClientHelloInner as described in {{encoding-inner}}.
 
-1. It constructs a second partial ClientHelloOuterAAD message. This message MUST
-   be syntactically valid. The extensions MAY be copied from the original
-   ClientHelloOuter unmodified, or omitted. If not sensitive, the client MAY
-   copy updated extensions from the second ClientHelloInner for compression.
+1. It constructs a second partial ClientHelloOuterAAD. This MUST be constructed
+   afresh as described in {{encrypting-clienthello}}. The extensions MAY be copied
```
davidben, line 760: This section doesn't even talk about constructing ClientHelloOuterAAD. Did you mean to cite something else? Though I'm not sure there is anything useful to cite here. I avoided citing {{real-ech}} when reorganizing this because that contains a ton of text specific to the first ClientHello.

"MUST be constructed afresh" also does not seem particularly meaningful or accurate. Ultimately, the ClientHelloOuter wrapper is entirely pointless here. Both sides know it will be ignored. Some clients will find it easiest to apply RFC8446 rules (reuse some code), while others cannot because whether HelloRetryRequest will consistently apply to both depends on complex rules we never managed to write down. (This was the crux of the HelloRetryRequest fuss.) Since we didn't want to define a whole new message without all the extra baggage, this section just says to do whatever. And then maybe we'll learn something about interop needs during experimentation.

davidben (commented):
> Note that the next bullet talks about using the "emtpy string for enc" - not sure if "empty string" precise enough

There have been so many iterations over the how to say "empty string". I think this is where we settled last time? All the variations seem perfectly understandable to me. Can we please move on?

> but if we omit enc from both AAD and ECH (during HRR) we may need align how we describe things.

Since the AAD and ClientHelloOuter are both constructed from the same thing, they'll both always be empty or non-empty.

draft-ietf-tls-esni.md
sftcd: So what was the intent of the original sentence? Maybe just deleting it would be fine, depending.

davidben: It probably can be dropped. It was mostly just "the other side should be able to parse it". We're in this weird situation where we're producing a ClientHelloOuter2 that may well not be consistent with ClientHelloOuter1 + HelloRetryRequest. But both sides have agreed it's meaningless... we're only sending it because I guess we haven't defined a better message.

There is probably a less goofy design buried in here somewhere. :-)

chris-wood: I propose we park this for now and file an issue to ask ourselves if we want a new message for this second CHOuter. @davidben or @sftcd, would one of you mind filing that issue?

sftcd: Ah - the "both sides have agreed it's meaningless" finally enlightened me - what the original odd MUST meant is "the client hello outer2 is mostly useless, but MUST be well formed even so". I'm fine if we park that, just delete the odd sentence or replace with whatever the editors' think better.

sftcd:
Closed but opened issue #515 so we can re-visit later.

draft-ietf-tls-esni.md
davidben: Parking SGTM. I am also okay deleting the sentence. I agree with you it's not really meaningful. I also agree this whole situation is confusing and silly. :-)

PR #498, opened 2021-08-05
retry configs: tweak about subsequent connections 

sftcd:
When retry-configs are defined in section 5 we say they are for "subsequent connection attempts" but in 6.1.6 we have a MUST NOT for using those for "subsequent connections." That seems inconsistent and maybe even wrong if one thinks about how e.g. wget can crawl web pages, or how applications might use libcurl, so this change tries to fix that. Not sure if short duration is the right way to phrase it though.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -795,11 +795,12 @@ If at least one of the values contains a version supported by the client, it can
 regard the ECH keys as securely replaced by the server. It SHOULD retry the
 handshake with a new transport connection, using the retry configurations
 supplied by the server. The retry configurations may only be applied to the
-retry connection. The client MUST continue to use the previously-advertised
-configurations for subsequent connections. This avoids introducing pinning
-concerns or a tracking vector, should a malicious server present
-client-specific retry configurations in order to identify the client in a
-subsequent ECH handshake.
+retry connection. The client MUST only use a retry configuration for subsequent
+connections for a short (client-defined) duration and MUST revert to use the
+previously-advertised configurations after that duration has passed.  This
+avoids introducing pinning concerns or a tracking vector, should a malicious
+server present client-specific retry configurations in order to identify the
+client in a subsequent ECH handshake.
 
 If none of the values provided in "retry_configs" contains a supported version,
 the client can regard ECH as securely disabled by the server. As below, it
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -214,7 +214,7 @@ The ECH configuration is defined by the following `ECHConfig` structure.
         uint16 version;
         uint16 length;
         select (ECHConfig.version) {
-          case 0xfe0c: ECHConfigContents contents;
+          case 0xfe0d: ECHConfigContents contents;
         }
     } ECHConfig;
 ~~~~
@@ -331,7 +331,7 @@ ClientHelloInner.
 
 ~~~
     enum {
-       encrypted_client_hello(0xfe0c), (65535)
+       encrypted_client_hello(0xfe0d), (65535)
     } ExtensionType;
 ~~~
 
@@ -374,8 +374,8 @@ payload
 : The serialized and encrypted ClientHelloInner structure, encrypted using HPKE
 as described in {{real-ech}}.
 
-When the client offers the "encrypted_client_hello" extension, if the payload
-is the outer variant, then the server MAY include an "encrypted_client_hello"
+When the client offers the "encrypted_client_hello" extension, if the payload is
+the `outer` variant, then the server MAY include an "encrypted_client_hello"
 extension in its EncryptedExtensions message with the following payload:
 
 ~~~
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -795,11 +795,12 @@ If at least one of the values contains a version supported by the client, it can
...
 handshake with a new transport connection, using the retry configurations
 supplied by the server. The retry configurations may only be applied to the
-retry connection. The client MUST continue to use the previously-advertised
-configurations for subsequent connections. This avoids introducing pinning
-concerns or a tracking vector, should a malicious server present
-client-specific retry configurations in order to identify the client in a
-subsequent ECH handshake.
+retry connection. The client MUST only use a retry configuration for subsequent
+connections for a short (client-defined) duration and MUST revert to use the
+previously-advertised configurations after that duration has passed.  This
```
davidben, line 800: Using a duration across the entire client application would introduce exactly the tracking vector the text was written to prevent. This allows information to flow across application contexts if, during that duration, two contexts need to talk to the same origin.

The intent of this text was that you do it just for exactly the one high-level connection attempt (the client application likely has a "connect to server with these parameters" procedure somewhere) and nothing else. That is, you use it _exactly_ once, and within the same application context.

That is, "subsequent connections" refers to the ones after the retry connection.

sftcd: - The existing uses of "subsequent" are still conflicting.
- I get the tracking issue, but wonder what'd we expect wget to do if asked to crawl a site? It seems wrong to expect it to try/fail/try-again for every URL and better to use a retry_config that worked for the all the requests corresponding to one command line invocation.
- I agree "duration" isn't great though.

davidben: Yup, agreed that the "subsequent" is unclear. Sorry, I was just trying to explain the intent, so the PR could rephrase it better.

> It seems wrong to expect it to try/fail/try-again for every URL and better to use a retry_config that worked for the all the requests corresponding to one command line invocation.

HTTP connection reuse would smooth it over (which needs to define its tracking vector implications elsewhere) but, yes, there is definitely room to be more permissive here. I wrote this bit originally, and my aim was to write down the simplest thing possible because:

* At the time, we hadn't resolved the multi-CDN issue, so it was unclear how a sticky signal would work. Now we have, so this is possible to resolve, if we want to. (It's only good for those cached DNS entries.)
* I didn't want to go down a huge rabbithole to define how stickiness would interact with the client application's correlation goals, at least not when introducing an error correction case. As we saw from the other discussion and the introduction of a tracking vector in this PR, tracking is a complex domain-specific problem and people tend to get very confused when you try to actually define it. 😄

I'm not opposed to defining a more sticky option, though it's no longer an editorial change and we'd include to define solutions to the above problems. (Given it's a lot more fuss, my inclination at the time was to leave it alone until/unless we learn it's a problem.)

sftcd: So how about saying something like "clients MUST NOT continue to use an ECHConfig from a set of retry_configs for extended periods - ideally a client would only use such an ECHConfig once and ensure the ensuing TLS session is then used for multiple application messages. Clients might however establish multiple TLS sessions using one such ECHConfig

sftcd: Oh and btw, it's fair to say part of this isn't editorial and so should just be punted for later.

draft-ietf-tls-esni.md
```diff
@@ -795,11 +795,12 @@ If at least one of the values contains a version supported by the client, it can
...
+avoids introducing pinning concerns or a tracking vector, should a malicious
+server present client-specific retry configurations in order to identify the
+client in a subsequent ECH handshake.
```
chris-wood, line 803:
```suggestion
retry connection. The client MUST NOT use retry configurations for connections
beyond the retry. This avoids introducing pinning concerns or a tracking 
vector, should a malicious server present client-specific retry configurations
in order to identify the client in a subsequent ECH handshake.
```

How about this proposal which flips the text around somewhat and avoids the possible confusion around "subsequent" connections?

chris-wood: cc @davidben @sftcd

chris-wood (approved):

draft-ietf-tls-esni.md
```diff
@@ -795,11 +795,10 @@ If at least one of the values contains a version supported by the client, it can
 regard the ECH keys as securely replaced by the server. It SHOULD retry the
 handshake with a new transport connection, using the retry configurations
 supplied by the server. The retry configurations may only be applied to the
-retry connection. The client MUST continue to use the previously-advertised
-configurations for subsequent connections. This avoids introducing pinning
-concerns or a tracking vector, should a malicious server present
-client-specific retry configurations in order to identify the client in a
-subsequent ECH handshake.
+retry connection. The client MUST NOT use retry configurations for connections
+beyond the retry. This avoids introducing pinning concerns or a tracking
```
chris-wood, line 799:
```suggestion
beyond the retry. This avoids introducing pinning concerns or a tracking
```

PR #500, opened 2021-08-05
ech_retry_requested is a gonner

sftcd:
I'm guessing this is almost a typo.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -214,7 +214,7 @@ The ECH configuration is defined by the following `ECHConfig` structure.
         uint16 version;
         uint16 length;
         select (ECHConfig.version) {
-          case 0xfe0c: ECHConfigContents contents;
+          case 0xfe0d: ECHConfigContents contents;
         }
     } ECHConfig;
 ~~~~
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -1112,7 +1112,7 @@ loses its ECH keys, or if a deployment of ECH must be rolled back on the server.
 
 The retry mechanism repairs inconsistencies, provided the server is
 authoritative for the public name. If server and advertised keys mismatch, the
-server will respond with ech_retry_requested. If the server does not understand
+server will respond with "retry_configs". If the server does not understand
```
davidben, line 1115: Nice catch! Perhaps:

> the server will reject ECH and respond with "retry_configs"

(I think, at the time, ech_retry_requested was part of the accept/reject signal.)

sftcd: Either fix is fine by me.

chris-wood:
```suggestion
server will reject ECH and respond with "retry_configs". If the server does
not understand
```

chris-wood (approved):

PR #502, opened 2021-08-05
not sure if omitting CH in IANA registry for ech_outer_extensions is right or not

sftcd:
I don't know and pretty much don't care, but just in case this is an error... One could argue that "CH" should be set of ech_outer_extensions. I'm fine if this is just closed without discussion if it's not an error.


davidben:
I think omitting it is better because it's not allowed in any actual ClientHello message. But this depends on whether you believe EncodedClientHelloInner counts as a ClientHello, so either seems defensible. :-)

chris-wood:
My impression is that we consider EncodedClientHelloInner a transient representation of a ClientHello, but not an actual handshake message, so I think we can safely omit this.

PR #508, opened 2021-08-05
Rework Handshaking with ClientHelloInner to remove conflicts and...

pkelsey:
…eliminate redundancies

- Avoid restatement by relying on the itemization of conditions that
  constitute ECH rejection given in 6.1.4.

- The prior text described aborting the connection with an alert,
  *then* processing retry_configs, which processing could then lead to
  additional instructions to abort the (aborted) connection with an
  alert (and possibly a different alert than at first).

- Remove the use of the "illegal_parameter" alert in the case of "if
  the field contains any other value", which apparently referred to
  the retry_config version fields.  First, this conflicts with the
  requirement that the connection be aborted with ech_required.
  Second, it's not clear what "any other value" is supposed to mean as
  this determination follows determining whether the members of the
  set of values are either "supported" or "unsupported".  Perhaps it
  meant a value not considered valid as of the specification version
  the client was implemented according to?  Certainly, in the case
  where there is at least one supported version found but also at
  least one such invalid value, it would not be desriable to use an
  illegal_parameter alert, as this situation will naturally arise when
  servers deploying a new ECH version alongside an existing one serve
  clients that have not yet been upgraded.  It's also unclear what
  value client discrimination of unsupported-recognized-as-valid from
  unsupported-not-recognized-as-valid provides.

- Linearize the presentation.

- Fix the reference to the obsolete "ech_retry_request". (See also
  tlswg/draft-ietf-tls-esni#500).

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -811,7 +811,7 @@ otherwise. If an earlier TLS version was negotiated, the client MUST NOT enable
 the False Start optimization {{RFC7918}} for this handshake. If both
 authentication and the handshake complete successfully, the client MUST perform
 the processing described below then abort the connection with an "ech_required"
-alert.
+alert before sending any application data to the server.
 
 If the server provided "retry_configs" and if at least one of the values
 contains a version supported by the client, the client can regard the ECH keys
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -812,8 +812,8 @@ the client can regard ECH as securely disabled by the server.
 
 If an earlier TLS version was negotiated, the client can regard ECH as
 securely disabled by the server, the client MUST NOT enable the False Start
-optimization {{RFC7918}} for this handshake, and the client MUST abort the
-connection with an "ech_required" alert.
+optimization {{RFC7918}} for this handshake (it may be enabled for the retry
+connection), and the client MUST abort the connection with an "ech_required" alert.
 
 If according to the above, the client regards ECH as securely disabled
 by the server, it SHOULD retry the handshake with a new transport
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -787,36 +787,29 @@ return a failure to the calling application. It MUST NOT use the retry
 configurations. It MUST NOT treat this as a secure signal to
 disable ECH.
 
-Otherwise, if both authentication and the handshake complete successfully, the
-client then processes the server's response as follows.
-
 If the server supplied an "encrypted_client_hello" extension in its
 EncryptedExtensions message, the client MUST check that it is syntactically
 valid and the client MUST abort the connection with a "decode_error" alert
-otherwise.  If the extension is valid, the client MUST process the
-"retry_configs" from the extension as described below and then abort the
-connection with an "ech_required" alert.
+otherwise. If an earlier TLS version was negotiated, the client MUST NOT enable
+the False Start optimization {{RFC7918}} for this handshake. If both
+authentication and the handshake complete successfully, the client MUST perform
+the processing described below then abort the connection with an "ech_required"
+alert.
 
-If at least one of the values contains a version supported by the client, it can
-regard the ECH keys as securely replaced by the server. It SHOULD retry the
-handshake with a new transport connection, using the retry configurations
-supplied by the server. The retry configurations may only be applied to the
-retry connection. The client MUST continue to use the previously-advertised
+If the server provided "retry_configs" and if at least one of the values
+contains a version supported by the client, the client can regard the ECH keys
+as securely replaced by the server. It SHOULD retry the handshake with a new
+transport connection, using the retry configurations supplied by the
+server. The retry configurations may only be applied to the retry
+connection. The client MUST continue to use the previously-advertised
 configurations for subsequent connections. This avoids introducing pinning
 concerns or a tracking vector, should a malicious server present
 client-specific retry configurations in order to identify the client in a
 subsequent ECH handshake.
 
 If none of the values provided in "retry_configs" contains a supported version,
-the client can regard ECH as securely disabled by the server.
-
-If an earlier TLS version was negotiated, the client can regard ECH as
-securely disabled by the server, the client MUST NOT enable the False Start
-optimization {{RFC7918}} for this handshake (it may be enabled for the retry
-connection), and the client MUST abort the connection with an "ech_required" alert.
-
-If according to the above, the client regards ECH as securely disabled
-by the server, it SHOULD retry the handshake with a new transport
+or an earlier TLS version was negotiated, the client can regard ECH as securely
+disabled by the server, and it SHOULD retry the handshake with a new transport
 connection and ECH disabled.
 
 Clients SHOULD implement a limit on retries caused by receipt of "retry_configs"
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -788,12 +788,14 @@ configurations. It MUST NOT treat this as a secure signal to
 disable ECH.
 
 Otherwise, if both authentication and the handshake complete successfully, the
-client MUST perform the processing described below then abort the connection
-with an "ech_required" alert.
+client then processes the server's response as follows.
 
 If the server supplied an "encrypted_client_hello" extension in its
-EncryptedExtensions message, the "retry_configs" field is processed as
-follows.
+EncryptedExtensions message, the client MUST check that it is syntactically
+valid and the client MUST abort the connection with a "decode_error" alert
+otherwise.  If the extension is valid, the client MUST process the
+"retry_configs" from the extension as described below and then abort the
+connection with an "ech_required" alert.
 
 If at least one of the values contains a version supported by the client, it can
 regard the ECH keys as securely replaced by the server. It SHOULD retry the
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -784,12 +784,16 @@ If the server rejects ECH, the client proceeds with the handshake,
...
 Otherwise, if both authentication and the handshake complete successfully, the
-client MUST abort the connection with an "ech_required" alert. It then
-processes the "retry_configs" field from the server's "encrypted_client_hello"
-extension.
+client MUST perform the processing described below then abort the connection
+with an "ech_required" alert.
+
+If the server supplied an "encrypted_client_hello" extension in its
+EncryptedExtensions message, the "retry_configs" field is processed as
+follows.
```
chris-wood, line 796:
```suggestion
client then processes the server's response as follows. If the server supplied
an "encrypted_client_hello" extension in its EncryptedExtensions message, the
client MUST check that it is syntactically valid and the client MUST abort the
connection with a "decode_error" alert otherwise. If the extension is valid,
the client processes the "retry_configs" from the extension as described below
and then abort the connection with an "ech_required" alert.
```

The intent of the differentiator was to fail early and loudly in the presence of server bugs, though I concur that "illegal_parameter" was perhaps not the best alert. I tried to fix that with this suggestion. What do you think?

pkelsey: The decode_error alert language makes sense and is nicely consistent with the requirements for processing a GREASE response.  However, the relocation of the  'abort with ech_required' language ties it to the extension present case and makes it no longer apply to the earlier-TLS-version case down below.  I'm going to take a stab at either trying to refactor this a bit with your new decode_error language on board.

pkelsey: It seemed the only reasonable way to resolve things was to duplicate the ech_required language in the earler-TLS-version part.  I pushed a new commit to that effect.

chris-wood (approved):

draft-ietf-tls-esni.md
chris-wood: The latest change LGTM 👍

draft-ietf-tls-esni.md
```diff
@@ -784,12 +784,18 @@ If the server rejects ECH, the client proceeds with the handshake,
...
 return a failure to the calling application. It MUST NOT use the retry
-configurations.
+configurations. It MUST NOT treat this as a secure signal to
+disable ECH.
 
 Otherwise, if both authentication and the handshake complete successfully, the
-client MUST abort the connection with an "ech_required" alert. It then
-processes the "retry_configs" field from the server's "encrypted_client_hello"
-extension.
+client then processes the server's response as follows.
```
chris-wood, line 791: I understand the intent here, but this seems somewhat out of order. The checks below happen _before_ the handshake completes, but this text suggests they happen afterwards. Perhaps we can rewrite this as follows?

```
Otherwise, in authenticating the server and completing the handshake, 
the client then processes the server's response as follows.
```

draft-ietf-tls-esni.md
```diff
@@ -802,31 +808,21 @@ client-specific retry configurations in order to identify the client in a
...
-client MUST NOT enable the False Start optimization {{RFC7918}} for this
-handshake. If authentication or the handshake fails, the client MUST return a
-failure to the calling application. It MUST NOT treat this as a secure signal to
-disable ECH.
+the client can regard ECH as securely disabled by the server.
+
+If an earlier TLS version was negotiated, the client can regard ECH as
+securely disabled by the server, the client MUST NOT enable the False Start
+optimization {{RFC7918}} for this handshake, and the client MUST abort the
+connection with an "ech_required" alert.
```
chris-wood, line 816:
```suggestion
securely disabled by the server. If the client retries, the client MUST NOT
enable the False Start optimization {{RFC7918}} for this handshake, and the
client MUST abort the connection with an "ech_required" alert.
```

chris-wood (approved):

draft-ietf-tls-esni.md
pkelsey: Good catch - this ordering issue is one I was trying to expunge during the first pass, and it has now crept back in with the decode_error processing.  There's possibly another instance of this in the suggestion for the comment below, depending on how the question I'm asking there resolves.

draft-ietf-tls-esni.md
pkelsey: > securely disabled by the server. If the client retries, the client MUST NOT

I had been under the impression that the prohibition on RFC7918 use was for the handshake that contains the ECH rejection (that is, don't send application data into a handshake that is going to be aborted once it completes).  It's not clear why this prohibition would be conditioned on the client retrying the connection, or now whether it is meant to apply only to a retry handshake?

chris-wood: The existing text already predicated RFC7918 on whether or not the client retried. (You can't do RFC7918 if you don't retry in this instance, that is.) This suggestion is meant as an editorial improvement (but YMMV), not as a semantic one.

pkelsey: So I think we're understanding this the same way - that the point is not to plow through with RFC7918 in a connection that is being aborted due to ECH rejection, but the client is free to do so on the retry connection.  My mileage is definitely still varying on the "if the client retries", as it's really outside of the context of 'this handshake', which is what that part is about.

Would you say this would be equivalent?

```suggestion
securely disabled by the server, the client MUST NOT enable the False Start
optimization {{RFC7918}} for this handshake (it may be enabled for the retry
connection), and the client MUST abort the connection with an "ech_required" alert.
```


I'm going to push a commit shortly that I think irons out the ordering issue you flagged.

chris-wood: Yep! (And reading that pointed out the confusion I introduced, so, thanks for clarifying.)

pkelsey: The new commit I just pushed I think incorporates everything we've talked about (and is actually a bit more compact in the end) except that parenthetical about RFC7918 on the retry connection was still an open question when I did the commit.  If you think the new commit is otherwise good and that parenthetical will clarify it further, I'm all for it.

draft-ietf-tls-esni.md
```diff
@@ -784,49 +784,38 @@ If the server rejects ECH, the client proceeds with the handshake,
...
+disable ECH.
+
+If the server supplied an "encrypted_client_hello" extension in its
+EncryptedExtensions message, the client MUST check that it is syntactically
+valid and the client MUST abort the connection with a "decode_error" alert
+otherwise. If an earlier TLS version was negotiated, the client MUST NOT enable
+the False Start optimization {{RFC7918}} for this handshake. If both
+authentication and the handshake complete successfully, the client MUST perform
+the processing described below then abort the connection with an "ech_required"
+alert.
```
chris-wood, line 797:
```suggestion
alert before sending any application data to the server.
```

draft-ietf-tls-esni.md
```diff
@@ -784,49 +784,38 @@ If the server rejects ECH, the client proceeds with the handshake,
...
+
+If the server provided "retry_configs" and if at least one of the values
+contains a version supported by the client, the client can regard the ECH keys
+as securely replaced by the server. It SHOULD retry the handshake with a new
+transport connection, using the retry configurations supplied by the
+server. The retry configurations may only be applied to the retry
+connection. The client MUST continue to use the previously-advertised
```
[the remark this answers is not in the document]
pkelsey: Looks like I missed a merge...

draft-ietf-tls-esni.md
```diff
@@ -801,48 +801,37 @@ If the server rejects ECH, the client proceeds with the handshake,
...
 
-Otherwise, when the handshake completes successfully with the public name
-authenticated, the client MUST abort the connection with an "ech_required"
-alert. The client can then regard ECH as securely disabled by the server. It
-SHOULD retry the handshake with a new transport connection and ECH disabled.
+If the server supplied an "encrypted_client_hello" extension in its
+EncryptedExtensions message, the client MUST check that it is syntactically
+valid and the client MUST abort the connection with a "decode_error" alert
+otherwise. If an earlier TLS version was negotiated, the client MUST NOT enable
+the False Start optimization {{RFC7918}} for this handshake. If both
```
chris-wood, line 811: This bit seems somewhat out of place -- why include it here? Or include it at all, given the "before sending any application data to the server" clause below?

pkelsey: They are not redundant statements.  Forbidding RFC7918 prevents the client from sending application data to the server before the handshake completes.  The additional statement that you suggested ("before sending application data to the server") requires that the connection be aborted before sending any application data to the server after the handshake completes successfully.

I put the RFC7918 statement where it is as it affects what is done (or not) as the handshake progresses, so it is more concise to discuss it prior to everything that follows "If both the authentication and the handshake complete successfully...".  The requirement to validate the syntax of the encrypted_client_hello extension is co-located there under similar reasoning, which is that the syntax must be validated in order to proceed with the "after the handshake" processing that follows (although I think there is currently leeway there for an implementation to postpone the decode until after the handshake completes).

chris-wood: > The additional statement that you suggested ("before sending application data to the server") requires that the connection be aborted before sending any application data to the server after the handshake completes successfully.

I don't see how this is the case. The statement says "do not send any application data to the server," which includes, IMO, False Start data, or data that would otherwise be sent when the handshake normally completes.

pkelsey: It's the predicate that begins that sentence "If both authentication and the handshake complete successfully,..." that I think places the effect of prohibiting sending application data to the server between the completion of the handshake and the sending of the ech_required alert.

chris-wood: It seems redundant, but I can live with it.

chris-wood (approved):

PR #509, opened 2021-08-05
Clarify the document's assumption of restricted attacker location

pkelsey:


chris-wood:
This isn't accurate -- we assume the attacker can exist anywhere. Closing as a result.

pkelsey:
Do you assume the attacker can exist on both network segments in split mode?

chris-wood:
Yep, we assume the attacker can exist anywhere.

PR #511, opened 2021-08-05
Remove sentence that is redundant with the paragraph that follows

pkelsey:


cbartle891:
I can see how that sentence is redundant, but it seems like it would be helpful to implementers.

chris-wood (approved):

PR #558, opened 2023-10-09
Remove alternative designs

chris-wood:
This doesn't make sense to keep in the final document, so let's remove it.


davidben (approved):
Don't feel strongly whether it goes or stays in, but if you want to remove it, SGTM.

PR #560, opened 2023-10-09
Note behavior and give an example alongside configurations

### LINKED ISSUE #519: Anonymity set definition should include behavior
The current version of the spec mentions that: 

> Co-located servers with consistent externally visible TLS configurations, including supported versions and cipher suites, form an anonymity set.

Yet decisions a server makes during the handshake (such as which cipher suite it selects from a proposed list) can also be used to distinguish connection establishment to one server from another.  In the most extreme case, this would result in anonymity sets being restricted to servers using the same TLS implementation and configuration. I'm not sure where text describing this should go (maybe a section on anonymity set considerations? maybe just update the definition to read "Co-located servers with consistent externally visible TLS configurations, [...], and behavior...") but I think it's worth a mention in the spec.

chris-wood:
Closes #519

cc @klinvill, @davidben, @dennisjackson, @cjpatton

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -95,10 +95,11 @@ This document specifies a new TLS extension, called Encrypted Client Hello
 (ECH), that allows clients to encrypt their ClientHello to such a deployment.
 This protects the SNI and other potentially sensitive fields, such as the ALPN
 list {{?RFC7301}}. Co-located servers with consistent externally visible TLS
-configurations, including supported versions and cipher suites, form an
-anonymity set. Usage of this mechanism reveals that a client is connecting to a
-particular service provider, but does not reveal which server from the
-anonymity set terminates the connection.
+configurations and behavior, including supported versions and cipher suites and
+how they respond to incoming client connections, form an anonymity set. Usage
+of this mechanism reveals that a client is connecting to a particular service
+provider, but does not reveal which server from the anonymity set terminates
+the connection.
 
 ECH is supported in TLS 1.3 {{!RFC8446}}, DTLS 1.3 {{!RFC9147}}, and
 newer versions of the TLS and DTLS protocols.
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -96,10 +96,13 @@ This document specifies a new TLS extension, called Encrypted Client Hello
 This protects the SNI and other potentially sensitive fields, such as the ALPN
 list {{?RFC7301}}. Co-located servers with consistent externally visible TLS
 configurations and behavior, including supported versions and cipher suites and
-how they respond to incoming client connections, form an anonymity set. Usage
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -95,10 +95,11 @@ This document specifies a new TLS extension, called Encrypted Client Hello
...
 list {{?RFC7301}}. Co-located servers with consistent externally visible TLS
-configurations, including supported versions and cipher suites, form an
-anonymity set. Usage of this mechanism reveals that a client is connecting to a
-particular service provider, but does not reveal which server from the
-anonymity set terminates the connection.
+configurations and behavior, including supported versions and cipher suites and
+how they respond to incoming client connections, form an anonymity set. Usage
+of this mechanism reveals that a client is connecting to a particular service
+provider, but does not reveal which server from the anonymity set terminates
+the connection.
```
davidben, line 102: I'm not sure what you mean by "how they respond to incoming client connections". Which certificate you send is arguably how part of "how they respond" but presumably we don't mean that! I suppose it could cover other externally-visible implementation choices like division of data into record-layer boundaries. That'd be a pretty solid example... but #519 talks about cipher suite selection, and that's already covered by configurations. So I'm just very confused right now.

chris-wood: The idea here is that how a server responds, and how that response is visible to the attacker, influences the anonymity set. So on that basis I would consider what certificate they send to be covered here. Note that we don't need to be very rigorous here -- this is just the introduction and the goal is to convey the high-level goal of ECH.

chris-wood: As above, I don't think we need to be rigorous or precise here. This suggestion is a lot more text than I think is warranted for this small point.

chris-wood: @klinvill after thinking on this some more I decided to fold in a variant of your suggestion has a parenthetical. Please have another look?

klinvill (approved):
Thanks for the changes Chris, this looks good to me! I think the explicit parenthetical will help readers of this document understand that external configurations and behavior can include implementation-specific behaviors.

On a related note, I noticed that the security goals section includes a parenthetical that states: `(The set of hosts which share the same ECHConfig and TLS configuration is referred to as the anonymity set.)`. Perhaps this should be updated to reflect the text here (e.g. include "externally visible"), or better yet drop the parenthetical all together and add an explicit definition of "anonymity set". I'll open a separate issue for this though since it's a different concern from what this PR addresses.

klinvill:
> @klinvill we talk about the anonymity set in the security considerations -- we could simply point there?

Anonymity sets are discussed there, but not precisely defined. I think it might be nicer to have a canonical precise definition that could be referenced by the rest of the spec. I opened #568 to discuss this further (if warranted) and would be happy to take a stab at what such revisions would look like (as a PR).

cjpatton (approved):

PR #561, opened 2023-10-09
Cross-implementation decisions may be differentiators

### LINKED ISSUE #512: Lessen possibility of client implementation choices undermining GREASE cover
Section 10.9.4 (Do Not Stick Out) indicates plans "to deploy GREASE ECH widely enough to disincentivize differential treatment of the real ECH protocol by the network".  This implies that successfully bootstrapping ECH deployment will necessarily involve significant amounts of GREASE ECH from TLS stacks being operated in a GREASE-only mode, driven by applications that do not have a concept of how they'd configure real ECH.

Guidance on the construction of GREASE ECH is given in Section 6.2, and it involves selecting the length of the payload field based on "the size of the EncodedClientHelloInner the client would compute when offering ECH".  What is the length that the client might compute when offering (real) ECH for clients that may have no sophisticated notion of offering real ECH?  It seems there is plenty of room for implementations to arise which generate GREASE payload lengths that are susceptible to effective differentiation by heuristics simple enough to be within reach of the very lazy attacker of 10.9.4.  Perhaps such implementations use a fixed size 'prototype ClientHello' concept, or use something highly self-similar (e.g., verbatim non-ECH ClientHello contents), or [insert other debacle].  Absent additional guardrails for GREASE ECH construction, there appears to be a risk that knowledge of the extant implementations may be sufficient to provide differentiation capabilities to an attacker who employs only single-connection passive traffic analysis.

chris-wood:
Closes #512

The idea here is that some real and fake ECH differentiator mitigations may require different implementations to agree on how things are done, which is out of scope. This PR also cleans up some text in the Do Not Stick Out section.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -1449,7 +1449,7 @@ following capabilities:
 Moreover, real ECH and GREASE ECH are designed so that the following features
 do not vary:
 
-1. the code points of extensions negotiated in the clear;
+1. the code points of extensions negotiated in the clear, and their order;
 1. the length of messages; and
 1. the values of plaintext alert messages.
 
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -1445,9 +1445,13 @@ following capabilities:
 1. The attacker does not know the ECHConfigList used by the server.
 1. The attacker keeps per-connection state only. In particular, it does not
    track endpoints across connections.
-1. ECH and GREASE ECH are designed so that the following features do not vary:
-   the code points of extensions negotiated in the clear; the length of
-   messages; and the values of plaintext alert messages.
+
+Moreover, real ECH and GREASE ECH are designed so that the following features
+do not vary:
+
+1. the code points of extensions negotiated in the clear;
+1. the length of messages; and
+1. the values of plaintext alert messages.
 
 This leaves a variety of practical differentiators out-of-scope. including,
 though not limited to, the following:
@@ -1459,8 +1463,9 @@ though not limited to, the following:
 1. HRR issuance, which may depend on ECH acceptance.
 
 These can be addressed with more sophisticated implementations, but some
-mitigations require coordination between the client and server. These
-mitigations are out-of-scope for this specification.
+mitigations require coordination between the client and server, and even
+across different client and server implementations. These mitigations are
+out-of-scope for this specification.
 
 ### Maintain Forward Secrecy
 
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -1447,7 +1447,7 @@ following capabilities:
    track endpoints across connections.
 
 Moreover, real ECH and GREASE ECH are designed so that the following features
-do not vary:
+do not noticeably vary to the attacker, i.e., they are not distinguishers:
 
 1. the code points of extensions negotiated in the clear, and their order;
 1. the length of messages; and


draft-ietf-tls-esni.md
```diff
@@ -1445,9 +1445,13 @@ following capabilities:
...
    track endpoints across connections.
-1. ECH and GREASE ECH are designed so that the following features do not vary:
-   the code points of extensions negotiated in the clear; the length of
-   messages; and the values of plaintext alert messages.
+
+Moreover, real ECH and GREASE ECH are designed so that the following features
+do not vary:
+
+1. the code points of extensions negotiated in the clear;
+1. the length of messages; and
```
davidben, line 1453: This is a little interesting. The length of the ECH ciphertext only matches if the implementation accurately simulates things. The length of the encrypted certificate only matches if implementation applies the same padding between the two cases.

Not sure how best to phrase this. It's not out-of-scope per se in that we don't completely ignore it. But we also can't claim they were designed not to not vary without any qualifications.

chris-wood: This is existing text (just formatted differently), so I'd be inclined to address that in a followup PR or do nothing.

chris-wood:
@davidben would you like to see changes here, or can we merge this? @dennisjackson, I'd also appreciate your review.

draft-ietf-tls-esni.md
```diff
@@ -1445,9 +1445,13 @@ following capabilities:
...
 1. The attacker keeps per-connection state only. In particular, it does not
    track endpoints across connections.
-1. ECH and GREASE ECH are designed so that the following features do not vary:
-   the code points of extensions negotiated in the clear; the length of
-   messages; and the values of plaintext alert messages.
+
+Moreover, real ECH and GREASE ECH are designed so that the following features
+do not vary:
+
+1. the code points of extensions negotiated in the clear;
```
cjpatton, line 1452:
```suggestion
1. the code points of extensions negotiated in the clear, and their order;
```

draft-ietf-tls-esni.md
cjpatton: What about qualifying "do not very"? "do not very significantly"?

Question: Do we believe it's possible for an implementation to make real ECH computationally indistinguishable from grease ECH? Would such an implementation be non-compliant?

cjpatton (approved):

draft-ietf-tls-esni.md
```diff
@@ -1445,9 +1445,13 @@ following capabilities:
 1. The attacker does not know the ECHConfigList used by the server.
 1. The attacker keeps per-connection state only. In particular, it does not
    track endpoints across connections.
-1. ECH and GREASE ECH are designed so that the following features do not vary:
-   the code points of extensions negotiated in the clear; the length of
-   messages; and the values of plaintext alert messages.
+
+Moreover, real ECH and GREASE ECH are designed so that the following features
+do not vary:
```
chris-wood, line 1450:
```suggestion
do not noticeably vary to the attacker, i.e., they are not distinguishers:
```

draft-ietf-tls-esni.md
chris-wood: I took the suggestion to qualify "do not vary" -- @cjpatton please have a look.

cjpatton: I think it's fine as is, but my suggestion was to tack on "significantly". See suggestion above.

draft-ietf-tls-esni.md
cjpatton:
```suggestion
Moreover, real ECH and GREASE ECH are designed so that implementations can ensure the following features
do not vary significantly:
```

chris-wood: I don't really know if this is a helpful qualifier because we don't have a definition of "significantly," so I think I'd prefer to leave this off for now.

PR #562, opened 2023-10-09
Punt on new transport connection specifics

### LINKED ISSUE #524: How to retry in ECH is ambiguous
Not sure what the right answer is, or whether this belongs in this draft, or some upper layer document. (Alas, though I'm not even sure whether those upper layer documents exist... SVCB? Part of HTTP?) When the recovery flow triggers, we are supposed to "retry the handshake with a new transport connection".

The transport connection is mostly out of TLS's scope, but _somewhere_ we should decide whether this means...

* Do you connect with the exact same IP?
* Do you connect with the same SVCB record, but other IPs within that route are in scope?
* Do you connect anew and we just say, meh, this is a recovery mechanism so our aim is just to work most of the time?
* Or maybe we'll learn this is a problem and we need to jam the retry in the same connection (big protocol change, shifts some complexity from upper layers into TLS)
* What happens with proxies, which typically make DNS queries for you? Should we ever decide to integrate SVCB fetching into HTTP CONNECT, the answer here will have implications for that scheme.

Hopefully, with initial experiments, we'll understand this better. But filing this now so we don't forget to address it later.

chris-wood:
Closes #524 

As always, happy to take suggestions or alternative PRs.

cc @dennisjackson, @davidben, @martinthomson

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -837,11 +837,16 @@ If the server provided "retry_configs" and if at least one of the values
 contains a version supported by the client, the client can regard the ECH keys
 as securely replaced by the server. It SHOULD retry the handshake with a new
 transport connection, using the retry configurations supplied by the
-server. The retry configurations may only be applied to the retry
-connection. The client MUST NOT use retry configurations for connections beyond
-the retry. This avoids introducing pinning concerns or a tracking vector,
-should a malicious server present client-specific retry configurations in order
-to identify the client in a subsequent ECH handshake.
+server. (Clients can implement a new transport connection in a way that best
+suits their deployment. For example, clients can reuse the same IP address
+when establishing the new transport connection or they can choose to use a
+different IP address if provided with options from DNS. ECH does not mandate
+any specific implementation choices when establishing this new connection.)
+The retry configurations may only be applied to the retry connection. The
+client MUST NOT use retry configurations for connections beyond the retry.
+This avoids introducing pinning concerns or a tracking vector, should a
+malicious server present client-specific retry configurations in order to
+identify the client in a subsequent ECH handshake.
 
 If none of the values provided in "retry_configs" contains a supported version,
 or an earlier TLS version was negotiated, the client can regard ECH as securely


draft-ietf-tls-esni.md
```diff
@@ -837,11 +837,16 @@ If the server provided "retry_configs" and if at least one of the values
...
-connection. The client MUST NOT use retry configurations for connections beyond
-the retry. This avoids introducing pinning concerns or a tracking vector,
-should a malicious server present client-specific retry configurations in order
-to identify the client in a subsequent ECH handshake.
+server. (Clients can implement a new transport connection in a way that best
+suits their deployment. For example, clients can reuse the same IP address
+when establishing the new transport connection or they can choose to use a
+different IP address if provided with options from DNS. ECH does not mandate
+any specific implementation choices when establishing this new connection.)
+The retry configurations may only be applied to the retry connection. The
```
martinthomson, line 845:
```suggestion
The retry configurations may only be applied to the retried connection. The
```

martinthomson, line 845: Is this a MUST only?

draft-ietf-tls-esni.md
```diff
@@ -837,11 +837,16 @@ If the server provided "retry_configs" and if at least one of the values
 contains a version supported by the client, the client can regard the ECH keys
 as securely replaced by the server. It SHOULD retry the handshake with a new
 transport connection, using the retry configurations supplied by the
-server. The retry configurations may only be applied to the retry
-connection. The client MUST NOT use retry configurations for connections beyond
-the retry. This avoids introducing pinning concerns or a tracking vector,
-should a malicious server present client-specific retry configurations in order
-to identify the client in a subsequent ECH handshake.
+server. (Clients can implement a new transport connection in a way that best
```
martinthomson, line 840: This paragraph got thicc.  Do you want to turn it into three?

draft-ietf-tls-esni.md
```diff
@@ -837,11 +837,16 @@ If the server provided "retry_configs" and if at least one of the values
...
+suits their deployment. For example, clients can reuse the same IP address
+when establishing the new transport connection or they can choose to use a
+different IP address if provided with options from DNS. ECH does not mandate
+any specific implementation choices when establishing this new connection.)
+The retry configurations may only be applied to the retry connection. The
+client MUST NOT use retry configurations for connections beyond the retry.
```
martinthomson, line 846: Unrelated to this change, but if a client is making multiple connections (hello HTTP/1.1), can it use the retry configuration it learns from its first connection to guide all connections in that pool?

Let's say that the client is making these connections within the same context and so has no concerns about the connection attempts being correlated (maybe it was going to send the same cookies in all of them).  Should the client only use the retry configuration for a single connection, even knowing that it will be forced to retry multiple times, or can it propagate that information immediately?

I like that this text largely avoids 2119 language (questionable lowercase "may" and "should" usage notwithstanding), but this one comes with a "MUST NOT" that seems misguided.

chris-wood: Indeed -- unrelated, since this is existing text. I filed [this issue](https://github.com/tlswg/draft-ietf-tls-esni/issues/565) to track resolution separately.

kazuho (approved):

PR #573, opened 2023-10-17
Limit DNS label length

chris-wood:
cc @davidben, @dennisjackson, @martinthomson

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -295,7 +295,9 @@ parsable as a dot-separated sequence of LDH labels, as defined in
 : Clients SHOULD ignore the `ECHConfig` if it contains an encoded IPv4 address.
 To determine if a public_name value is an IPv4 address, clients can invoke the
 IPv4 parser algorithm in {{WHATWG-IPV4}}. It returns a value when the input is
-an IPv4 address.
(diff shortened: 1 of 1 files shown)


martinthomson (commented):
Why "SHOULD" and not "MUST"?

Personally, I wouldn't bother implementing this check. The server won't be able to obtain a valid certificate for a name that violates this constraint, or -- if it can -- you are trusting it based on a bogus root certificate.  So this doesn't really help.  It doesn't hurt either, but I'm not sure that you need defense in depth.  That suggests "MAY" instead.

draft-ietf-tls-esni.md
```diff
@@ -295,7 +295,9 @@ additionally SHOULD ignore the structure if the final LDH label either consists
 of all ASCII digits (i.e. '0' through '9') or is "0x" or "0X" followed by some,
 possibly empty, sequence of ASCII hexadecimal digits (i.e. '0' through '9', 'a'
 through 'f', and 'A' through 'F'). Thus avoids public_name values that may be
-interpreted as IPv4 literals.
+interpreted as IPv4 literals. Additionally, clients SHOULD ignore the
```
chris-wood, line 298:
```suggestion
interpreted as IPv4 literals. Additionally, clients MUST ignore the
```

chris-wood:
I'm OK with MAY (and would also be OK with a MUST, heh).

PR #577, opened 2023-10-29
Define ECHConfigExtensionType before referencing it.

jmcrawford45:
Everything else in the definition is defined before being referenced. The only other details on this field occur pretty far into the doc, so I thought it'd be nice to define it with a pointer to the relevant section with more details


chris-wood (approved):

chris-wood:
Thanks!

PR #587, opened 2023-11-15
Proposed changes to the overview section to give better justice to ECH.

taddhar:
Currently the overview section is a sentence that is actually not giving an overview of the protocol. The sentence is simply the highlight for the topologies sub-section. But the topologies sub-section is not a protocol overview. ECH is much more than just a topology. 

The proposal is to give justice to ECH and improve the overview section. 

As I am both coming at a very late state of the development of this text and this is the first time I do this review I contemplate that there could be more options. 

For example the first part of my text is a bit redundant with some of the text in the introduction, calling for whether my proposal should be moved to introductions or to rearrange the sections. 

Yet, there is no where in the text a description on all the key design ideas and maybe my selection could be completed by more ideas, e.g. list DoH and DPRIVE too in this list, which will then be calling to overlaps with introduction section.

Anyway, this is my current proposal because right now, the overview is not giving justice nor anywhere in the text to ECH.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -118,11 +118,36 @@ notation comes from {{RFC8446, Section 3}}.
 
 # Overview
 
+ECH is a protocol extension to TLS1.3 which objective is to encrypt the
+ClientHello to fulfill a number of goals (see {{goals}}), in particular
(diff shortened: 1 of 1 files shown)


ekr:
See #601

PR #589, opened 2023-11-15
Minor editorial issues and fixing metadata to stop annoying make errors.

taddhar:
- Fixing some minor editorial issues, e.g. the abuse of 'first' with no 'then' or 'second' or 'secondly' etc. 
- In X.509 don't forget the dot
- Boring make errors because of metadata issues

About the latter, I get the warning that I don't know how to fix

```
warnings.warn(
*** attributes left {"authors"=>[{"ins"=>"K. Bhargavan", "org"=>"Inria"}, {"ins"=>"V. Cheval", "org"=>"Inria"}, {"ins"=>"C. Wood", "org"=>"Cloudflare"}]}!
```

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -5,7 +5,8 @@ docname: draft-ietf-tls-esni-latest
 category: std
 
 ipr: trust200902
-area: General
+submissiontype: IETF
+area: SEC
 workgroup: tls
 keyword: Internet-Draft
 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -1424,7 +1430,7 @@ consideration when reasoning about the privacy properties that ECH provides.
 
 ECH requires encrypted DNS to be an effective privacy protection mechanism.
 However, verifying the server's identity from the Certificate message,
-particularly when using the X509 CertificateType, may result in additional
+particularly when using the X.509 CertificateType, may result in additional
```
ekr, line 1433: This is not correct. Yes, they are X.509 certificates, but the name of the enum in RFC 8446 (S 4.4.2i) ix "X509"

```suggestion
particularly when using the X509 CertificateType, may result in additional
```

ekr:
There's nothing wrong with "First"..."later". I don't think we should change these.

ekr (changes requested):

ekr:
I have fixed the metadata issue in #599. 

As noted above, I don't believe the proposed text changes are necessary/right and as they are within editorial discretion, I am closing the PR unmerged.

PR #592, opened 2023-11-15
Deployment considerations: Improving the compatibility middlebox sect…

taddhar:
…ion a little bit.

In this proposal I believe what is missing is the intention of this section. So I simply made it explicit so that it is easier to read what then comes after. 

If not, it took me a long time and a discussion with colleagues to understand it without this context. Now you have a 'direction' of reading and it falls better in at least my mind.


ekr:
Superceded by PR #600

PR #609, opened 2024-02-25
Explain when clients can remember rejected ECH. Fixes #604

### LINKED ISSUE #604: Memory for ECH rejection
In https://mailarchive.ietf.org/arch/msg/tls/bvvWbtxJAiMfilfy32EvdaCszQ4/, Elardus Erasmus suggests that the suggestion to limit retries for servers which have rejected ECH is too broad, and we should have different guidance for key/config changes than for just disabling ECH. He also suggests a "holdoff period" when connections were disabled.  I do think that this gets at something real, but I'm not sure of the proposed resolution. 

ISTM that there are two issues:

# Limiting Retries

If the server securely disables ECH, then the client should expect a subsequent connection to succeed and there is no way for the server to ask for a retry. So I don't think we need a limit here.

However, if the server supplies a new key, it's possible to get into some kind of loop, for instance, if there are two servers with different keys and the client bounces between them. Rather than suggesting that there be an unspecified limit, I think it would be better just to forbid a connection initiated from a retry_config from itself causing a new connection. That means that if servers are so misconfigured that they offer a retry_config that then doesn't work that they will just fail, but that should be rare, and in any case, you're in deep trouble in that case.

# Remembering Retry

Unless I've misread the text, it implies but doesn't say that the retry_configs and secure disablement are for this connection only and that the client should not use them even for a connection which is initiated immediately after the one where it retries (as in multiple HTTP 1.1 connections). As Erasmus indicates, this increases load on the server as well as latency and has to be balanced with the observation that the retry_configs are a tracking vector (as noted in the text). He suggests a "holdoff" where you don't retry ECH if it was securely disabled, but no change in cases where a valid retry_config was provided.

I wonder if we should be a bit more expansive about this, given that we have other TLS-related tracking vectors that clients already need to mitigate (e.g., resumption) and there are already mechanisms to handle this. With that in mind, how about:

- Clients should remember updated configuration or disabled ECH for up to the lifetime of the ECHConfig (this also gives some content to "holdoff").
- Clients should partition retry_configs in both time and space the same way they would for resumption PSKs (this is something browsers already do).

Am I missing something important?

ekr:


ekr:
@martinthomson

draft-ietf-tls-esni.md
```diff
@@ -916,6 +916,30 @@ and session IDs presented by the server. These connections are only used to
...
+
+Clients MAY use information learned from a rejected ECH for future
+connections to avoid repeatedly connecting to the same server and
+being forced to retry. However, they MUST handle ECH rejection for
+those connections as if it were a fresh connection, rather than
+enforcing the single retry limit from {{rejected-ech}}.
+
+Clients MUST NOT use ECHConfigs with IP addresses other than
+associated with the DNS queries for the original
+connection. Otherwise, it might become possible for the client to have
```
martinthomson, line 930:
```suggestion
Clients MUST NOT use ECHConfigs with IP addresses other than those
associated with the DNS queries for the original
connection. Otherwise, it might become possible for the client to have
```

However, do you think that you should instead insist on some sort of equality check on something else?  Like the SVCB (or HTTPS) record being for the same QNAME or simply validating that the SVCB record has the same public name?

That latter check seems key, so I might instead say:

```suggestion
Clients MUST NOT reuse ECHConfigs that identify a different public name from
the one used to acquire the IP address of the server. ECHConfigs MUST be freshly
retrieved or taken from a cached SVCB (or HTTPS) record that is within its TTL.
Otherwise, it might become possible for the client to have
```

ekr: @martinthomson  ping

martinthomson: There is no IP in the ECHConfig though, so I'm struggling to understand this.

The trick that I was suggesting was that a server that uses multiple providers could use public name rather than IP to separate those providers.  This is on the presumption that |the set of IP addresses for each provider| > 1 but the number of public names is under their direct control.

ekr: I have to admit that I have kind of lost track of how the DNS pieces fit together, but...

As I remember, the basic design requirement for multi-CDN was that we had to ensure that there was no way that the client could get ECHConfig for CDN A and the IP address for CDN B, because that just won't work. When I had last checked, that was intended to be accomplished by having the IP address delivered alongside the ECHConfig, which I thought was done with ipv6hint. But maybe I've misremembered or things evolved.

@tfpauly @bemasc?

tfpauly: The record can contain v4 and v6 hints, and also is associated with A and AAAA records for that specific service (which would be distinct from the IP addresses for other cdns, etc), that are queried after receiving the HTTPS record.

ekr: Can you elaborate on how that second process works? By dereferencing the TargetName?

tfpauly: Yes, from the target name. See both of these:

https://www.rfc-editor.org/rfc/rfc9460.html#section-7.3

https://www.rfc-editor.org/rfc/rfc9460.html#section-3

dennisjackson:
Reading the discussion, are you sure the complexity of doing this is worth the gain? The principle mechanism server operators have for controlling ECH attempts is updating their DNS records. Presuming they can do that reasonably quickly, aren't most clients going to maintain an open connection for the relevant time period anyway? How long is the window in which we won't have an open usable connection (with ECH or not) but we'd be willing to remember the past ECH failure/retry-config for? 

I'd also point out that we also can't assume that ECHConfigs always come from DNS for the purposes of working out an expiration time. It could be that applications are configured by some other out of band mechanism, or even that they deliberately use a known-bad config and intend to rely on a provided retry-config.

ekr:
The complexity of specifying it or the complexity of implementing it? Right now I read the spec as just prohibiting it.

WRT multiple connections, the case I am concerned about is HTTP connection sharding when you might have 4 or so concurrent connections. I suppose we can just say that this is a reason to use H2 or H3.

WRT not-DNS, I think we could practically just limit this kind of memory to DNS.

Anyway, i don't feel strongly about this, and we could always relax this restriction later, so if others aren't enthusiastic, we can just close this unchanged.

draft-ietf-tls-esni.md
bemasc: In SVCB, ECH is a property of the "endpoint", which is represented by one ServiceMode record (out of an RRset that might contain several).  An endpoint also specifies a pool of IP addresses (by the A+AAAA records on the TargetName, or by IP hints in the SvcParams).  As far as SVCB is concerned, ECH can keep state within the scope of a single endpoint, but that state must not leak across to other endpoints of the same "service" (i.e. origin).

This draft doesn't have the concept of an "endpoint", but I think we can define a close enough notion here (without requiring DNS, which is not required for ECH!): call it an "ECHConfig source".

```suggestion
Any persisted information MUST be associated with the ECHConfig source used
to bootstrap the connection, such as DNS SVCB ServiceMode record {{ECH-IN-DNS}}.
Clients MUST limit any sharing of persisted ECH-related state to connections that use
the same ECHConfig source. Otherwise, it might become possible for the client to have
```

Note that "matching ECHConfig" is arguably too broad a scope: if I have two ServiceMode records pointing at separate server pools that initially have the same ECHConfig, I should be able to update one pool without changing the connection behaviors observed by the other one.

draft-ietf-tls-esni.md
```diff
@@ -916,6 +916,30 @@ and session IDs presented by the server. These connections are only used to
...
+the wrong public name for the server, thus making recovery impossible.
+
+ECHConfigs learned from ECH rejection can be used as a tracking
+vector. Clients SHOULD give them the same lifetime as other server-based
+tracking vectors such as PSKs.
+
+In general, it is safest to try to retrieve a new ECHConfig unless
+a subsequent connection is made shortly after the connection where
+ECH was rejected. This makes it most likely that the client will
```
bemasc, line 939: I really can't tell what this is trying to say.

```suggestion
In general, the safest way for clients to minimize ECH retries is to
comply with any freshness rules (e.g., DNS TTLs) imposed by the
ECHConfig source. This makes it most likely that the client will
```

draft-ietf-tls-esni.md
```diff
@@ -916,6 +916,30 @@ and session IDs presented by the server. These connections are only used to
...
+those connections as if it were a fresh connection, rather than
+enforcing the single retry limit from {{rejected-ech}}.
+
+Clients MUST NOT use ECHConfigs with IP addresses other than
+associated with the DNS queries for the original
+connection. Otherwise, it might become possible for the client to have
+the wrong public name for the server, thus making recovery impossible.
+
+ECHConfigs learned from ECH rejection can be used as a tracking
+vector. Clients SHOULD give them the same lifetime as other server-based
```
bemasc, line 934:
```suggestion
vector. Clients SHOULD impose the same lifetime and scope restrictions
that they apply to other server-based
```

ekr:
I think this may need a little more bake time. I'm going to ship the next version without it and hopefully we can refine over the next few weeks, or alternately decide we don't need it.

sftcd:
On 04/03/2024 02:18, Eric Rescorla wrote:
> I'm going to ship the next version

Yes, shipping a version good enough for WGLC and then IETF LC
is the right plan. There'll be plenty of chances for fixing
wording at those stages.

S.

ekr:
Revisiting this....

1. I think we should make the "single retry" change. I split that into #616. There's no reason that should happen.

2. I'm less sure about the "you can remember" stuff, but on balance I think it's worth it. I've updated it to take @bemasc's comments and now I think it's correct. @martinthomson  @bemasc @dennisjackson @davidben ?

draft-ietf-tls-esni.md
```diff
@@ -884,9 +884,9 @@ version was negotiated, the client can regard ECH as securely disabled
 by the server, and it SHOULD retry the handshake with a new transport
 connection and ECH disabled.
 
-Clients SHOULD implement a limit on retries caused by receipt of "retry_configs"
-or servers which do not acknowledge the "encrypted_client_hello" extension. If
-the client does not retry in either scenario, it MUST report an error to the
+Clients SHOULD NOT accept "retry_config" in response to
+a connection initiated in response to a "retry_config".
```
martinthomson, line 888: I commented on the other PR, but maybe this can point at the new section as a way of explaining the "SHOULD".

draft-ietf-tls-esni.md
```diff
@@ -916,6 +916,32 @@ and session IDs presented by the server. These connections are only used to
...
 instance, by reporting a failed connection with a dedicated error code.
 
+
+### Impact of Retry on Future Connections
+
+Clients MAY use information learned from a rejected ECH for future
+connections to avoid repeatedly connecting to the same server and
+being forced to retry. However, they MUST handle ECH rejection for
+those connections as if it were a fresh connection, rather than
+enforcing the single retry limit from {{rejected-ech}}.
```
martinthomson, line 926: This "however" was a little unclear to me at first.

I think that what you are saying is that when you save a retry config that worked for a future connection, you need to treat a retry that results as though you did not have a retry.  The result is that you will use the retry config and try again.

This needs to be part of the text that explains that you shouldn't accept a retry config if you just used one.  Otherwise, it's a little confusing.

ekr:
@martinthomson  how's this

draft-ietf-tls-esni.md
```diff
@@ -917,6 +920,38 @@ and session IDs presented by the server. These connections are only used to
...
+the server, thus making recovery impossible.
+
+ECHConfigs learned from ECH rejection can be used as a tracking
+vector. Clients SHOULD impose the same lifetime and scope restrictions
+that they apply to other server-based
+tracking vectors such as PSKs.
+
+In general, the safest way for clients to minimize ECH retries is to
+comply with any freshness rules (e.g., DNS TTLs) imposed by the
+have up to date information.
```
chris-wood, line 952:
```suggestion
comply with any freshness rules (e.g., DNS TTLs) imposed by the ECH configuration.
```

chris-wood (approved):

bemasc (approved):

draft-ietf-tls-esni.md
```diff
@@ -917,6 +920,37 @@ and session IDs presented by the server. These connections are only used to
 trigger retries, as described in {{rejected-ech}}. This may be implemented, for
 instance, by reporting a failed connection with a dedicated error code.
 
+
+### Impact of Retry on Future Connections
+
+Clients MAY use information learned from a rejected ECH for future
```
elardus-erasmus, line 926: Wondering if this `MAY` should not be a stronger `SHOULD`. If not implemented by the client, it does mean that we have a doubling in the number of (expensive PKI involving) connections when ECH is disabled, in situations like TLS-terminating proxies, temporary disabling of ECH on a server, or indefinitely in case of a server config issue.

ekr: If it's indefinitely, then the DNS records will time out. I think probably a SHOULD is too strong.

draft-ietf-tls-esni.md
```diff
@@ -917,6 +920,37 @@ and session IDs presented by the server. These connections are only used to
...
+and then the client tries to use that to connect some time
+later, it is possible that the server has been forced to
+change its configuration again and is now trying to recover.
+
+Any persisted information MUST be associated with the ECHConfig source
+used to bootstrap the connection, such as a DNS SVCB ServiceMode record
+{{ECH-IN-DNS}}. Clients MUST limit any sharing of persisted ECH-related
+state to connections that use the same ECHConfig source. Otherwise, it
+might become possible for the client to have the wrong public name for
+the server, thus making recovery impossible.
```
bemasc, line 943:
```suggestion
the server, making recovery impossible.
```

PR #610, opened 2024-03-04
Encourage greasing PSK. Fixes #606

### LINKED ISSUE #606: Recommend greasing PSK?
In Watson Ladd's review, he asks:

  Should we use RFC 2119 language for the server as well? Right now we
  only say what the client must do when the server violates the rules.

We do recommend greasing ECH. My sense is that this is on the bubble
and we could leave it as-is, but I could be persuaded otherwise.

@davidben  @dennisjackson @martinthomson

ekr:


ekr:
@martinthomson @dennisjackson

dennisjackson:
LGTM

PR #615, opened 2024-05-27
Explain about reused keyshares. Fixes #612

### LINKED ISSUE #612: Mention that the server identity may be revealed by ServerHellp
https://mailarchive.ietf.org/arch/msg/tls/5oKWlf---OqjQf37giXVVzIbPHw/

>    "ECH is not in itself sufficient to protect the identity of the
>    server.  The target domain may also be visible through other
>    channels, such as plaintext client DNS queries or visible server IP
>    addresses.  However, DoH [RFC8484] and DPRIVE [RFC7858] [RFC8094]
>    provide mechanisms for clients to conceal DNS lookups from network
>    inspection, and many TLS servers host multiple domains on the same IP
>    address.  Private origins may also be deployed behind a common
>    provider, such as a reverse proxy.  In such environments, the SNI
>    remains the primary explicit signal used to determine the server's
>    identity."
 
This text only discusses that the identity of the server may be revealed by
"other channels". I strongly think the document needs to mention that the
identity of the server may also be reveled by the unencrypted information
in the ServerHello. In particular a reused KeyShare is problematic.
 
Suggested addition:
 
The identity of the server may also be reveled by the unencrypted information
in the ServerHello. Most of the current information in ServerHello is not unique.
The exception is KeyShare, which if reused provides a unique identifier of the server.
 
Cheers,
John Preuß Mattsson

ekr:


ekr:
@martinthomson

martinthomson (approved):
...despite strong recommendations not to reuse a KeyShare value...

This seems almost unnecessary, really.  But it's correct.

PR #616, opened 2024-05-27
Don't retry from retry

ekr:
This was split off from PR#609. Hopefully this is uncontroversial.

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -889,6 +889,7 @@ client reached a node with configuration A in the first connection and
 a node with configuration B in the second. If a client does not retry,
 it MUST report an error to the calling application.
 
+
 ### Authenticating for the Public Name {#auth-public-name}
 
 When the server rejects ECH, it continues with the handshake using the plaintext
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -886,8 +886,14 @@ initiated in response to a "retry_config".  Sending a "retry_config"
 in this situation is a signal that the server is misconfigured, e.g.,
 the server might have multiple inconsistent configurations so that the
 client reached a node with configuration A in the first connection and
-a node with configuration B in the second. If a client does not retry,
-it MUST report an error to the calling application.
+a node with configuration B in the second. Note that this guidance does not apply
+the server does not supply a "retry_config", thus disabling ECH,
+as the client reconnecting in this case will not elicit another
+"retry_config"; clients SHOULD follow the guidance in the previous
+paragraph. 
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -885,9 +885,9 @@ version was negotiated, the client can regard ECH as securely disabled
 by the server, and it SHOULD retry the handshake with a new transport
 connection and ECH disabled.
 
-Clients SHOULD implement a limit on retries caused by receipt of "retry_configs"
-or servers which do not acknowledge the "encrypted_client_hello" extension. If
-the client does not retry in either scenario, it MUST report an error to the
+Clients SHOULD NOT accept "retry_config" in response to
+a connection initiated in response to a "retry_config".
```
martinthomson, line 889: Do we have a little more color to add here?

This is mostly about the case where a retry config is supplied by one server instance, but other server instances don't work.  That is a server configuration error, unless the client decided to keep the retry config for too long, in which case it is a signal that the client shouldn't hold on to old state.

Adding that sort of context might make the SHOULD a little more useful.

ekr:
I added some text.

draft-ietf-tls-esni.md
```diff
@@ -885,10 +885,13 @@ version was negotiated, the client can regard ECH as securely disabled
...
-Clients SHOULD implement a limit on retries caused by receipt of "retry_configs"
-or servers which do not acknowledge the "encrypted_client_hello" extension. If
-the client does not retry in either scenario, it MUST report an error to the
-calling application.
+Clients SHOULD NOT accept "retry_config" in response to a connection
+initiated in response to a "retry_config".  Sending a "retry_config"
+in this situation is a signal that the server is misconfigured, e.g.,
+the server might have multiple inconsistent configurations so that the
+client reached a node with configuration A in the first connection and
+a node with configuration B in the second. If a client does not retry,
```
elardus-sophos, line 893: The previous paragraph also talks about a "retry". But that is an ECH *disabled* retry.
In this paragraph, based on the context, it is referring to an ECH *enabled* retry.
Should the client be allowed to perform an ECH disabled retry here? And should anything about that, allowed or not allowed, be noted in this paragraph also?

ekr: I think this is a reasonable question. Updated the text.

Note that I also broke out the requirement to report the error separately as I think it applies to both cases.

ekr:
@martinthomson?

martinthomson (approved):
What is the merge conflict?

ekr:
I had two retry change PRs off the same place and the first one merged.

PR #625, opened 2024-08-15
Remove extraneous registry reference

jsalowey:
Seems like this is an old reference to a different registry name


ekr:
I think we actually want to make this match the rest of 8447 and allow an I-D. @jsalowey @seanturner what say you about the process?

PR #627, opened 2024-09-14
Update the IANA considerations text to match our other TLS policies.

ekr:
@seanturner  @jsalowey how would you like to handle this.


jsalowey (approved):

jsalowey:
I think this is consistent with other TLS registries and should be merged.

PR #635, opened 2024-11-24
Update the key rotation text. Fixes #629

### LINKED ISSUE #629: Should we recommend how often to rotate keys?
From AD review.

See Section 10.10.5, which just says "frequently".

ekr:


draft-ietf-tls-esni.md
```diff
@@ -1412,14 +1415,21 @@ less useful without encryption of DNS queries in transit mechanisms.
 A malicious client-facing server could distribute unique, per-client ECHConfig
 structures as a way of tracking clients across subsequent connections. On-path
 adversaries which know about these unique keys could also track clients in this
-way by observing TLS connection attempts.
+way by observing TLS connection attempts.
```
martinthomson, line 1418:
```suggestion
way by observing TLS connection attempts.
```

martinthomson (approved):

PR #636, opened 2024-11-24
Invalid names end in .invalid. Fixes #630

### LINKED ISSUE #630: Extraneous configurations MUST have invalid DNS names?
Suggested in AD review.

ekr:
@bemasc

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -951,6 +951,8 @@ configuration.
 
 ## GREASE ECH {#grease-ech}
 
+### Client Greasing
+
 If the client attempts to connect to a server and does not have an ECHConfig
 structure available for the server, it SHOULD send a GREASE {{?RFC8701}}
 "encrypted_client_hello" extension in the first ClientHello as follows:
@@ -985,6 +987,24 @@ invalid. It otherwise ignores the extension. It MUST NOT save the
 Offering a GREASE extension is not considered offering an encrypted ClientHello
 for purposes of requirements in {{real-ech}}. In particular, the client
 MAY offer to resume sessions established without ECH.
+
+### Server Greasing
+
+{{config-extensions-iana}} describes a set of Reserved extensions
+which will never be registered. These can be used by servers to
+"grease" the contents of the ECH configuration, as inspired by
+{{?RFC8701}}. This helps ensure clients process ECH extensions
+correctly. When constructing ECH configurations, servers SHOULD
+randomly select from reserved values with the high-order bit
+clear. Correctly-implemented client will ignore those extensions.
+
+The reserved values with the high-order bit set are mandatory, as
+defined in {{config-extensions}}. Servers SHOULD randomly select from
+these values and include them in extraneous ECH configurations.
+Correctly-implemented clients will ignore these configurations because
+they do not recognize the mandatory extension.  These extraneous ECH
+configurations SHOULD have invalid keys, and invalid public names,
+ending in ".invalid" (see {{?RFC2606}}).
 
 # Server Behavior {#server-behavior}
 
@@ -1914,21 +1934,8 @@ registry:
       The expert may provide more in depth reviews, but their approval
       should not be taken as an endorsement of the extension.
 
-This document defines several Reserved values for ECH configuration extensions.
-These can be used by servers to "grease" the contents of the
-ECH configuration, as inspired by {{?RFC8701}}. This helps ensure clients
-process ECH extensions correctly. When constructing ECH configurations,
-servers SHOULD randomly select from reserved values with the high-order
-bit clear. Correctly-implemented client will ignore those extensions.
-
-The reserved values with the high-order bit set are mandatory, as defined
-in {{config-extensions}}. Servers SHOULD randomly select from these
-values and include them in extraneous ECH configurations.
-Correctly-implemented clients will ignore these configurations
-because they do not recognize the mandatory extension.
-These extraneous ECH configurations SHOULD have invalid keys, and invalid
-public names, ending in ".invalid" (see {{?RFC2606}}).
-
+This document defines several Reserved values for ECH configuration extensions
+to be used for "greasing" as described in {{server-greasing}}.
 
 The initial contents for this registry consists of multiple reserved values,
 with the following attributes, which are repeated for each registration:
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -1002,10 +1002,11 @@ The reserved values with the high-order bit set are mandatory, as
 defined in {{config-extensions}}. Servers SHOULD randomly select from
 these values and include them in extraneous ECH configurations.
 Correctly-implemented clients will ignore these configurations because
-they do not recognize the mandatory extension.  These extraneous ECH
-configurations SHOULD have invalid keys, and invalid public names,
-ending in ".invalid" (see {{?RFC2606}}).
-
+they do not recognize the mandatory extension.  Servers SHOULD ensure
+that any client using these configurations encounters a warning or error
+message.  This can be accomplished by giving the extraneous configurations
+distinctive config IDs or public names, and rejecting the TLS connection or
+inserting an application-level warning message when these are observed.
 # Server Behavior {#server-behavior}
 
 As described in {{topologies}}, servers can play two roles, either as
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -1004,9 +1004,18 @@ these values and include them in extraneous ECH configurations.
 Correctly-implemented clients will ignore these configurations because
 they do not recognize the mandatory extension.  Servers SHOULD ensure
 that any client using these configurations encounters a warning or error
-message.  This can be accomplished by giving the extraneous configurations
-distinctive config IDs or public names, and rejecting the TLS connection or
-inserting an application-level warning message when these are observed.
+message.  This can be accomplished in several ways, including:
+
+* By giving the extraneous configurations distinctive config IDs or
+  public names, and rejecting the TLS connection or inserting an
+  application-level warning message when these are observed.
+
+* By giving the extraneous configurations an invalid public
+  key and a public name not associated with the server, so that
+  the initial ClientHelloOuter will not be decryptable and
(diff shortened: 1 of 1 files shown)


ekr:
@letoams

ekr:
@paulwouters: PTAL

paulwouters (approved):

draft-ietf-tls-esni.md
```diff
@@ -986,6 +988,24 @@ Offering a GREASE extension is not considered offering an encrypted ClientHello
...
+correctly. When constructing ECH configurations, servers SHOULD
+randomly select from reserved values with the high-order bit
+clear. Correctly-implemented client will ignore those extensions.
+
+The reserved values with the high-order bit set are mandatory, as
+defined in {{config-extensions}}. Servers SHOULD randomly select from
+these values and include them in extraneous ECH configurations.
+Correctly-implemented clients will ignore these configurations because
+they do not recognize the mandatory extension.  These extraneous ECH
+configurations SHOULD have invalid keys, and invalid public names,
```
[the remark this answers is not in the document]
martinthomson: See my comment on #637 for why I am not enthusiastic about this general direction.  I've suggested a different structure there.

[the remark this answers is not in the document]
ekr: I don't have a strong opinion about the substance here; this was the previous text.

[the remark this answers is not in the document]
martinthomson: Yeah, I don't think that any of those arguments hold.

For a connection to be established, the server needs to hold the corresponding private key and be willing to use it.  The client needs to use the public name and (somehow) validate it successfully.  The client also needs to ignore the greased codepoint, despite not understanding it and knowing that it means that the entire config should be ignored.  Sure, the client doesn't know that the key pair is or isn't good, but it has multiple cues to go from.

We owe clients that are beyond incompetent no such affordance in specifications.

To take a step back, you only need one clear and unambiguous way to signal that a config should be ignored.  That's the presence of a mandatory, unknown extension.  Adding in name checks a) complicates the validation process and b) prevents me from doing my new, totally awesome thing that might rely on invalid names.  Which I realize that I'm going to have to share now.  Here, it's incomplete, but it might at least make sense: https://github.com/martinthomson/ech-pnmasq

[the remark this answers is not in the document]
sftcd: I scanned Martin's document and it seems quite complex. I do agree we ought revisit the problem, but *not until after the ECH RFC has issued* so can we rather try get this thing out the door?

paulwouters:
It seems Martin's raised point is a new issue. I am fine if we go that way and omit the specifics like how to formulate an invalid hostname. If text will remain about hostnames, it would still be good to suggest using .invalid as an example, without any 2119 enforcement.

draft-ietf-tls-esni.md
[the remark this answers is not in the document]
bemasc: > Adding in name checks a) complicates the validation process and b) prevents me from doing my new, totally awesome thing ...

The text here is not attempting to add name checks.  In fact, this text assumes that clients do _not_ perform any name checks (beyond syntax).  If client applies special treatment to ".invalid" names, the client has "special-cased the grease" and the procedure is broken.

Would you be happier with an additional sentence like "Clients MUST NOT apply special treatment to names based on the ".invalid" TLD."?

ekr:
I tend to agree with @bemasc that in order for the GREASE to induce a failure (which is the point!) we need the ECHConfig to actually be invalid, not just have an unexpected extension. I don't see how this creates a future compatibility issue. @martinthomson is there a specific change here that will avoid creating a problem for your draft? If not, I'd prefer to land this (happy to change the SHOULD to can if you feel strongly).

martinthomson:
The specific thing that I was concerned about is the requirement to check the name.  I believe that it should be possible to have a valid ECH config with a garbage public name in it.  That said, an extension can override the need to run any checks in the main spec, so I can live with a "SHOULD" if pressed.

However, greasing only needs to trip one check.  The mandatory extension one seemed fine for that, particularly since you can reserve a mandatory-to-understand codepoint specifically for that purpose.  Forcing the configuration to be bad is unnecessary as a requirement: that is, it is not necessary to guarantee interoperability, so is a misuse of 2119 language.

bemasc:
@martinthomson I'm confused.  This PR doesn't introduce any requirement for either endpoint to check a name.

> However, greasing only needs to trip one check. The mandatory extension one seemed fine for that, particularly since you can reserve a mandatory-to-understand codepoint specifically for that purpose. Forcing the configuration to be bad is unnecessary as a requirement: that is, it is not necessary to guarantee interoperability, so is a misuse of 2119 language.

If the public name is not "bad", then the grease will "fail to fail".  Broken client implementations (i.e. those that ignore unrecognized mandatory extensions instead of rejecting the config) will trigger the recovery flow, resulting in a successful connection (using a non-grease ECHConfig retrieved in-band) instead of angry users filing bug reports against the client (which is the point of this grease).

martinthomson:
I'm objecting to this:

> These extraneous ECH configurations SHOULD have invalid keys, and invalid public names,

And this claim:

> If the public name is not "bad", then the grease will "fail to fail". 

Which is untrue.  It is only necessary to have one thing wrong.  Operating on the assumption that clients will deliberately violate a clear and well-specified rule in the specification is sometimes appropriate, but only when there is evidence that this has happened.

The goal is achieved when the rule is inoperable.  A strong recommendation (i.e., "SHOULD") does not improve interoperability.

draft-ietf-tls-esni.md
bemasc, line 1006: @martinthomson Is this closer to what you're looking for?

```suggestion
configurations SHOULD have keys and public names that the server does not accept.
Otherwise, incorrectly implemented clients may not fail to connect as intended.
For example, the public name could end in ".invalid" (see {{?RFC2606}}).
```

martinthomson: I'm still stuck on the "SHOULD" and this notion that we need to concern ourselves with incorrectly implemented clients.

ekr: Isn't the entire purpose of this piece of greasing to detect incorrectly implemented clients?

martinthomson: So let them connect and collect their IP addresses?

I guess that I'm struggling with the use of 2119 language here.  Grease is discretionary and serves no direct interoperability purpose.  I don't see why you would so forcibly recommend action like this when this could be a simple "we reserved some codepoints and other stuff, which you can use this way if you are interested".

martinthomson: I should have said, the greased ECH ClientHello extension serves a purpose that depends on it being used, but this one doesn't.

ekr: > So let them connect and collect their IP addresses?

The reasoning here is as follows:

- You provide an ECHConfig with a mandatory unknown extension which should cause a compliant client to skip the ECHConfig
- The ECHConfig is otherwise invalid (but in a way that is not readily apparent), so if it doesn't skip it, it will experience a connection failure.
- The user will report the connection failure and that will help get the client fixed.


> I guess that I'm struggling with the use of 2119 language here. Grease is discretionary and serves no direct interoperability purpose. I don't see why you would so forcibly recommend action like this when this could be a simple "we reserved some codepoints and other stuff, which you can use this way if you are interested".

I don't have a strong opinion either way about the SHOULD, but I do think we should explain the logic above. Perhaps:

> The reserved values with the high-order bit set are mandatory, as
defined in {{config-extensions}}. In order to GREASE correct client
behavior in ignoring these values, servers can randomly select from these values and include them in extraneous ECH configurations which have invalid keys, and invalid public names,
Correctly-implemented clients will ignore these configurations because
they do not recognize the mandatory extension. Incorrectly implemented
clients which skip the extension will attempt to connect to the server
using the ECHConfig, but fail to connect because the server will not
be able to decrypt the encrypted ClientHello or perform the recovery
flow, thus causing a failure on the client.

martinthomson: Why not invalid IP addresses?  There's no place like 127.0.0.1.

My point here is that you don't have a good measure for exactly how stupid the implementation needs to be to be caught out this way.  A perfectly valid approach would include a completely valid configuration, aside from the mandatory extension that no one can support.  Then, if a connection attempt is made, the server can log everything, including the inner ClientHello.  I don't know what you /do/ with that information, but it's a trap for a very particular profile of stupid.

You can make the config more and more obviously silly, but who then are you planning to trap?  For instance, I don't see why you couldn't connect with "foo.invalid" and expect it to work, aside from any retry.

draft-ietf-tls-esni.md
```diff
@@ -986,6 +988,24 @@ Offering a GREASE extension is not considered offering an encrypted ClientHello
...
+ending in ".invalid" (see {{?RFC2606}}).
+
```
bemasc, line 1008: @martinthomson Here's another try:

```suggestion
they do not recognize the mandatory extension.  Servers SHOULD ensure
that any client using these configurations encounters a warning or error
message.  This can be accomplished by giving the extraneous configurations
distinctive config IDs or public names, and rejecting the TLS connection or
inserting an application-level warning message when these are observed.
```

draft-ietf-tls-esni.md
bemasc: I think we are talking past each other slightly (the text assumes that clients MUST NOT special-case .invalid!), but I also think @martinthomson is onto something here.  We don't have to rely on "natural" failure modes (as in the current text).  The server can explicitly recognize the GREASE configs and inject failures.  This allows better GREASE (the config can be otherwise valid) and richer failure modes (e.g. injecting a warning message into webpage content).

I've added a suggestion below.

draft-ietf-tls-esni.md
martinthomson: I really like this one.  I do think that your list could include making the configuration broken as well (not with a SHOULD, but listed with your "cam be accomplished" phrasing or similar), but this new stuff works for me.

ekr:
@martinthomson  @bemasc how does this look?

martinthomson (approved):

bemasc (approved):

PR #637, opened 2024-11-25
Issue628 dns names

ekr:


ekr:
@paulwouters: PTAL

paulwouters (approved):

draft-ietf-tls-esni.md
```diff
@@ -291,14 +291,14 @@ as described in {{rejected-ech}}.
...
-`ECHConfig` if the length of any label in the DNS name is longer than 63
-octets, as this is the maximum length of a DNS label.
+{{!RFC5890, Section 2.3.1}} or which begins or end with an ASCII dot.
+Clients additionally SHOULD ignore the structure if it represents an
+IPv4 address {{!RFC791}} in textual or hexadecimal form (IPv6
+addresses are invalid DNS names due to the presence of the ":"
+character, and thus are excluded by the previous
+requirement). Additionally, clients MUST ignore the `ECHConfig` if the
+length of any label in the DNS name is longer than 63 octets, as this
+is the maximum length of a DNS label.
```
martinthomson, line 301: How many would properly ignore such an ECH config?

martinthomson (commented):
I'm not a fan of the MUST here.  I have a proposal (unpublished) that points out that the public_name field can be any value as long as the client has a means of validating it.

That means that as long as the client needs to follow the validation rules in this document, then it will need to validate the public name on the basis of it being valid.  But that's just one way to perform that validation (and a particularly expensive one as it happens).  It is entirely possible to define other validation methods that do not require that the name be valid for DNS.

If I may, can I suggest instead that these requirements be moved to `{{auth-public-name}}` with a note there that says that clients MAY ignore ECH configs for which validation is obviously impossible.  Then you would list *there* things like the IP address matching and name construction as examples of things for which it is impossible to obtain a valid DNS-backed certificate.

ekr:
I think this version is now OBE.

ekr:
This should now be resolved.

PR #640, opened 2024-11-27
Use the DTLS ClientHello structure in ECH with DTLS

### LINKED ISSUE #639: Clarify interaction with DTLS 1.3
The draft claims to work with DTLS, but it's slightly ambiguous. We use `ClientHello` structures directly in ECH structures, but `ClientHello` is one of the structures that DTLS changes. Do we use the DTLS version or the TLS version? The differences are:

1. The `legacy_version` field is hardcoded to a different value.
2. There is a goofy `legacy_cookie` field.

The goofy `legacy_cookie` field is interesting. In principle, it should really should get the same treatment as `legacy_session_id` and be copied over, because there's nothing to hide there. But I think it's actually always empty, since HelloVerifyRequest is banned in DTLS 1.3. But then we probably need to answer dumb questions like:

1. What happens if a client sends an ECH-ful ClientHello and gets HelloVerifyRequest back?
2. What happens if a server receives an ECH-ful ClientHello but the legacy_cookie field isn't empty?
3. Do we have injectivity problems allowing the two ClientHellos in there? I think we're fine, because `legacy_version` distinguishes the two ClientHellos but, ugh, why oh why did we do this in DTLS??

Not terribly difficult, but it's ambiguous enough that I think we need to say something here. My inclination is to say that we use the DTLS ClientHello, and pick the least annoying answers for (1) and (2). But I think we do need to write them down somewhere.

davidben:
This PR implements the option if we decide to use DTLS's ClientHello as the payload and AAD, rather than somehow fit TLS's in there.

Fixes #639.

I haven't yet gotten this working for BoringSSL (probably will finish that after the holiday), but I suspect this is the more straightforward option for us. @martinthomson what has NSS implemented / would be more convenient for NSS?

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -1373,7 +1373,7 @@ has size k = 1. Client-facing servers SHOULD deploy ECH in such a way so as to
 maximize the size of the anonymity set where possible. This means client-facing
 servers should use the same ECHConfig for as many hosts as possible. An
 attacker can distinguish two hosts that have different ECHConfig values based
-on the ECHClientHello.config_id value. 
+on the ECHClientHello.config_id value.
 
 This also means public information in a TLS handshake should be
 consistent across hosts. For example, if a client-facing server
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -481,10 +481,12 @@ into a EncodedClientHelloInner structure, defined below:
 ~~~
 
 The `client_hello` field is computed by first making a copy of ClientHelloInner
-and setting the `legacy_session_id` field to the empty string. Note this field
-uses the ClientHello structure, defined in {{Section 4.1.2 of RFC8446}} which
-does not include the Handshake structure's four byte header. The `zeros` field
-MUST be all zeroes.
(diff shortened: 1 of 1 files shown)


davidben:
NB: This option means needing to reason about injectivity. But the first two bytes of a DTLS ClientHello are different, so I think we're alright on that front? (Although nothing is stopping someone from sending 0xfefd in the legacy version of a TLS ClientHello, so... 🤷

martinthomson (approved):
This seems right.

davidben:
> I haven't yet gotten this working for BoringSSL

I can now report that there is at least one implementation. 🙂 Possibly two if NSS had already done it before all this. Also our Go testing stack if you count that one, but that stack is weird.

It was a lot of chasing down of the TLS vs DTLS header (our DTLS stack carries the old style header around because it was the way you spelled it in 1.2), and making sure all the ClientHello processing bits picked up the right now. But nothing fundamentally hard. Just more ways that we pay for DTLS 1.2's sins.

ekr:
I believe this is the correct resolution. @seanturner @jsalowey @dconnolly  are you OK with merging it?

davidben:
Also got https://github.com/tlswg/draft-ietf-tls-esni/pull/641. 😞

PR #644, opened 2025-03-19
Remove extra confusing sentence. Fixes #643

### LINKED ISSUE #643: Meaning of "authenticated" in name validation section
First off, apologies for this issue right after #642 was merged. I thought I had left a comment there but apparently the feature in Github I used left a "Pending" comment that doesn't gets submitted till I complete my review. 

In essence though, the change I wanted to ask about is:

https://github.com/tlswg/draft-ietf-tls-esni/blob/29fe76cebd1b216d4f8c0bd16b8dc5e7ccdef9b8/draft-ietf-tls-esni.md?plain=1#L911-L912

what does "ensure that the public_name can be authenticated" mean in this context?

ekr:
@martinthomson @bemasc

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -908,8 +908,8 @@ and session IDs presented by the server. These connections are only used to
 trigger retries, as described in {{rejected-ech}}. This may be implemented, for
 instance, by reporting a failed connection with a dedicated error code.
 
-Prior to attempting a connection, a client SHOULD validate the `ECHConfig` to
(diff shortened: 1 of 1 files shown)


bemasc (approved):

bemasc:
Yeah, I think this is correct.  "Can be authenticated" is a pretty hazy concept and possibly not correct.  The point is that `public_name` needs to be syntactically valid for the SNI extension field, so that the server can't cause the client to violate the SNI RFC.  IP addresses are excluded because of the syntax rule even though they _could_ be authenticated.

PR #657, opened 2025-07-11
Fix ECHConfigList range

tweksteen:
From the paragraph just below: "The ECHConfigList structure contains one or more ECHConfig structures in decreasing order of preference."

### DIFF
--- a/draft-ietf-tls-esni.md
+++ b/draft-ietf-tls-esni.md
@@ -327,7 +327,7 @@ The client-facing server advertises a sequence of ECH configurations to clients,
 serialized as follows.
 
 ~~~~
-    ECHConfig ECHConfigList<4..2^16-1>;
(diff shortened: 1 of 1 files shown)


draft-ietf-tls-esni.md
```diff
@@ -327,7 +327,7 @@ The client-facing server advertises a sequence of ECH configurations to clients,
 serialized as follows.
 
 ~~~~
-    ECHConfig ECHConfigList<4..2^16-1>;
+    ECHConfig ECHConfigList<1..2^16-1>;
```
martinthomson, line 330: This is a game we play often.  There's a view that says that you don't set bounds like this.  Because then you get someone (like me) coming along to say that the minimum size of a single ECHConfig is 4 bytes.

tweksteen: Fair enough :) Thanks for the context.

PR #659, opened 2025-11-15
RPC1

ekr:


rfc9849.md
```diff
@@ -1,58 +1,93 @@
...
-       name: Christopher A. Wood
-       organization: Cloudflare
+       org: Cloudflare
        email: caw@heapingbits.net
 
 
 normative:
   RFC2119:
   RFC7918:
+  RFCYYY1:
```
ekr, line 46: This should be informative

rfc9849.md
```diff
@@ -63,9 +98,48 @@ informative:
...
+(https://url.spec.whatwg.org/), and adding a "commit snapshot" URL to the
+reference. 
+
+Current:
+[WHATWG-IPV4]
+           WHATWG, "URL - IPv4 Parser", WHATWG Living Standard, May
+           2021, <https://url.spec.whatwg.org/#concept-ipv4-parser>.
+
+b) RFC 6125 has been obsoleted by RFC 9525.  May we replace RFC 6125
+with RFC 9525?
```
ekr, line 130: Yes.

rfc9849.md
```diff
@@ -63,9 +98,48 @@ informative:
...
+
+We recommend updating this reference to the most current version of the WHATWG
+Living Standard, replacing the URL with the more general URL to the standard
+(https://url.spec.whatwg.org/), and adding a "commit snapshot" URL to the
+reference. 
+
+Current:
+[WHATWG-IPV4]
+           WHATWG, "URL - IPv4 Parser", WHATWG Living Standard, May
+           2021, <https://url.spec.whatwg.org/#concept-ipv4-parser>.
```
ekr, line 127: I'm not sure that that's advisable in this case. This is an informative reference, so why wouldn't we want the latest version?

rfc9849.md
```diff
@@ -63,9 +98,48 @@ informative:
...
+
+b) RFC 6125 has been obsoleted by RFC 9525.  May we replace RFC 6125
+with RFC 9525?
+
+c) Informative Reference RFC 5077 has been obsoleted by RFC 8446.  We
+recommend replacing RFC 5077 with RFC 8446.  However, if RFC 5077 must be
+referenced, we suggest mentioning RFC 8446 (e.g., RFC 5077 has been obsoleted
+by RFC 8446).  See Section 4.8.6 in the RFC Style Guide (RFC 7322).
```
ekr, line 135: This is correct as-is. 5077 is being referenced for a specific construct which doesn't have a correspondent in RFC 8466. It's just an example, and so is fine as informational.

rfc9849.md
```diff
@@ -80,11 +154,10 @@ Indication (SNI) extension in ClientHello messages, which leaks the target
...
 left unencrypted in TLS 1.3.
 
-This document specifies a new TLS extension, called Encrypted Client Hello
-(ECH), that allows clients to encrypt their ClientHello to the TLS server.
+This document specifies a new TLS extension called Encrypted Client Hello
+(ECH) that allows clients to encrypt their ClientHello to the TLS server.
 This protects the SNI and other potentially sensitive fields, such as the
-Application Layer Protocol Negotiation (ALPN)
-list {{?RFC7301}}. Co-located servers with consistent externally visible TLS
+Application-Layer Protocol Negotiation (ALPN) list {{?RFC7301}}. Co-located servers with consistent externally visible TLS
```
ekr, line 160: Rewrap.

rfc9849.md
```diff
@@ -166,7 +239,7 @@ does not have access to the plaintext of the connection beyond the unencrypted
 portions of the handshake.
 
 In the remainder of this document, we will refer to the ECH-service provider as
-the "client-facing server" and to the TLS terminator as the "backend server".
+the "client-facing server" and the TLS terminator as the "backend server".
```
ekr:
```suggestion
the "client-facing server" and to the TLS terminator as the "backend server".
```

The parallel construction reads better.

rfc9849.md
```diff
@@ -306,21 +379,21 @@ order. ECHConfigExtension values are described below ({{config-extensions}}).
...
 key using a KEM they do not support.
 
-public_key
+public_key:
 : The HPKE public key used by the client to encrypt ClientHelloInner.
 
-cipher_suites
-: The list of HPKE KDF and AEAD identifier pairs clients can use for encrypting
+cipher_suites:
+: The list of HPKE Key Derivation Function (KDF) and Authenticated Encryption with Associated Data (AEAD) identifier pairs clients can use for encrypting
```
ekr, line 396: Rewrap.

rfc9849.md
```diff
@@ -637,7 +710,7 @@ ClientHello, with the exception of the following rules:
 1. When the client offers the "pre_shared_key" extension in ClientHelloInner, it
    SHOULD also include a GREASE "pre_shared_key" extension in ClientHelloOuter,
    generated in the manner described in {{grease-psk}}. The client MUST NOT use
-   this extension to advertise a PSK to the client-facing server. (See
+   this extension to advertise a Pre-Shared Key (PSK) to the client-facing server. (See
```
ekr:
```suggestion
   this extension to advertise a PSK to the client-facing server. (See
```

I don't think this is helpful, given that we've already expanded it twice above, albeit not inline with parentheses.

rfc9849.md
```diff
@@ -796,6 +869,17 @@ If the server accepts ECH, the client handshakes with ClientHelloInner as
...
+<!-- [rfced] In the following sentence, does "length other than 8" refer to
+bytes? If yes, may we update as follows?
+
+Current: 
+Otherwise, if it has a length other than 8, the client aborts the
+handshake with a "decode_error" alert.
+
+Perhaps: 
+Otherwise, if it has a length other than 8 bytes, the client aborts
+the handshake with a "decode_error" alert.  -->
```
ekr, line 881: I don't believe this is necessary. Lengths in bytes in TLS.

rfc9849.md
```diff
@@ -897,8 +981,8 @@ application.
 ### Authenticating for the Public Name {#auth-public-name}
 
 When the server rejects ECH, it continues with the handshake using the plaintext
-"server_name" extension instead (see {{server-behavior}}). Clients that offer
-ECH then authenticate the connection with the public name, as follows:
+"server_name" extension instead (see {{server-behavior}}). Then, clients that offer
+ECH authenticate the connection with the public name as follows:
```
ekr:
```suggestion
"server_name" extension instead (see {{server-behavior}}). Clients that offer
ECH then authenticate the connection with the public name as follows:
```

I prefer the original.

rfc9849.md
```diff
@@ -925,17 +1009,17 @@ instance, by reporting a failed connection with a dedicated error code.
 Prior to attempting a connection, a client SHOULD validate the `ECHConfig`.
 Clients SHOULD ignore any
 `ECHConfig` structure with a public_name that is not a valid host name in
-preferred name syntax (see {{Section 2 of ?DNS-TERMS=RFC9499}}).  That is, to be
-valid, the public_name needs to be a dot-separated sequence of LDH labels, as
+preferred name syntax (see {{Section 2 of RFC9499}}).  That is, to be
+valid, the public_name needs to be a dot-separated sequence of LDH labels, as
```
ekr:
```suggestion
valid, the public_name needs to be a dot-separated sequence of LDH labels, as
```

rfc9849.md
```diff
@@ -1013,6 +1097,17 @@ Offering a GREASE extension is not considered offering an encrypted ClientHello
...
+the sentence below and updated as follows. Please let us know if that is not
+accurate.
+
+Original:
+Correctly-implemented client will ignore those extensions.
+
+Current:
+Correctly implemented clients will ignore those extensions.
+-->
+
```
ekr, line 1110: Yes, this is correct.

rfc9849.md
```diff
@@ -1214,6 +1309,16 @@ HelloRetryRequest.  This means that the client-facing server either needs to
...
 
+<!-- [rfced] May we rephrase the following text for an improved sentence flow?
+
+Original:
+The backend server embeds in ServerHello.random a string derived from
+the inner handshake.
+
+Perhaps: 
+A string derived from the inner handshake is embedded into
+ServerHello.random by the backend server.  -->
```
ekr, line 1320: Instead:
The backend server embeds   string derived from
the inner handshake in ServerHello.random.

rfc9849.md
```diff
@@ -1331,16 +1435,46 @@ when different endpoints might have different configurations.
...
+  TLS-terminating proxies even in cases where the server supports ECH
+  but the proxy does not, as detailed below.
+
+Option B (some requirements require it):
+  The requirements in [RFC8446], Section 9.3 that require proxies to
+  act as conforming TLS client and server provide interoperability with
+  TLS-terminating proxies even in cases where the server supports ECH
+  but the proxy does not, as detailed below.
+-->
+
```
ekr, line 1467: I don't think interpretation of this text should depend on "which" vs. "that". Rather the text should be clear.

I'm not sure that this distinction actually matters, but under the assumption that it does, the first and second bullets seem not to apply to proxies at all.

```
  [RFC8446], Section 9.3 requires proxies to act as conforming TLS client and
  servers. This provides interoperability with TLS-terminating proxies even in
  cases where the server supports ECH but the proxy does not, as detailed below.
```

rfc9849.md
```diff
@@ -1557,10 +1691,7 @@ ECH requires encrypted DNS to be an effective privacy protection mechanism.
 However, verifying the server's identity from the Certificate message,
 particularly when using the X509 CertificateType, may result in additional
 network traffic that may reveal the server identity. Examples of this traffic
-may include requests for revocation information, such as OCSP or CRL traffic, or
-requests for repository information, such as authorityInformationAccess. It may
-also include implementation-specific traffic for additional information sources
-as part of verification.
+may include requests for revocation information, such as Online Certificate Status Protocol (OCSP) or Certificate Revocation List (CRL) traffic, or requests for repository information, such as authorityInformationAccess. It may also include implementation-specific traffic for additional information sources as part of verification.
```
ekr, line 1694: Reflow

rfc9849.md
```diff
@@ -1935,27 +2065,26 @@ requirements are detailed in {{encoding-inner}}.
...
+   to "Y".
+1. ech_outer_extensions (0xfd00), with the "TLS 1.3" column values set to "CH",
+   "DTLS-Only" column set to "N", "Recommended" column set to "Y", and the
    "Comment" column set to "Only appears in inner CH."
 
 ## Update of the TLS Alert Registry {#alerts}
 
-IANA is requested to create an entry, ech_required(121) in the existing registry
-for Alerts (defined in {{!RFC8446}}), with the "DTLS-OK" column set to
+IANA has created an entry, ech_required (121) in the existing "TLS Alerts" registry (defined in {{!RFC8446}}), with the "DTLS-OK" column set to
```
ekr, line 2080: Reflow.

rfc9849.md
```diff
@@ -2008,15 +2137,61 @@ Recommended:
...
+expansion in the document carefully to ensure correctness.
+-->
+
+<!-- [rfced] Please review the "Inclusive Language" portion of the online
+Style Guide <https://www.rfc-editor.org/styleguide/part2/#inclusive_language>
+and let us know if any changes are needed.  Updates of this nature typically
+result in more precise language, which is helpful for readers. Note that our
+script did not flag any words in particular, but this should still be reviewed
+as a best practice. -->
+
```
ekr, line 2191: This is fine.

rfc9849.md
```diff
@@ -2008,15 +2137,61 @@ Recommended:
...
+payload
+public_key
+ServerHello.random
+zeros
+-->
+
+<!-- [rfced] We note that these terms are used inconsistently. Please let us
+know which form you prefer.
+
+split-mode vs. Split Mode
```
ekr, line 2176: split mode (except in title case obviously)

rfc9849.md
```diff
@@ -2008,15 +2137,61 @@ Recommended:
...
+GREASE vs. Grease (IANA Section)
```
ekr, line 2177: GREASE

rfc9849.md
```diff
@@ -2008,15 +2137,61 @@ Recommended:
...
+ECHConfigContents
+ECHConfigList
+EncodedClientHelloInner
+inner
+maximum_name_length
+outer
+payload
+public_key
+ServerHello.random
+zeros
```
ekr, line 2170: Thanks. Fixed width should be used for field names and other PDUs. 

I notice that some of these are regular words (zeros) so you have to determine from context whether it's referring to some protocol element or just to the concept "carries an encrypted payload" versus "the `payload` field". Do you want to take a cut at changing as many of these as make sense and then I can review, or would you prefer I make the changes?

One question is what to do in definition lists. My sense is that the list heds should be non-fixed-width but maybe you have a convention.

rfc9849.md
martinthomson: FWIW, WHATWG have directly asked us not to do commit references, but to reference the latest version always.  I would instead take the current, remove the date, and leave it at that.  (I understand that maybe the tooling is not able to handle citations without dates, so you could just change that to December 2025 instead.)

ekr: Do you have a reference to WHATWG's statement on this?

martinthomson: https://whatwg.org/working-mode#anchors talks about their policy with respect to referencing their work.

This includes:

> other standards organizations are discouraged from referencing these snapshots

ekr: Thank you.

rfc9849.md
ekr: Done.

rfc9849.md
ekr: Done.

rfc9849.md
ekr: Done.

rfc9849.md
ekr: Done.

rfc9849.md
ekr: Done.

PR #675, opened 2026-02-07
Minor editorial

ekr:


ekr:
@martinthomson @caw please give me a quick check

rfc9849.md
```diff
@@ -1067,8 +1070,8 @@ structure available for the server, it SHOULD send a GREASE {{?RFC8701}}
 - Set the `config_id` field to a random byte.
 
 - Set the `cipher_suite` field to a supported HpkeSymmetricCipherSuite. The
-  selection SHOULD vary to exercise all supported configurations, but MAY be
-  held constant for successive connections to the same server in the same
+  selection SHOULD vary, so that all supported configurations are exercised,
+  but MAY be held constant for successive connections to the same server in the same
   session.
```
martinthomson, line 1075: The risk of making editorial changes is that you draw attention to things.

In this case, the word "supported" here.  If the client is greasing without expecting to do ECH, what does "supported" mean there?  The "don't stick out" logic suggests that the value should be plausible, but the client that doesn't implement ECH doesn't have a set of supported HPKE cipher suites.  So maybe the word "plausible" can be used and a note added:

> A "plausible" configuration is one that an observer might expect to see.
> A client that fully supports ECH will have a set of supported HPKE cipher suites to select from.
> A client that only supports GREASE ECH has no such list,
> so it should select from a set of values that are in common usage.

rfc9849.md
```diff
@@ -1383,7 +1386,7 @@ to client, client-facing server, and backend server. Coordination between
 client-facing and backend server requires care, as deployment mistakes
 can lead to compatibility issues. These are discussed in {{compat-issues}}.
 
-Beyond coordination difficulties, ECH deployments may also induce challenges
+Beyond coordination difficulties, ECH deployments may also create  challenges
```
martinthomson:
```suggestion
Beyond coordination difficulties, ECH deployments may also create challenges
```

rfc9849.md
```diff
@@ -1383,7 +1386,7 @@ to client, client-facing server, and backend server. Coordination between
...
 for use cases of information that ECH protects. In particular,
```
martinthomson:
```suggestion
for uses of information that ECH protects. In particular,
```

martinthomson (approved):
LGTM, though you inadvertently shone a spotlight on some text and I had thoughts.

PR #676, opened 2026-02-16
Plausible, per MT's suggestion

ekr:
@martinthomson


martinthomson (approved):
That note will be part of the bullet, but this seems fine.

ekr:
Yes, I decided it was best as part of the bullet.

PR #678, opened 2026-02-16
Clarify the retry rules. Fixes #673

### LINKED ISSUE #673: IP address for reconnecting:
> Clients can implement a new transport connection in a way that best
suits their deployment. For example, clients can reuse the same server
IP address when establishing the new transport connection or they can
choose to use a different IP address if provided with options from
DNS. ECH does not mandate any specific implementation choices when
establishing this new connection.

This seems a bit overbroad. Could you do a complete new resolution and use new IPs from that? I think the answer is "no".

ekr:
@martinthomson @chris-wood @davidben I think we need to say something like this.

### DIFF
--- a/rfc9849.md
+++ b/rfc9849.md
@@ -947,13 +947,13 @@ SHOULD retry the handshake with a new transport connection using the
 retry configurations supplied by the server.
 
 Because the new ECH configuration replaces the old ECH configuration,
-so clients can implement a new transport connection in any way that is
+clients can implement a new transport connection in any way that is
 consistent with the previous ECH configuration. For example, clients
 can reuse the same server IP address when establishing the new
 transport connection or they can choose to use a different IP address
 if DNS provided other IP addresses for the previous configuration.
 However, it is not safe to use IP addresses discovered with a new
-DNS query, as those may reflect inconsistent a different ECH server
+DNS query, as those may correspond to a different ECH server
 configuration, for instance associated with a different ECH server
 with a different `public_name`.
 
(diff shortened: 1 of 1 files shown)


rfc9849.md
```diff
@@ -946,12 +946,16 @@ regard the ECH configuration as securely replaced by the server. It
...
 retry configurations supplied by the server.
 
-Clients can implement a new transport connection in a way that best
-suits their deployment. For example, clients can reuse the same server
-IP address when establishing the new transport connection or they can
-choose to use a different IP address if provided with options from
-DNS. ECH does not mandate any specific implementation choices when
-establishing this new connection.
+Because the new ECH configuration replaces the old ECH configuration,
+so clients can implement a new transport connection in any way that is
```
martinthomson:
```suggestion
clients can implement a new transport connection in any way that is
```

rfc9849.md
```diff
@@ -946,12 +946,16 @@ regard the ECH configuration as securely replaced by the server. It
...
+consistent with the previous ECH configuration. For example, clients
+can reuse the same server IP address when establishing the new
+transport connection or they can choose to use a different IP address
+if DNS provided other IP addresses for the previous configuration.
+However, it is not safe to use IP addresses discovered with a new
+DNS query, as those may reflect inconsistent a different ECH server
```
martinthomson:
```suggestion
DNS query, as those may correspond to a different ECH server
```

martinthomson: Or ...

```suggestion
DNS query without also using the ECH configuration from that query.
A fresh DNS query can return a different ECH server
```

martinthomson (approved):

ekr:
@paulwouters @seanturner Do you think this needs WG consultation

paulwouters (approved):

paulwouters:
> @paulwouters @seanturner Do you think this needs WG consultation

no I think this is minor and obvious enough.

